PR #1348, opened 2025-02-03
Add WildChat support

penfever:
# Description

Added support for the WildChat dataset from AllenAI.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


configs/recipes/smollm/sft/1.7m/full_train_distill_r1_670b_math.yaml
```diff
@@ -0,0 +1,49 @@
+model:
```
wizeng23, line 1: Do you mind adding a comment header for your new config files similar to the ones in existing configs? It could be because I don't have context, but I can't tell what this config is for based on the name alone

configs/recipes/smollm/sft/1.7m/full_train_distill_r1_670b_math.yaml
wizeng23, line 1: nit: The pattern I've been following for existing configs is to have their suffix reflect their role. Could you rename this to configs/recipes/smollm/sft/1.7m/full_distill_r1_670b_math_train.yaml or something like that? Thanks!

configs/recipes/smollm/sft/1.7m/gcp_full_train_distill_r1_670b_math.yaml
wizeng23, line 1: Ditto, what about `configs/recipes/smollm/sft/1.7m/full_distill_r1_670b_math_gcp_job.yaml`?

penfever:
@wizeng23 , I removed the 1.7m configs for now, we can revisit in a later PR if we want to

PR #1349, opened 2025-02-03
Update sft_datasets.md

penfever:
# Description

Updated the documentation for sft_datasets with more precise instructions for how to add a new dataset and how to override dataset_name to utilize existing datasets.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/docs/resources/datasets/sft_datasets.md
+++ b/docs/resources/datasets/sft_datasets.md
@@ -113,7 +113,7 @@ For more advanced SFT dataset implementations, explore the `oumi.datasets` modul
 
 ### Using an Unregistered Dataset Whose Format is Identical to a Registered Dataset
 
-Many datasets on HuggingFace share the same format as Oumi registered datasets. It is not necessary to register each dataset explicitly to use it. Instead, you can override the `dataset_name` parameter.
+Many datasets on HuggingFace share the same format as Oumi registered datasets. It is not necessary to register each dataset explicitly to use it. Instead, you can override the `dataset_name` parameter using a keyword argument; see the code snippet below for an example of how to do this.
 
 ```
 - dataset_name: registered_hf_dataset_with_compatible_class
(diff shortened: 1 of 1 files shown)


docs/resources/datasets/sft_datasets.md
```diff
@@ -72,6 +72,7 @@ To add a new SFT dataset:
 
 1. Subclass {py:class}`~oumi.core.datasets.BaseSftDataset`
 2. Implement the {py:meth}`~oumi.core.datasets.BaseSftDataset.transform_conversation` method to define the dataset-specific transformation logic.
+3. Register your new dataset to the dataset class by adding it to {py:class}`~oumi.core.datasets.__init__.py` and {py:class}`~oumi.datasets.sft.__init__.py`.
```
taenin, line 75: Does `{py:class}` work properly when referencing an `__init__.py` file?

taenin (approved):

docs/resources/datasets/sft_datasets.md
```diff
@@ -110,6 +111,18 @@ class CustomSftDataset(BaseSftDataset):
 For more advanced SFT dataset implementations, explore the `oumi.datasets` module, which contains implementations of several [open source datasets](https://github.com/oumi-ai/oumi/tree/main/src/oumi/datasets).
 ```
 
+### Using an Unregistered Dataset Whose Format is Identical to a Registered Dataset
+
+Many datasets on HuggingFace share the same format as Oumi registered datasets. It is not necessary to register each dataset explicitly to use it. Instead, you can override the `dataset_name` parameter.
```
xrdaukar, line 116: should we say `... dataset_name_override:` kwargs parameter or somesuch for extra clarity?

xrdaukar (approved):

docs/resources/datasets/sft_datasets.md
penfever: Nope, good catch. Refactored to {py:mod}

docs/resources/datasets/sft_datasets.md
penfever: Thanks @xrdaukar, I added a bit more text here to make it clear that users should refer to the example for further clarification.

PR #1379, opened 2025-02-05
Evaluation - LM Harness: Adding vLLM support

kaisopos:
# Description

**_TEXT_**
- Adding support for the `VLLM` inference engine in our LM Harness integration (`vllm`):
- Tested: `SmolLM2-135M-Instruct` LM Harness eval with both engines (`NATIVE` and `VLLM`)

**_VISION_**
- Adding support for the `VLLM` inference engine in our LM Harness integration (`vllm-vlm`):
- Tested: `Qwen/Qwen2-VL-2B-Instruct` LM Harness eval with both engines (`NATIVE` and `VLLM`)
   - `NATIVE` works fine
   - `VLLM` does _NOT work for the latest transformers version_ (see [bug](https://github.com/vllm-project/vllm/issues/11711)), but verified that it works for 4.47.0.

## Related issues

Towards [OPE-940](https://linear.app/oumi/issue/OPE-940)

## Before submitting

## Reviewers


src/oumi/evaluation/lm_harness.py
```diff
@@ -91,6 +94,25 @@ def evaluate(
...
+        if device == "cuda:0":
+            logger.warning(
+                "Since you have GPU support, it is highly recommended that you set "
+                "the `inference_engine` to `VLLM`, instead of the `NATIVE`, for faster "
+                "evaluation."
+            )
+    elif inference_engine_type == InferenceEngineType.VLLM:
+        vllm_engine = True
+        if device != "cuda:0":
+            raise ValueError("The `VLLM` inference_engine requires a CUDA-enabled GPU.")
```
xrdaukar, line 109: nit: replace `VLLM` with `vLLM` everywhere. it's their official name

src/oumi/evaluation/lm_harness.py
```diff
@@ -74,6 +76,7 @@ def evaluate(
         model_params: The parameters of the model to evaluate.
         task_params: The LM Harness parameters to use for evaluation.
         generation_params: The generation parameters to use for evaluation.
+        inference_engine_type: The inference engine to use (`VLLM` or `NATIVE`).
```
xrdaukar, line 79: `vLLM`

src/oumi/evaluation/lm_harness.py
```diff
@@ -116,7 +138,7 @@ def evaluate(
             _create_extra_lm_harness_model_params_for_vlm(model_params)
         )
     else:
-        lm_harness_model = "hf"
```
xrdaukar, line 119: let's mirror these changes for Multimodal case above:  `hf-multimodal` vs  'vllm-vlm' ?

src/oumi/evaluation/lm_harness.py
```diff
@@ -91,6 +94,25 @@ def evaluate(
         device = "cpu"
         logger.warning("No GPU available.")
 
+    # Ensure the requested inference engine type is applicable.
+    if inference_engine_type == InferenceEngineType.NATIVE:
+        vllm_engine = False
+        if device == "cuda:0":
```
xrdaukar, line 100: `cuda:0` may be too specific. How about `startswith("cuda")`  or somesuch ?

src/oumi/evaluation/lm_harness.py
```diff
@@ -91,6 +94,25 @@ def evaluate(
...
+            logger.warning(
+                "Since you have GPU support, it is highly recommended that you set "
+                "the `inference_engine` to `VLLM`, instead of the `NATIVE`, for faster "
+                "evaluation."
+            )
+    elif inference_engine_type == InferenceEngineType.VLLM:
+        vllm_engine = True
+        if device != "cuda:0":
```
xrdaukar, line 108: ditto for `cuda:0`

configs/recipes/smollm/evaluation/135m/eval.yaml
```diff
@@ -28,4 +28,5 @@ tasks:
     eval_kwargs:
       num_fewshot: 5
 
-enable_wandb: True
+inference_engine: NATIVE
+enable_wandb: False
```
xrdaukar, line 32: I noticed `enable_wandb: True` sometimes leads to errors  (IIRC if multiple eval tasks are configured ). Have you observed it too ?  Shall we file a bug for this?

src/oumi/evaluation/lm_harness.py
kaisopos: The reason I used VLLM with "`", is that this is the actual value that we use in the yaml file:
`inference_engine: VLLM`

Do you think it is preferable to format it the right way (but not be the value that our yaml expects)?

src/oumi/evaluation/lm_harness.py
kaisopos: done

src/oumi/evaluation/lm_harness.py
kaisopos: done

configs/recipes/smollm/evaluation/135m/eval.yaml
kaisopos: I run into errors too, but I haven't set it up and I am not using wandb; possibly for others that have properly set it up there are no errors. That said, IMO, it's a better user experience if that's **_disabled by default_** so that users not using wandb do not get errors, while the flag is still there (set to `False`) for discoverability. 

TL;DR: +1 let's file a bug and have it `False` by default everywhere.

src/oumi/evaluation/lm_harness.py
kaisopos: I run into issues, let's discuss offline.

src/oumi/core/configs/params/model_params.py
```diff
@@ -186,18 +186,19 @@ class ModelParams(BaseParams):
     other parts fixed.
     """
 
-    def to_lm_harness(self) -> dict[str, Any]:
+    def to_lm_harness(self, vllm_engine=False) -> dict[str, Any]:
```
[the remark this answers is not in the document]
kaisopos: ouch - circular dependency. 

`inference_config.py` (i) defines `InferenceEngineType` and (ii) imports ModelParams (which with this PR import `InferenceEngineType`). 

Any suggestions to fix this cleanly?
Should I move `InferenceEngineType` out of inference_config?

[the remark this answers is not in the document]
xrdaukar: Probably. No objection for moving the enum into some narrower scope

configs/recipes/smollm/evaluation/135m/eval.yaml
xrdaukar: SG. If you can still reproduce these errors while testing this change, please file a bug with repro steps  (I don't recall exact details )

xrdaukar (approved):

src/oumi/core/configs/params/model_params.py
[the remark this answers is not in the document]
kaisopos: fixed @ https://github.com/oumi-ai/oumi/pull/1395

src/oumi/evaluation/lm_harness.py
xrdaukar: Some params (`convert_img_format`, `image_token_id`) are specific to `hf-multimodal` . There are some examples and documentation here: https://github.com/EleutherAI/lm-evaluation-harness/releases/tag/v0.4.5
for `hf-multimodal` vs `vllm-vlm`

kaisopos: Thanks - looking into this now!

Issue #1382, opened 2025-02-05
[Feature] Add Slurm support in Oumi Launch


taenin:
### Feature request

We should add support for sshing into a slurm head node and kicking off jobs. This is great for local and dedicated clusters.

### Motivation / references

This greatly simplifies workflows with Oumi when working on local clusters that are already running slurm.

### Your contribution

I will contribute this feature.

PR #1389, opened 2025-02-06
Create a client for communicating with a Slurm node via SSH.

taenin:
# Description

Establish a client for communicating with a Slurm head node. This is the first PR in a series that will add Slurm support in `oumi launch`.

We launch jobs with
- `sbatch ...`

List jobs with
- `sacct ...`

And cancel jobs with
- `scancel ...`

Verified on our local Slurm cluster.

## Related issues

Towards #1382

## Before submitting

## Reviewers


src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+@dataclass
+class SlurmResponse:
+    """A response from Slurm."""
+
+    stdout: str
+    stderr: str
+    exit_code: int
+
+
+def retry_auth(user_function):
```
wizeng23, line 59: any typing you can add here?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            user: The user to act as.
+            slurm_host: The Slurm Head Node to connect to.
+            cluster_name: The name of the cluster this client communicates with.
+        """
+        self._user = user
+        self._slurm_host = slurm_host
+        self._cluster_name = cluster_name
+        self._refresh_creds()
+
+    def _is_job_done(self, job_state: str) -> bool:
```
wizeng23, line 86: nit: since this is a static function, you could move outside the class? Makes it clear it doesn't depend on anything from the class

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            ssh_cmd,
+            shell=True,
+            capture_output=True,
+            timeout=10,
+        )
+    except subprocess.TimeoutExpired:
+        raise _SlurmAuthException("Timeout while checking connection.")
+    if child.returncode == 0:
+        return
+    raise _SlurmAuthException("Connection to Slurm host is closed.")
```
xrdaukar, line 47: do we want to include child output to error message (or, at least , log it) on error to help with debugging ?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            raise ValueError(f"Expected 5 fields, but found {len(column_lengths)}.")
+        fields = []
+        for i in range(len(column_lengths)):
+            start = sum(column_lengths[:i]) + i
+            end = start + column_lengths[i]
+            fields.append(line[start:end].strip())
+        return JobStatus(
+            id=fields[0],
+            name=fields[1],
+            status=fields[3].split(" ")[0],
```
wizeng23, line 143: Could you explain this splitting logic? Could give an example of the before and after, for instance

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            timeout=10,
+        )
+    except subprocess.TimeoutExpired:
+        raise _SlurmAuthException("Timeout while checking connection.")
+    if child.returncode == 0:
+        return
+    raise _SlurmAuthException("Connection to Slurm host is closed.")
+
+
+@dataclass
```
xrdaukar, line 50: (out of scope): both `pydantic` and `dataclasses` are used in oumi  codebase.  Should we try to converge to just one of them, or at least establish some guidance when to use one or another ?  In this case, `NamedTuple` can also be used

@oelachqar @taenin

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            "NODE_FAIL",
+            "OUT_OF_MEMORY",
+            "PREEMPTED",
+            "TIMEOUT",
+            "SUSPENDED",
+            "STOPPED",
+        }
+        return job_state in terminal_states
+
+    def _split_status_line(
```
wizeng23, line 113: ditto, though you'd have to add a cluster_name arg. This function does some non-trivial string parsing, so IMO a small benefit of having it split out is being easier to unit test.

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+        Outputs only the job id number and the cluster name if present.
+        The values are separated by a semicolon. Errors will still be displayed.
+
+        Args:
+            sbatch_output: The result of sbatch
+
+        Returns:
+            The job ID.
+        """
+        return sbatch_output.split(";")[0]
```
xrdaukar, line 218: would it be useful to do manual validations on sbatch_output, and raise `ValueError` including what's wrong, and maybe including sbatch_output  value as `str`

e.g., is the absence of `;` an error?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+    ) -> JobStatus:
+        """Splits a status line into a JobStatus object.
+
+        The expected order of job fields is:
+        0. Job ID
+        1. Job Name
+        2. User
+        3. Job State
+        4. Job State Reason
+
```
wizeng23, line 124: For non-trivial string parsing logic, I personally like having an example input string in the comments somewhere. As a reader, it's harder to fully follow the logic if I don't know what it's parsing. For example, the `+i` seems a bit confusing when assigning start, but I just make the assumption that there's whitespace or something similar in between the entries.

Consider also adding a comment why a split() call doesn't work here

EDIT: I see now that there's an example string in sacct.txt. You could consider pointing to that in this file wherever i left a comment about having an example string

xrdaukar (approved):

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+        Returns:
+            A list of users.
+        """
+        # List all active users with an open SSH tunnel to the Slurm head node.
+        command = f"ls ~/.ssh/ | egrep 'control-{slurm_host}-.*-.*'"
+        result = subprocess.run(command, shell=True, capture_output=True)
+        if result.returncode != 0:
+            return []
+        ssh_tunnel_pattern = rf"control-{slurm_host}-[^-]*-(.*)"
```
wizeng23, line 192: ditto, consider adding an example string so that this logic is less opaque. For me it's hard to understand regex code without it

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+        lines = result.stdout.decode("utf-8").strip().split("\n")
+        users = set()
+        for line in lines:
+            match = re.match(ssh_tunnel_pattern, line.strip())
+            if match:
+                users.add(match.group(1))
+        return list(users)
+
+    def _compute_duration_debug_str(self, start_time: float) -> str:
```
wizeng23, line 201: ditto, factor out, or make static?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        response_format = "JobId%-30,JobName%30,User%30,State%30,Reason%30"
+        command = f"sacct --user={self._user} --format='{response_format}'"
+        result = self.run_commands([command])
+        if result.exit_code != 0:
+            raise RuntimeError(f"Failed to list jobs. stderr: {result.stderr}")
+        # Parse STDOUT to retrieve job statuses.
+        lines = result.stdout.strip().split("\n")
+        jobs = []
+        if len(lines) < 4:
```
wizeng23, line 308: Ditto, consider adding comments to explain logic when doing non-trivial string parsing logic

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -0,0 +1,400 @@
...
+            job_id: The ID of the job to cancel.
+
+        Returns:
+            The job status if found, None otherwise.
+        """
+        command = f"scancel {job_id}"
+        result = self.run_commands([command])
+        if result.exit_code != 0:
+            raise RuntimeError(f"Failed to cancel job. stderr: {result.stderr}")
+        return self.get_job(job_id)
```
wizeng23, line 347: Is the job guaranteed to be there after you call cancel on it?

wizeng23 (commented):
Could you add a comment mentioning what form of testing has been done?

src/oumi/launcher/clients/slurm_client.py
taenin: If the job exists in any state (queued, running, failed, succeeded, etc) we will return that status. If not, `get_job()` will throw an exception, which is intended.

src/oumi/launcher/clients/slurm_client.py
taenin: I can add minimal typing, but it's a generic decorator so `Callable` will have to suffice.

src/oumi/launcher/clients/slurm_client.py
taenin: Done.

src/oumi/launcher/clients/slurm_client.py
taenin: Since it's a private method I won't plan on unit testing this method either way. It'll be tested in the function that calls it.

src/oumi/launcher/clients/slurm_client.py
taenin: Added an example.

src/oumi/launcher/clients/slurm_client.py
taenin: Added.

src/oumi/launcher/clients/slurm_client.py
taenin: Added

src/oumi/launcher/clients/slurm_client.py
taenin: Moved out.

src/oumi/launcher/clients/slurm_client.py
taenin: The semi colon will only be included if Slurm is running with a cluster name. Outputting `jobId` is valid. 
I've added some validation to split on `;` and return an error if the length of the split is > 2.

src/oumi/launcher/clients/slurm_client.py
taenin: Added

src/oumi/launcher/clients/slurm_client.py
taenin: Added

PR #1396, opened 2025-02-06
Train+Inference with Qwen 2.5 VL (3B)

optas:
# Description

This PR adds the minimal changes needed to train and run inference with the newer Qwen VL model (2.5v) inside Oumi; with single image and multi-turn text data.

Worth noting:
 - Training was tested with both a single and four A:100 GPUs (the latter via `DataParallel` mode).
 - Training was tested with `spda` and `flash_attention_2` attention mechanisms. The latter had approx. **12%** more `train_tokens_per_second'.
 - [**Important**]   Qwen2.5 is integrated on the latest transformers **dev** version (as of 02/05/25):
      -- Specifically, Oumi has tested this model with transformers 4.49.0.dev0
      -- Upgrading to this version seems necessary to train/use this model to avoid encountering `KeyError: 'qwen2_5_vl'
` , but it might break other Oumi utilities, which at the time is _fully_ tested to be compatible with `transformers>=4.48.0,<4.49`.

Towards: OPE-988

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


configs/recipes/vision/qwen2_5_vl_3b/sft/gcp_job.yaml
```diff
@@ -0,0 +1,70 @@
...
+  ~/.netrc: ~/.netrc # WandB credentials
+
+  # Mount HF token, which is needed to download locked-down models from HF Hub.
+  # This is created on your local machine by running `huggingface-cli login`.
+  ~/.cache/huggingface/token: ~/.cache/huggingface/token
+
+envs:
+  WANDB_PROJECT: oumi-train
+  OUMI_RUN_NAME: qwen2_5-vl.fft.oumi
+  ACCELERATE_LOG_LEVEL: info
```
xrdaukar, line 36: please remove ACCELERATE_LOG_LEVEL , TOKENIZERS_PARALLELISM vars. There was recent cleanup by @wizeng23  (I believe Launcher will set these automatically)

scripts/benchmarks/minimal_multimodal_training.py
```diff
@@ -169,7 +175,7 @@ def test_multimodal_trainer(
     else:
         print("Not initializing distributed process group")
 
-    if model_name == ModelName.QWEN2_VL and batch_size != 1:
+    if model_name in [ModelName.QWEN2_VL, ModelName.QWEN2_5_3B_VL] and batch_size != 1:
```
xrdaukar, line 178: nit: let's use tuple instead of array  ` ... in (ModelName.QWEN2_VL, ModelName.QWEN2_5_3B_VL)`
It's arguably cleaner because the list is supposed to be  immutable

src/oumi/core/configs/internal/supported_models.py
```diff
@@ -148,6 +149,31 @@ def _create_qwen2_vl_vlm_config() -> InternalModelConfig:
     return config
 
 
+def _create_qwen2_5_vl_vlm_config() -> InternalModelConfig:
+    config = _create_default_vlm_config(pixel_values_variable_shape=True)
+    config.chat_template = "qwen2.5-vl-instruct"
+    config.padding_side = InternalPaddingSide.PAD_RIGHT
```
xrdaukar, line 155: Do we need to set this just for this model ? What is the default: "left" ? Please add a comment why we're doing it

src/oumi/datasets/chat_templates/qwen2.5-vl-instruct.jinja
```diff
@@ -0,0 +1,37 @@
+{% set image_count = namespace(value=0) %}
+{% set video_count = namespace(value=0) %}
+
+{%- for message in messages -%}
+    {%- if loop.first and message['role'] != 'system' -%}
+        {{ '<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n' }}
+    {%- endif -%}
```
xrdaukar, line 7: is this template different from `qwen-vl-instruct`? If not let's just use the existing one to reduce maintenance cost

src/oumi/core/configs/internal/supported_models.py
```diff
@@ -148,6 +149,31 @@ def _create_qwen2_vl_vlm_config() -> InternalModelConfig:
     return config
 
 
+def _create_qwen2_5_vl_vlm_config() -> InternalModelConfig:
```
xrdaukar, line 152: this function is almost the same as `_create_qwen2_vl_vlm_config()` . 

Can we do this: 

```
def _create_qwen2_5_vl_vlm_config() -> InternalModelConfig:
  config = _create_qwen2_vl_vlm_config()
  # Override few params specififc to 2.5
   config.foo= ...
   return config
```

src/oumi/core/configs/internal/supported_models.py
```diff
@@ -148,6 +149,31 @@ def _create_qwen2_vl_vlm_config() -> InternalModelConfig:
...
+            )
+            for feature_name in ("image_grid_thw",)
+        }
+    )
+    config.processor_kwargs.update(
+        # Defaults per Qwen2.5-VL:
+        # https://github.com/QwenLM/Qwen2.5-VL/blob/main/qwen-vl-utils/src/qwen_vl_utils/vision_process.py # noqa: E501
+        {
+            "min_pixels": 4 * 28 * 28,
+            "max_pixels": 16384 * 28 * 28,
```
xrdaukar, line 171: is increasing  `max_pixels` important/essential ? 

Context: CUDA OOM-s for `HuggingFaceH4/llava-instruct-mix-vsft`: https://github.com/oumi-ai/oumi/pull/908

The reduced value for 2.0 solves that - but may affect quality. We may have to look into that and adjust for 2.0 as well

configs/recipes/vision/qwen2_5_vl_3b/sft/train.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - dataset_name: "merve/vqav2-small"
+        split: "validation"
+        shuffle: True
+        seed: 42
+        transform_num_workers: "auto"
+        dataset_kwargs:
+          processor_name: "Qwen/Qwen2.5-VL-3B-Instruct"
+          # limit: 4096 # Uncomment to limit dataset size!
+          return_tensors: True
+      # - dataset_name: "HuggingFaceM4/Docmatix"
```
xrdaukar, line 36: Let's remove the commented out  `Docmatrix` entry here. I plan to remove it from otehr configs as well (the dataset is not quiet ready yet)

xrdaukar:
Please update PR title (it's not just serving)

src/oumi/core/configs/internal/supported_models.py
xrdaukar: if this investigation requires non-insignificant effort, let's move this hparams tuning into separate sub-task

tests/unit/datasets/test_chat_templates.py
```diff
@@ -161,6 +161,12 @@ def get_hf_chat_template(
         test_image=True,
         image_placeholder="<|vision_start|><|image_pad|><|vision_end|>",
     ),
+    ChatTemplateTestSpec(
+        chat_template_name="qwen2.5-vl-instruct",
+        model_name="Qwen/Qwen2.5-VL-3B-Instruct",
```
xrdaukar, line 166: this model is unusable during testing because it requires building `transformers` from source 

Let's plug `Qwen/Qwen2-VL-2B-Instruct` temporarily, and add a comment about this

xrdaukar (approved):

PR #1411, opened 2025-02-08
Add requirements header to configs and clean them up

wizeng23:
# Description

This adds a Requirements section to the top of all configs. This explicitly lists out the steps required to get the config to work, aside from installing Oumi.

Comprehensive list of all requirements added:
- Log into HF for configs requiring a HF token
- Log into WandB if needed
- Set up Skypilot GCP for GCP configs
- Set up ALCF and the username for Polaris configs, and call out that Polaris is only for the Oumi core team (to not confuse users)
- Request access to required gated models/datasets (ex. Llama)
- Install relevant packages (SGLang, flash-attn, vLLM) for non-job configs (ex. SGLang inference configs)
- Optionally mount buckets if we mention it in the config

Miscellaneous config comments cleanup:
- Rename my-oumi-cluster to something unique to prevent collisions
- Delete redundant comments
- Fix comment on where Polaris working_dir is mounted
- Create more detailed description for files missing it.

Config cleanup:
- Delete `adapter_model: null`. This has no functional effect.

For reviewers: This PR only updates comments, aside from deleting `adapter_model: null` which is a no-op. Feel free to take a look at just a few files to get the general idea; no need to review them all. I manually double-checked every change in this PR.

## Related issues

Fixes OPE-1022
Fixes OPE-1001

## Before submitting


configs/recipes/llama3_1/evaluation/8b_polaris_job.yaml
```diff
@@ -16,7 +21,6 @@ user: your_username
 resources:
   cloud: polaris
 
-# Upload working directory to /home/$USER/oumi_launcher/llama8b_eval.
```
taenin, line 19: Any reason to remove these comments on all polaris jobs?

configs/recipes/smollm/evaluation/135m/quickstart_alpaca_v2_eval.yaml
```diff
@@ -22,11 +23,15 @@ model:
...
-  batch_size: 4
+  # NOTE: Adjust based on your available GPU memory.
+  batch_size: 64
 
 tasks:
-  - evaluation_platform: alpaca_eval
+  - evaluation_platform: alpaca_eval # 805 examples.
+    # NOTE: We use a small subset for demonstration purposes, and to reduce the cost of
+    # OpenAI inference. Comment out for full evaluation.
+    num_samples: 3
```
taenin, line 33: This is changing more than just comments

configs/recipes/llama3_1/evaluation/8b_polaris_job.yaml
wizeng23: iirc it's not correct anymore because the directory is timestamped instead of named after the job

configs/recipes/smollm/evaluation/135m/quickstart_alpaca_v2_eval.yaml
wizeng23: Good catch, will move elsewhere!

configs/recipes/llama3_1/evaluation/8b_polaris_job.yaml
taenin: Consider correcting it instead of removing it:
`# Upload working directory to /home/$USER/oumi_launcher/{submission_time}`

The other comment you removed in this file should be correct, though. That one I don't think we should remove.

wizeng23: Done!

taenin (approved):

configs/recipes/llama3_3/evaluation/70b_gcp_job.yaml
```diff
@@ -1,5 +1,11 @@
 # Job config to eval Llama 3.3 70B Instruct on GCP.
 #
+# Requirements:
```
xrdaukar, line 3: Any plans to create some sort of automation for these types of config comments? 

I fear it's going to be near impossible  to keep such comments coherent manually over time in all configs

xrdaukar (approved):

configs/recipes/llama3_3/evaluation/70b_gcp_job.yaml
wizeng23: No plans for now. The aim of this PR is to cover all possible requirements for our existing configs. If I'm missing any, please LMK! My hope for the future is that with all configs following this convention, people creating new configs will adhere to the existing practice if they reference existing configs. Adding requirements for a new config should be much easier compared to backfilling old configs.

kaisopos (approved):

PR #1359, opened 2025-02-04
Define `OUMI_USE_SPOT_VM` env var and start using it to override `use_spot` param 

xrdaukar:
# Description

-- If `OUMI_USE_SPOT_VM` is unspecified, empty, or set as "config"  then no-change : use the value from Job config
-- If `OUMI_USE_SPOT_VM` is set to `spot` or `preemptable` then launch SpotVM.
-- If `OUMI_USE_SPOT_VM` is set to `non-spot` or `non-preemptable` (dash and case are ignored) then launch non-SpotVM.

## Related issues

Fixes OPE-976

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [X] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [X] Did you link the issue(s) related to this PR in the section above?
- [ ] Did you add / update tests where needed?

## Reviewers

At least one review from a member of `oumi-ai/oumi-staff` is required.

PR #1420, opened 2025-02-11
Add `OUMI_FORCE_EDITABLE_INSTALL` env var to do editable Oumi install from source in job configs

wizeng23:
# Description

Inspired by #1359. Our configs install oumi from PyPi for convenience for external users, but it's more friction for developers. If `OUMI_FORCE_EDITABLE_INSTALL` is set to a truthy value ("true", "yes", "1", etc.), and we're running a dev build of oumi, we will attempt to modify `pip install oumi` commands in the setup section of job configs to install editable from source instead (pip install oumi[gpu] -> pip install -e '.[gpu]'). The logic tries to handle many possible edge cases, but may not be perfect. Hopefully this should be fine since it's mainly an experimental flag for development.

This PR also modifies the dev GCP job to do an editable install from source by default. It also adds documentation for the dev setup guide.

Tested that this works on a GCP job.

## Further justification

This PR is helpful for anyone making core code changes to Oumi that can't be tested locally, ex. it affects a large model. The pain points of doing this currently are:
- Need to manually modify target configs to do editable install. This is easy to forget (at least for me) since we've only recently started installing from PyPi.
- When making a commit, the above all modifications have to be undone first, then added back later if the PR isn't complete.

Pros of this PR:
- Allow devs to have editable install by default. IMO the only time I'd use the PyPi package is to repro a user-reported issue.
- Since the logic is gated by an env var, it shouldn't affect anyone who doesn't enable it

Cons of this PR:
- Potentially error prone, as it's hard to cover all possible edge cases.
- Code bloat, as this may not be useful for the majority of users. It's fairly isolated from the main body of code though.

## Related issues

Fixes OPE-1043.

## Before submitting


taenin (commented):
My 2 cents for this change:
1) We should avoid using an environment variable for this.
2) We should make this a change that's limited to only files in the CLI (we shouldn't touch job_config.py)

I'd recommend a change similar to https://github.com/oumi-ai/oumi/blob/main/src/oumi/cli/launch.py#L53

wizeng23:
Agree with 2. For 1, while I like your approach in the code you linked, I'm wondering if it doesn't work as well in this case. I imagine two cases:
1. We assume that everyone using a dev version of Oumi wants to enact this logic. However, what if that's not the case for some reason, and they have no way to disable this logic?
2. We add a CLI prompt every time the user kicks off a job, which feels like too much friction to me. We could save the answer after they respond for the first time, but to me that feels equivalent to the env var approach. I guess to me, the difference is that for your working_dir code, the prompt would be to catch an edge case, whereas here, this is a core question that devs face when using the job launcher.

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
...
+    for i, line in enumerate(setup_lines):
+        # Skip comments.
+        if line.strip().startswith("#"):
+            continue
+
+        # Only consider lines with the words "pip" "install" "oumi" in that order.
+        # This is to try to handle cases where those words may appear multiple times
+        # in the line for some reason, but this logic may not be perfect.
+        pip_idx = line.find("pip")
+        install_idx = line.find("install", pip_idx)
```
xrdaukar, line 160: Would it make sense to switch to regexp  for this kind of parsing ?

src/oumi/core/configs/job_config.py
```diff
@@ -160,3 +166,16 @@ class JobConfig(BaseConfig):
...
+    def __finalize_and_validate__(self):
+        """Finalizes and validates the configuration."""
+        # (experimental) If the OUMI_TRY_EDITABLE_INSTALL env var is set to a truthy
+        # value, attempt to modify the setup script in the job config to install Oumi in
+        # editable mode from source, as opposed to installing from PyPI.
+        if get_editable_install_override() and self.setup:
+            logger.info(
+                "OUMI_TRY_EDITABLE_INSTALL detected! Attempting to modify job config's "
+                "`setup` to install Oumi in editable mode from source..."
+            )
```
xrdaukar, line 180: should we also do it un `run:` section ?

src/oumi/utils/str_utils.py
wizeng23: Yep! I listed it as a todo in the PR description. I could either do it as part of this PR or leave it as future work if reviewers are ok with that

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
...
+        # Ex. oumi[gpu,dev] -> '.[gpu,dev]'
+        local_install = f"'.{oumi_install[4:]}'"
+        line = line.replace(oumi_install, local_install)
+        # Add the editable flag if not already present.
+        # If already present, the flag should be after "install" and before the packages
+        # to install, i.e. oumi in this case.
+        edit_flag_idx = line.find("-e", install_idx)
+        # If the editable flag is not present or is after the oumi install, add it.
+        if edit_flag_idx == -1 or edit_flag_idx > oumi_idx:
+            line = line[:install_idx] + line[install_idx:].replace(
```
xrdaukar, line 186: Can this replace replace 2+ occurrences of the word "install"  than needed  ?  `pip install ... # install install `

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
...
+        if not (pip_idx != -1 and install_idx != -1 and oumi_idx != -1):
+            continue
+        # Find the last occurrences of "pip" and "install", in case one line has
+        # multiple pip installs, ex. `pip install uv && uv pip install oumi`.
+        install_idx = line.rfind("install", 0, oumi_idx)
+        pip_idx = line.rfind("pip", 0, install_idx)
+
+        # The oumi import could be followed by optional dependencies, ex. "oumi[gpu]".
+        # We need to extract the exact oumi install string.
+        oumi_end_idx = oumi_idx + 4
```
xrdaukar, line 171: nit: Replace magic number 4 with `len("oumi")` ?

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
...
+        # to install, i.e. oumi in this case.
+        edit_flag_idx = line.find("-e", install_idx)
+        # If the editable flag is not present or is after the oumi install, add it.
+        if edit_flag_idx == -1 or edit_flag_idx > oumi_idx:
+            line = line[:install_idx] + line[install_idx:].replace(
+                "install", "install -e"
+            )
+        # Replace the line in the setup script.
+        logger = logging.getLogger("oumi")
+        logger.info(f"Detected the following oumi installation: `{setup_lines[i]}`")
```
xrdaukar, line 191: optional: Ideally, would be good to also detect editable installs, and says something like  "already editable install detected, no change"

xrdaukar (approved):
LGTM.  Please follow-up with other reviewers who may have more context  here to resolve remaining questions.

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
     # This is inefficient: allocates a temporary copy of string content.
     # FIXME Can we do better?
     return len(s.encode("utf-8"))
+
+
+def get_editable_install_override() -> bool:
+    """Returns whether OUMI_TRY_EDITABLE_INSTALL env var is set to a truthy value."""
+    s = os.environ.get("OUMI_TRY_EDITABLE_INSTALL", "")
```
taenin, line 127: Let's add one additional check here:
This condition should only return true if the current installation of oumi is also a dev build. Specifically, you should call this method: https://github.com/oumi-ai/oumi/blob/main/src/oumi/utils/version_utils.py#L19

src/oumi/core/configs/job_config.py
taenin: +1

src/oumi/utils/str_utils.py
taenin: Let's make this a regex now.

src/oumi/utils/str_utils.py
```diff
@@ -119,3 +120,75 @@ def compute_utf8_len(s: str) -> int:
...
+    mode = s.lower().strip()
+    bool_result = try_str_to_bool(mode)
+    if bool_result is not None:
+        return bool_result
+    return False
+
+
+# Experimental function, only for developer usage.
+def set_oumi_install_editable(setup: str) -> str:
```
taenin, line 136: Definitely use a regex for this logic. If you use a regex, this method can be just a few lines long.

taenin (changes requested):

wizeng23:
Appreciate everyone's detailed feedback! After my oncall, I'll switch to regex and address everyone's feedback before switching out of draft mode.

xrdaukar:
> Appreciate everyone's detailed feedback! After my oncall, I'll switch to regex and address everyone's feedback before switching out of draft mode.

+1 Would be real good to get this  in   in some form. Otherwise testing local changes against existing configs is a pain and error-prone .

src/oumi/utils/str_utils.py
wizeng23: No longer an issue with the regex

src/oumi/utils/str_utils.py
wizeng23: Ditto

src/oumi/utils/str_utils.py
wizeng23: That would require writing a separate regex for that case. If you still think it'd be helpful, let me know and I can add it. We have very few configs with an editable install, so IMO it isn't too important.

src/oumi/core/configs/job_config.py
```diff
@@ -161,3 +168,27 @@ class JobConfig(BaseConfig):
...
+        # Oumi from PyPI.
+        if get_editable_install_override_env_var() and is_dev_build():
+            logger.info("-" * 80)
+            logger.info(
+                "OUMI_FORCE_EDITABLE_INSTALL detected! Attempting to modify job "
+                "config to install Oumi in editable mode from source..."
+            )
+            if self.setup:
+                logger.info("Modifying setup script...")
+                self.setup = set_oumi_install_editable(self.setup)
```
nikg4, line 190: nit: consider comparing `self.setup` value before and after the function call, and adding log message like "modified" , "no changes". (you could move `logger.info()` after the call the function is near instant , not much value is writing "Starting ..." message)

nikg4: nvm. The function already does the logging

nikg4 (approved):

src/oumi/core/configs/job_config.py
wizeng23: Yep! I wanted to log on a per-line basis so it's clear which line was modified.

PR #1421, opened 2025-02-11
Update `save_conversations` tool

xrdaukar:
# Description

-- Add flags to split multi-turn conversations into N single-turn conversations
-- Modify `Conversation.filter_messages()` method to accept user-defined  predicate.

## Related issues

Towards OPE-355

## Before submitting

## Reviewers

### DIFF
--- a/scripts/datasets/save_conversations.py
+++ b/scripts/datasets/save_conversations.py
@@ -113,12 +113,12 @@ def _not_assistant_fn(m: Message):
 
 def _process_conversation(
     input_conversation: Conversation,
-    convert_multi_turn_to_single_turn: bool,
+    split_multi_turn_to_single_turn: bool,
     drop_assistant_messages: bool,
 ) -> list[Conversation]:
     result: list[Conversation] = []
-    if convert_multi_turn_to_single_turn:
-        proto_conversation = input_conversation
+    if split_multi_turn_to_single_turn:
+        proto_conversation = copy.copy(input_conversation)
         proto_conversation.messages = []
         for turn_messages in _split_message_list_on_user(input_conversation.messages):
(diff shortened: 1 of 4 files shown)


scripts/datasets/save_conversations.py
```diff
@@ -65,13 +66,83 @@ def _load_sft_dataset(
     return dataset
 
 
+def _split_message_list_on_user(messages: list[Message]) -> list[list[Message]]:
+    turn_start_indices: list[int] = []
```
taenin, line 70: Food for thought here: should you add a check for an empty input? I think `_split_message_list_on_user([])` should return `[]` not `[ [] ]`

src/oumi/core/types/conversation.py
```diff
@@ -391,15 +389,23 @@ def last_message(self, role: Optional[Role] = None) -> Optional[Message]:
...
+        *,
+        role: Optional[Role] = None,
+        filter_fn: Optional[Callable[[Message], bool]] = None,
+    ) -> list[Message]:
         """Gets all messages in the conversation, optionally filtered by role.
 
         Args:
             role: The role to filter messages by.
-                If None, returns all messages.
+                If None, no filtering by role is applied.
```
taenin, line 405: This implies that the filter_fn has to use to `role` field, which isn't necessarily true.

taenin (approved):

scripts/datasets/save_conversations.py
xrdaukar: right, updated!

src/oumi/core/types/conversation.py
xrdaukar: Added Optional to make it more clear

PR #1435, opened 2025-02-14
Create `pad_to_max_dim_and_stack()` function in torch_utils

xrdaukar:
# Description

-- The function pads input tensors to maximum size in each dimension, then stacks them
-- This is similar to `pad_sequences()` except it can pad to more than one dimension, which is needed e.g., for `cross_attention_mask` which has two variable dimensions: `seq_len`, `num_images` for multi-image multimodal examples
-- The integration of this new function into data pipeline is coming in a separate PR.
-- Minor error logging improvements in padding utils

## Related issues

Towards OPE-355

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/torch_utils.py
+++ b/src/oumi/utils/torch_utils.py
@@ -490,7 +490,7 @@ def _get_dims_min_max_size(tensors_list: list[torch.Tensor]) -> list[_DimMinMaxS
     ]
 
 
-def _pad_to_max_dim_and_stack(
+def _pad_to_max_dim_and_stack_impl(
     tensors_list: list[torch.Tensor],
     *,
     padding_value: float = 0,
@@ -579,7 +579,7 @@ def pad_to_max_dim_and_stack(
     input_tensors = convert_to_list_of_tensors(tensors_list)
 
     try:
-        return _pad_to_max_dim_and_stack(
+        return _pad_to_max_dim_and_stack_impl(
             input_tensors,
             padding_value=padding_value,
             pad_on_left_side=pad_on_left_side,
--- a/src/oumi/utils/torch_utils.py
+++ b/src/oumi/utils/torch_utils.py
@@ -354,9 +354,16 @@ def convert_to_list_of_tensors(values: list[T]) -> list[torch.Tensor]:
 def _pad_sequences_impl(
     sequences: list[torch.Tensor], *, padding_value: float = 0
 ) -> torch.Tensor:
-    return torch.nn.utils.rnn.pad_sequence(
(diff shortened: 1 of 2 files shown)


src/oumi/utils/torch_utils.py
```diff
@@ -441,6 +448,155 @@ def pad_sequences(
...
+                if curr_size < max_size:
+                    target_view = target_view.narrow(dim_idx, start=0, length=curr_size)
+
+        assert target_view.shape == input_tensor.shape
+        target_view[...] = input_tensor
+
+    return result
+
+
+def pad_to_max_dim_and_stack(
```
taenin, line 541: It's a bit confusing that you have a private method with the exact same name without the leading underscore. Consider changing `_pad_to_max_dim_and_stack` -> `_pad_to_max_dim_and_stack_impl`?

src/oumi/utils/torch_utils.py
```diff
@@ -441,6 +448,155 @@ def pad_sequences(
...
+    tensors_list: list[T],
+    *,
+    padding_value: float = 0,
+    padding_side: Optional[str] = None,
```
taenin, line 545: Would the typing `Literal["left", "right"] | None` make more sense here?

taenin (approved):

src/oumi/utils/torch_utils.py
xrdaukar: updated

src/oumi/utils/torch_utils.py
xrdaukar: I tried to do it but it leads to complications e.g., some callers store `padding_side` as  unvalidated `str`, which would make it difficult to use this function, so I decided to keep it as is

src/oumi/utils/torch_utils.py
```diff
@@ -441,6 +448,155 @@ def pad_sequences(
...
+    return [
+        _DimMinMaxSizes(
+            dim_index=idx, min_size=min_dim_sizes[idx], max_size=max_dim_sizes[idx]
+        )
+        for idx in range(num_dims)
+    ]
+
+
+def _pad_to_max_dim_and_stack(
+    tensors_list: list[torch.Tensor],
```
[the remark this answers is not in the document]
xrdaukar: Good point.  I think it should be a larger sweep over Oumi codebase if we decide to do that: many functions marked as `torch.Tensor`  may work inefficiently or break (?) if sparse tensors is passed. To my knowledge we don't have any `non-sparse` typing constraints anywhere , and don't really use sparse tensors.  Anyhow, maybe good idea to open a Linear task  to do something about it..

PR #1441, opened 2025-02-18
Create a script to calculate memory used during training

wizeng23:
# Description

This is a Python script that will estimate the memory needed to run training. This is derived from the memory calculator Google Sheet Jeremy work. In the future, this can be integrated as a subcommand for the CLI, but leaving it as a separate script for now to reduce visibility as it's a WIP. The initial implementation works for GPT2.

## Related issues

Towards OPE-672

## Before submitting


scripts/memcalc.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def get_gradient_bytes(config: TrainingConfig, model_bytes: int) -> int:
+    """Gets the total number of bytes used by gradients."""
+    if config.training.gradient_accumulation_steps > 1:
+        return model_bytes * 2
+    return model_bytes
+
+
+def get_activation_bytes(
```
oelachqar, line 152: Should we have a some of the core logic as a library under oumi.performance? The script would just call that.

In the future we could also have an `oumi train -c config --dry-run` that uses it

wizeng23: Sure! And I was just thinking about adding a dry-run flag as well :D I think that's the cleanest way to integrate it for sure.

scripts/memcalc.py
```diff
@@ -0,0 +1,230 @@
...
+
+    hidden_dim: int
+    """The hidden dimension aka embedding size."""
+
+    num_kv_heads: int
+    """The number of key value (KV) heads for the attention layer."""
+
+
+def get_bytes_per_unit(dtype: torch.dtype) -> int:
+    """Gets the number of bytes used per memory unit given the torch dtype.
```
xrdaukar, line 51: there is similar  helper function in torch_utils: https://github.com/oumi-ai/oumi/blob/a4de2020a207eb994151bf7cc981e234f84438f8/src/oumi/utils/torch_utils.py#L259

Let's converge to torch_utils version

scripts/memcalc.py
```diff
@@ -0,0 +1,230 @@
...
+        return 4
+    elif dtype == torch.bfloat16 or dtype == torch.float16 or dtype == torch.int16:
+        return 2
+    elif dtype == torch.uint8:
+        return 1
+    else:
+        raise ValueError(f"Unsupported torch dtype: {dtype}")
+
+
+def bytes_to_str(bytes: Union[int, float]) -> str:
```
xrdaukar, line 68: nit: bytes is a also a built-in Python type. Can we rename to something like num_bytes?

scripts/memcalc.py
```diff
@@ -0,0 +1,230 @@
...
+    A memory unit is essentially a number in memory, such as a parameter, gradient, or
+    activation.
+    """
+    if dtype == torch.float64 or dtype == torch.int64:
+        return 8
+    elif dtype == torch.float32 or dtype == torch.int32:
+        return 4
+    elif dtype == torch.bfloat16 or dtype == torch.float16 or dtype == torch.int16:
+        return 2
+    elif dtype == torch.uint8:
```
jgreer013, line 62: FP8?

taenin (commented):
Overall this change seems reasonable. +1 to Nikolai's comment on converting to our torch_utils where appropriate for this integration.

scripts/memcalc.py
wizeng23: Right now, the script only supports GPT2 configs. Once it's a bit more fleshed out, I can integrate it like you mentioned.

oelachqar (approved):

nikg4 (approved):

taenin (approved):

PR #1448, opened 2025-02-19
Support for multi-image VLM training

xrdaukar:
# Description

-- Update collators to support multiple variable dimensions
-- Minor updates and refactor in `scripts/benchmarks/minimal_multimodal_training.py`
-- Add `max_variable_sized_dims` param to  `pad_to_max_dim_and_stack()` in torch utils 
-- Update  `pad_to_max_dim_and_stack()`  to call `pad_sequences` for the common  case when only the first dimension is variable.
-- Add new VLM-specific fields to `InternalModelConfig`-s

## Related issues

Towards OPE-355

## Before submitting

## Reviewers


taenin (approved):

src/oumi/core/collators/vision_language_collator_with_padding.py
```diff
@@ -44,12 +44,15 @@ def __init__(
             `max_length`. Only has effect if `max_length` is specified.
         label_ignore_index:  If set, then label values of tokens that shouldn't
             contribute to the loss computation will be replaced by this special value.
+        allow_multi_image_inputs: Whether to allow-multi-image inputs.
```
wizeng23, line 47:
```suggestion
        allow_multi_image_inputs: Whether to allow multi-image inputs.
```

src/oumi/core/collators/vision_language_collator_with_padding.py
```diff
@@ -107,8 +110,12 @@ def __call__(self, batch) -> dict[str, Any]:
                     other_inputs[input_name].append(item[input_name])
 
             for input_name, values_list in other_inputs.items():
-                tensors_list = convert_to_list_of_tensors(values_list)
-                collated_value = torch.stack(tensors_list)
+                collated_value = pad_to_max_dim_and_stack(
+                    values_list,
+                    max_variable_sized_dims=(
+                        1 if self._allow_multi_image_inputs else 0
```
wizeng23, line 116: Noob question: Why is this 1 and 0, instead of 2 and 1 like above? Is it because this is for collating images, which only have one variable dimension? If you think it's helpful to the reader, consider adding a comment wherever you set `max_variable_sized_dims` to explain the values

src/oumi/utils/torch_utils.py
```diff
@@ -558,6 +578,8 @@ def pad_to_max_dim_and_stack(
 
     Args:
         tensors_list: list of tensors with potentially .
+        max_variable_sized_dims: Maximum number of variable-sized dimensions.
+            Negative values mean `Unlimited`.
         padding_value: value for padded elements. Default: 0.
         padding_side: side to apply padding to. Valid values:  'right', 'left'.
```
wizeng23, line 584: I know you didn't touch this code, but could you add here that the default value is 'right'?

src/oumi/utils/torch_utils.py
```diff
@@ -558,6 +578,8 @@ def pad_to_max_dim_and_stack(
 
     Args:
         tensors_list: list of tensors with potentially .
+        max_variable_sized_dims: Maximum number of variable-sized dimensions.
```
wizeng23, line 581: Consider mentioning why someone may want to set this to something besides unlimited, and what happens if the number of variable-sized dimensions exceeds this

src/oumi/utils/torch_utils.py
```diff
@@ -493,17 +493,36 @@ def _get_dims_min_max_size(tensors_list: list[torch.Tensor]) -> list[_DimMinMaxS
...
+        raise ValueError(
+            "Too many dimensions with variable size. "
+            f"Got: {num_variable_size_dims} variable size dimensions. "
+            f"Maximum allowed: {max_variable_sized_dims}."
+        )
+
+    if num_variable_size_dims == 0:
         # No need to pad anything, just `stack()`.
         return torch.stack(tensors_list)
+    elif num_variable_size_dims == 1 and dim_sizes[0].has_variable_sizes:
```
wizeng23, line 520: Is this equivalent to the logic below, i.e. would they generate the same result? If so, is there a reason we need this additional logic? I get the value in using prebuilt functions, but I'm worried the code could become more brittle/complex the more dependencies/code branching we add.

wizeng23 (commented):
Is it possible to do any E2E tests (ex. a training job) to verify that existing behavior is unaffected? Are you also able to test multi-image training now, or is additional code changes required to do so?

src/oumi/core/collators/vision_language_collator_with_padding.py
xrdaukar: Added comments

src/oumi/utils/torch_utils.py
xrdaukar: I did it for two reasons:
1. The majority of existing uses-cases will go to pre-built function (less risk, as the other codepath below is new , less tested) 
2. Pre-built function is likely more optimized.

src/oumi/utils/torch_utils.py
xrdaukar: updated

src/oumi/utils/torch_utils.py
xrdaukar: updated

xrdaukar:
> Is it possible to do any E2E tests (ex. a training job) to verify that existing behavior is unaffected? Are you also able to test multi-image training now, or is additional code changes required to do so?

I run few manual tests on GCP and using multimodal benchmarking script locally.

wizeng23 (approved):

PR #1451, opened 2025-02-19
Update oumi launch status to show clusters with no running jobs.

taenin:
# Description

Previously, running `oumi launch status` would omit running clusters if they had no jobs available.

This change lists running clusters, regardless of their child job status.

## Related issues

N/A

## Before submitting

## Reviewers


wizeng23 (approved):
Could you remind me how clusters work for Polaris? Is it treated like a cluster that's always up, and thus will always show up in `oumi launch status`?

taenin:
> Could you remind me how clusters work for Polaris? Is it treated like a cluster that's always up, and thus will always show up in `oumi launch status`?

It's a bit nuanced.

If you already have an existing SSH tunnel to polaris, you will see each queue listed as a cluster for your user.

If you don't have an ssh tunnel active you'll need to either create a tunnel or use `oumi launch up` to have the cluster picked up again. Ideally we add an env-var to handle this like we've done for slurm, but it's not high priority right now.

PR #1452, opened 2025-02-20
Update RemoteParams to no longer require an API URL.

taenin:
# Description

When using the RemoteInferenceEngine, you always want to supply an API URL. However, when using our wrapper engines such as the GeminiInferenceEngine, the base_url is already supplied.

Previously if I tried running `oumi infer` with additional remote_params such as `politeness_policy` or `num_workers` I'd receive an exception that the API URL is not provided.

This change allows setting these params while still respecting the base API URL.

## Related issues

N/A

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -393,9 +393,11 @@ class RemoteInferenceEngine(BaseInferenceEngine):
             output_path = inference_config.output_path
 
         if not remote_params.api_url:
-            remote_params.api_url = self.base_url
+            remote_params.api_url = self._remote_params.api_url or self.base_url
 
-        assert remote_params.api_url
+        # Validate API URL one final time.
+        if not remote_params.api_url:
+            raise ValueError("API URL is required for remote inference.")
         async with semaphore:
             api_input = self._convert_conversation_to_api_input(
                 conversation, generation_params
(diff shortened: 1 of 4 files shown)


src/oumi/inference/remote_inference_engine.py
```diff
@@ -392,6 +392,9 @@ async def _query_api(
             generation_params = inference_config.generation or self._generation_params
             output_path = inference_config.output_path
 
+        if not remote_params.api_url:
+            remote_params.api_url = self.base_url
```
oelachqar, line 396: I think we should use `self._remote_params.api_url` first, and if not set base_url. If none are set, raise a value error

oelachqar: Examples:

```python
# user manually sets a url
default_params = RemoteParams(api_url='custom_url')

engine = MyRemoteEngine(..., remote_params=default_params)
engine.infer(conversations)   # should work

# this should still use `custom_url`, not `self.base_url`, which may not be defined
request_specific_params = RemoteParams(max_retries=10) 
engine.infer(conversations, remote_params=request_specific_params)
```

taenin: Sounds reasonable to me

wizeng23 (approved):

PR #1455, opened 2025-02-20
[Evaluation] Migrate LM Harness integration point from `simple_evaluate` to `evaluate`

kaisopos:
# Description

1) Migrating LM Harness integration point from `simple_evaluate` to `evaluate`, to support (follow-up PR) LM Harness custom tasks. 
2) Adding unit tests for LM Harness and AlpacaEval backends.
3) Adding integration tests for LM Harness.

## Related issues

Towards OPE-1029

## Before submitting

## Reviewers


src/oumi/evaluation/lm_harness.py
```diff
@@ -12,16 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
+import copy  # noqa: I001
```
kaisopos, line 15: Should I remove the `noqa` and replace the snippet below (I prefer it as is, but no strong opinion)?

Current
```
from lm_eval.loggers.utils import (
    add_env_info as lm_harness_add_env_info,
    add_tokenizer_info as lm_harness_add_tokenizer_info,
    get_git_commit_hash as lm_harness_get_git_commit_hash,
)
```

After
```
from lm_eval.loggers.utils import (
    add_env_info as lm_harness_add_env_info,
)
from lm_eval.loggers.utils import (
    add_tokenizer_info as lm_harness_add_tokenizer_info,
)
from lm_eval.loggers.utils import (
    get_git_commit_hash as lm_harness_get_git_commit_hash,
)
```

oelachqar: I think the error is just complaining that the imports are not sorted. If possible we should stick with the style guide

Another option is 
```python
import  from lm_eval.loggers.utils as lme_logger_utils

lme_logger_utils.get_git_commit_hash(...)
```

oelachqar (commented):
@kaisopos if we don't have them already this is a good opportunity to add unit tests !

src/oumi/evaluation/lm_harness.py
```diff
@@ -159,6 +185,78 @@ def _generate_lm_harness_model_args(
...
+    task_dict: dict[Union[str, ConfigurableGroup], Union[Task, dict]],
+    fn: Callable,
+    fn_kwargs: Optional[dict[str, Any]] = None,
+) -> None:
+    """Apply the provided function `fn` to all tasks in the `task_dict`."""
+    for task_obj in task_dict.values():
+        if isinstance(task_obj, dict):
+            _apply_to_all_tasks(task_obj, fn, fn_kwargs)
+        elif isinstance(task_obj, Task):
+            fn_kwargs = fn_kwargs or {}
```
wizeng23, line 198: Move this line outside the for loop to the beginning of the fn?

src/oumi/evaluation/lm_harness.py
```diff
@@ -159,6 +185,78 @@ def _generate_lm_harness_model_args(
...
+            raise ValueError(
+                f"Inconsistent task naming. Task `{task_params.task_name}` was "
+                f"requested, but LM Harness returned the `task_dict`: {task_dict}."
+            )
+
+    # Apply the following function to each task in the task_dict, in order to overwrite
+    # the default parameters with the ones specified in `task_params`.
+    def overwrite_task_params(task: Task) -> None:
+        # Set the number of few shots to be added to the prompt.
+        getattr(task, "set_config")(key="num_fewshot", value=task_params.num_fewshot)
```
wizeng23, line 234: nit: any particular reason to use getattr instead of accessing the attribute directly? The latter may be preferable for type safety.

src/oumi/evaluation/lm_harness.py
```diff
@@ -159,6 +185,78 @@ def _generate_lm_harness_model_args(
...
+
+        # Set the random seed (for reproducibility and consistency with LM Harness).
+        getattr(task, "set_fewshot_seed")(seed=1234)
```
wizeng23, line 237: Could you make 1234 a global constant? It's possible we may want to add this as a config param later, and it'll be more visible as a constant.

src/oumi/evaluation/lm_harness.py
```diff
@@ -159,6 +185,78 @@ def _generate_lm_harness_model_args(
...
+
+    _apply_to_all_tasks(task_dict, fn=overwrite_task_params)
+
+    return task_dict
+
+
+def _set_random_seeds(
+    random_seed: Optional[int] = 0,
```
wizeng23, line 245: ditto. A link to LM Harness code to prove these are the same values they use would be great as well

src/oumi/evaluation/lm_harness.py
```diff
@@ -227,38 +327,36 @@ def evaluate(
 
     if model_params.adapter_model:
         logger.info(f"Loading adapter for eval: {model_params.adapter_model}")
-    assert task_params is not None
```
wizeng23, line 230: Do we no longer need this assertion?

src/oumi/evaluation/lm_harness.py
```diff
@@ -227,38 +327,36 @@ def evaluate(
...
-        tasks=[task_params.task_name],
-        num_fewshot=task_params.num_fewshot,
-        batch_size=batch_size,  # type: ignore
-        device=device,
+    lm_eval_output = lm_harness_evaluate(
+        lm=lm,
+        task_dict=task_dict,
         limit=task_params.num_samples,
-        log_samples=False,
+        log_samples=False,  # Set to `True` to log all responses or logits.
```
wizeng23, line 359: Do you think users may want to toggle this? If so, consider making it an optional argument in the function, or making it a config value

wizeng23 (commented):
Could you add in your PR description the steps you took to test this PR? Would be good to verify that the results didn't change as a result of this migration, and that the output platform_task_config looks correct.

src/oumi/evaluation/lm_harness.py
```diff
@@ -187,6 +285,8 @@ def evaluate(
     Returns:
         The evaluation results (dict of metric names and their corresponding values).
     """
+    _set_random_seeds()
```
xrdaukar, line 288: optional: Should we provide a way to set non-default seed e.g., can be useful to see its  effect on eval score?

src/oumi/evaluation/lm_harness.py
```diff
@@ -159,6 +185,78 @@ def _generate_lm_harness_model_args(
...
+    # Sanity checks for `task_dict`.
+    if not task_dict:
+        raise ValueError(f"Task `{task_params.task_name}` not available in LM Harness.")
+    elif len(task_dict) > 1:
+        raise ValueError(
+            "Unexpected `task_dict` from LM Harness, consisting of multiple tasks, "
+            f"while a single task ({task_params.task_name}) was requested: {task_dict}."
+        )
+    else:
+        task_name = next(iter(task_dict))
```
xrdaukar, line 221: nit: `assert len(task_dict) == 1`  for clarity (It should be always true )

src/oumi/evaluation/lm_harness.py
kaisopos: Done

src/oumi/evaluation/lm_harness.py
kaisopos: Moved initialization to `evaluate()` (for visibility + users being able to change this [unlikely]) and also added a "Note for random seeds" that points to LM Harness.

src/oumi/evaluation/lm_harness.py
kaisopos: Not needed; it was there to make pyright happy.

src/oumi/evaluation/lm_harness.py
kaisopos: If users toggle this, the result log files will become significantly larger. I am explicitly calling out what this is because it's different than the LM Harness default param (=True) (if someone notices, I want them to quickly understand what this does and that it has NO impact on the results). This would only be needed for debugging, thus I am not sure if there's value exposing this to our inputs (if users need to debug: they can hardcode this)

src/oumi/evaluation/lm_harness.py
```diff
@@ -295,6 +393,24 @@ def evaluate(
...
+        platform_task_config["config"]["task_params"] = task_params
+        platform_task_config["config"]["task_dict"] = task_dict
+
+        # Add other configuration settings.
+        platform_task_config["git_hash"] = lm_harness_get_git_commit_hash()
+        lm_harness_add_env_info(platform_task_config)
+        lm_harness_add_tokenizer_info(platform_task_config, lm)
+
         if output_dir:
             save_evaluation_output(
```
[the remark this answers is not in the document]
kaisopos: Yep, we absolutely need proper integration tests, but not sure if I should implement these as part of the current PR or a follow-up. 

That said, there are also a few trade-offs here, since if we have proper test coverage, our test time will significantly increase. I will discuss this during our team meeting today.

src/oumi/evaluation/lm_harness.py
kaisopos: Added the seeds as inputs in our `evaluate()` function, so that users can set non-default values, though I think it's pretty unlikely users will ever try to change them. IMO the main reason for seeds' existence is to be able to **_reproduce the exact same results_** for consistency, rather than explore the deviation around the measured/average value when running multiple evals.

xrdaukar (approved):

taenin (commented):
+1 to adding unit tests before you submit here.

kaisopos:
> +1 to adding unit tests before you submit here.

Yep, will add unit tests before checkin, as discussed.
Unsure about integration tests though (this PR or a follow-up)

taenin:
> > +1 to adding unit tests before you submit here.
> 
> Yep, will add unit tests before checkin, as discussed. Unsure about integration tests though (this PR or a follow-up)

Up to you! Personally I like adding integration tests separately, but adding them here is fine as well.

src/oumi/evaluation/lm_harness.py
```diff
@@ -47,6 +43,8 @@
 from oumi.evaluation.save_utils import save_evaluation_output
 from oumi.utils.logging import logger
 
+FEW_SHOT_SEED = 1234
```
wizeng23, line 46: nit: Add a comment to describe where it's used. If you're trying to match a value that LM Harness is using, please link to that as well

src/oumi/evaluation/lm_harness.py
```diff
@@ -281,11 +279,22 @@ def evaluate(
         inference_engine_type: The inference engine to use (`VLLM`, `NATIVE`, `REMOTE`).
         inference_remote_params: The parameters for remote inference, if applicable.
         run_name: Unique identifier for wandb for the current training run.
+        random_seed: The random seed to use for reproducibility.
```
wizeng23, line 282: This is a bit vague. Could you clarify that it's for the python random package?

src/oumi/evaluation/lm_harness.py
kaisopos: Done

src/oumi/evaluation/lm_harness.py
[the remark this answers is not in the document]
kaisopos: Added unit tests

src/oumi/evaluation/lm_harness.py
kaisopos: Clarified.

tests/integration/evaluate/test_evaluate_alpaca_eval.py
```diff
@@ -0,0 +1,172 @@
...
+    assert annotators_config == "weighted_alpaca_eval_gpt4_turbo"
+    assert fn_metric == "get_length_controlled_winrate"
+    assert max_instances == INPUT_CONFIG["num_samples"]
+
+    # Ensure the inference results (`model_outputs`) are the expected ones.
+    for i in range(INPUT_CONFIG["num_samples"]):
+        if model_outputs["output"][i] != EXPECTED_INFERENCES[i]:
+            raise ValueError(f"Unexpected model output: {model_outputs['output'][i]}")
+
+    # Mock the `alpaca_eval.evaluate` function (by returning the expected results).
```
taenin, line 144: Is there a reason you're mocking in an integration test? This looks more like something you'd use in a unit test to me.

tests/unit/evaluation/test_lm_harness.py
```diff
@@ -0,0 +1,332 @@
...
+        lm_harness_model,
+        is_multimodal,
+        device,
+        model_params,
+        generation_params,
+        inference_engine_type,
+        inference_remote_params,
+    )
+
+    assert model_args == expected_model_args
```
taenin, line 177: Consider asserting that your mocked methods are called as well (`mock_build_tokenizer.assert_has_calls(...)`)

tests/unit/evaluation/test_lm_harness.py
```diff
@@ -0,0 +1,332 @@
...
+    assert task.OUTPUT_TYPE == "multiple_choice"
+
+
+@pytest.mark.skip(reason="Temporarily disabled because it times out.")
+def test_get_task_dict_for_configurable_group():
+    task_params = LMHarnessTaskParams(
+        evaluation_platform="lm_harness", task_name="mmmu_val", num_fewshot=222
+    )
+
+    task_dict = _get_task_dict(task_params)
```
taenin, line 205: I assume this times out because you're not mocking out the underlying call to the LM harness method `lm_harness_get_task_dict`. Most methods that are imported from `lm_eval` should be mocked out via fixtures in your unit tests.

tests/unit/evaluation/test_lm_harness.py
```diff
@@ -0,0 +1,332 @@
...
+
+@patch("oumi.evaluation.lm_harness.save_evaluation_output")
+@patch("oumi.evaluation.lm_harness.is_world_process_zero")
+@patch("oumi.evaluation.lm_harness.lm_harness_evaluate")
+@patch("oumi.evaluation.lm_harness.lm_harness_get_model_class")
+@patch("oumi.evaluation.lm_harness._generate_lm_harness_model_args")
+@patch("oumi.evaluation.lm_harness._get_task_dict")
+@patch("oumi.evaluation.lm_harness.is_image_text_llm_using_model_name")
+@patch("oumi.evaluation.lm_harness._set_random_seeds")
+@patch("torch.cuda.is_available")
```
taenin, line 237: If multiple tests need these it's likely better to make them a shared fixture.

tests/unit/evaluation/test_lm_harness.py
```diff
@@ -0,0 +1,332 @@
...
+def test_evaluate(
```
taenin, line 238: Are there any failure cases / edge cases we need to consider? This test looks like it cover the happy path.

tests/integration/evaluate/test_evaluate_alpaca_eval.py
kaisopos: To avoid being charged (for using GPT) every time we run the test.

tests/unit/evaluation/test_lm_harness.py
kaisopos: Yes, `_get_task_dict()` is a thin wrapper over `lm_eval.tasks.get_task_dict()` (with some variable replacements). If I mock the latter, there is no point running this unit test (I can drop it). I do feel it's useful though - should I move it under integration tests? Is the timeout longer there?

tests/unit/evaluation/test_lm_harness.py
kaisopos: Added 2 test cases corresponding to the 2 exceptions that this fn can throw. 

Overall: There are not many non-happy paths to consider in this specific unit test, since the inputs have been validated before and most things that can go wrong are related to combining inputs that don't make sense together; but this is not known unless we call lm_harness (which is mocked).

tests/unit/evaluation/test_lm_harness.py
kaisopos: Moved to integration tests.

tests/unit/evaluation/test_lm_harness.py
kaisopos: Done - thanks!

tests/integration/evaluate/test_evaluate_alpaca_eval.py
kaisopos: I am leaving this one unresolved (in case you still want me to remove the mock). 
Let me know if you do.

kaisopos: Moving the test from integration tests to unit tests (per offline discussion with Matt).

taenin (approved):

Issue #1386, opened 2025-02-05
[Feature] Update `oumi infer` to support CLI argument for system prompt


xrdaukar:
### Feature request

Add a new command line argument for `oumi infer` command

### Motivation / references

Users may want to adjust system prompt for task-specific  instructions

### Your contribution

I can help with code review, and can answer questions

Towards OPE-1019

Spaarsh:
@xrdaukar hello! I would like to work on this. I do understand where and how I will add the code for accepting a system-prompt, but then where am I supposed to use this input? I suppose it shall be a part of the ```InferenceConfig```?

Thanks.

xrdaukar:
Thank you @Spaarsh !   I think, at minimum, we can add a new `system_prompt` parameter to `infer()` CLI function https://github.com/oumi-ai/oumi/blob/4ad8780ab2f795171e7f5e99b9c0048a384f62d0/src/oumi/cli/infer.py#L37. (similarly to `image` param, which provides visual context ), then modify this function  https://github.com/oumi-ai/oumi/blob/4ad8780ab2f795171e7f5e99b9c0048a384f62d0/src/oumi/infer.py#L101 to add `Role.SYSTEM` message to the conversation.  It's OK to only support it for `interactive` mode initially. 

We could also consider adding system prompt param to `InferenceConfig` or `GenerationParams` but it'd be more involved and require more changes, so my suggestion is to do the minimal change as described above as the first step. @taenin fyi

PR #1439, opened 2025-02-17
Added fetch command and modified infer command to resolve oumi://

Spaarsh:
# Description

A new CLI endpoint ```fetch``` has been added to allows users to directly download configs from the github repo. The same function is then reused to allow the resolution of ```oumi://``` to download a the config file from the github repo and provide the address of this newly downloaded config file for further use by the ```infer``` command.

A test has been added in the ```tests_cli_main.py``` to check for successful registration of the ```fetch``` entry-point.

A new test file has been added named ```tests_cli_fetch.py``` that has three tests that cover the cases of default directory, explicit ```--output-dir``` passing and passing output directory via ```OUMI_DIR``` environment variable.

## Related issues

Fixes #1374 (issue)

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [x] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [x] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

## Reviewers

At least one review from a member of `oumi-ai/oumi-staff` is required.

PR #1463, opened 2025-02-20
Minor cleanup of oumi fetch

taenin:
# Description

I've slightly refactored the `oumi fetch` logic to live in `cli_utils` to make testing a bit easier.
I've also removed the `output_dir` argument from the `oumi infer` cli as it was unclear that it was for downloading configs. 

I also updated `oumi fetch` to fetch configs if the user omits the `oumi://` prefix.

## Related issues

Related to #1386

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/cli/infer.py
+++ b/src/oumi/cli/infer.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-from pathlib import Path
 from typing import Annotated, Final, Optional
 
 import typer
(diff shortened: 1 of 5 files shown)


src/oumi/cli/infer.py
```diff
@@ -33,16 +32,6 @@ def infer(
             help="Path to the configuration file for inference.",
         ),
     ],
-    output_dir: Annotated[
```
wizeng23, line 36: Is there some other name for the flag that you think will be more clear? Also fine to delete if you feel that it's not likely to be set

wizeng23 (commented):
Do you mind addressing my comments in https://github.com/oumi-ai/oumi/pull/1439, if they make sense to you?

src/oumi/cli/infer.py
taenin: I think for all entrypoints excepts `oumi fetch` we should use the env_var to determine this value. No need for the param here.

wizeng23 (approved):

taenin:
> Do you mind addressing my comments in #1439, if they make sense to you?

I'll leave these for now in case the original author is working on them!

PR #1466, opened 2025-02-21
Enable pre-release install for uv in pyproject.toml

### LINKED ISSUE #1462: 0.1.5 depends on unreleased package
### What happened?

Tried to update to 0.1.5, but uv won't do it.

 × No solution found when resolving dependencies:
  ╰─▶ Because only omegaconf<2.4.0.dev3 is available and oumi==0.1.5 depends on
      omegaconf>=2.4.0.dev3, we can conclude that oumi==0.1.5 cannot be used.
      And because you require oumi[gpu]==0.1.5, we can conclude that your requirements are
      unsatisfiable.

      hint: `omegaconf` was requested with a pre-release marker (e.g., omegaconf>=2.4.0.dev3), but
      pre-releases weren't enabled (try: `--prerelease=allow`)

### Steps to reproduce the bug

```
$ uv pip install --verbose -U oumi[gpu,aws,gcp,lambda,runpod]==0.1.5
```

### System Info

Oumi version: 0.1.4
Python version: 3.10.12
Platform: Linux-6.8.0-52-generic-x86_64-with-glibc2.35

Installed dependencies:
PACKAGE              VERSION
accelerate           1.2.1
aiohttp              3.11.12
bitsandbytes         0.45.2
datasets             3.2.0
diffusers            <not installed>
einops               <not installed>
jsonlines            4.0.0
liger-kernel         0.5.3
llama-cpp-python     <not installed>
lm-eval              0.4.7
numpy                1.26.4
nvidia-ml-py         12.560.30
omegaconf            2.3.0
open_clip_torch      <not installed>
pandas               2.2.3
peft                 0.14.0
pexpect              4.8.0
pillow               10.3.0
pydantic             2.9.2
responses            0.25.6
skypilot             0.7.0
tensorboard          2.18.0
timm                 <not installed>
torch                2.5.1
torchdata            0.9.0
torchvision          0.20.1
tqdm                 4.67.1
transformers         4.45.2
trl                  0.11.4
typer                0.15.1
vllm                 <not installed>
wandb                0.19.6

Environment variables:
VARIABLE                            VALUE
ACCELERATE_DYNAMO_BACKEND           <not set>
ACCELERATE_DYNAMO_MODE              <not set>
ACCELERATE_DYNAMO_USE_DYNAMIC       <not set>
ACCELERATE_DYNAMO_USE_FULLGRAPH     <not set>
ACCELERATE_USE_FSDP                 <not set>
LOCAL_RANK                          <not set>
LOCAL_WORLD_SIZE                    <not set>
OUMI_EXTRA_DEPS_FILE                <not set>
RANK                                <not set>
WORLD_SIZE                          <not set>

PyTorch information:
CUDA available: True
CUDA version: 12.4
cuDNN version: 90.1.0
Number of GPUs: 2
GPU type: NVIDIA GeForce RTX 3090
GPU memory: 23.7GB

wizeng23:
# Description

Unlike pip, uv disallows prelease packages by default, which we need for `omegaconf==2.4.0dev3` for `oumi>=0.1.5`. This PR fixes that in our pyproject.toml.

## Related issues

Fixes #1462

## Before submitting

### DIFF
--- a/src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
+++ b/src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
@@ -31,7 +31,7 @@ envs:
 
 setup: |
   set -e
-  pip install uv && uv pip install --prerelease=allow oumi[gpu]
+  pip install uv && uv pip install oumi[gpu]
 
 run: |
   set -e  # Exit if any command failed.
--- a/src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
+++ b/src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
@@ -31,7 +31,7 @@ envs:
 
 setup: |
   set -e
-  pip install uv && uv pip install oumi[gpu]
+  pip install uv && uv pip install --prerelease=allow oumi[gpu]
 
 run: |
   set -e  # Exit if any command failed.
--- a/.github/workflows/doctests.yaml
+++ b/.github/workflows/doctests.yaml
@@ -32,7 +32,7 @@ jobs:
       run: |
         # Install in system python as we're in a sandbox env
         # Install in verbose mode to see what's going on
-        uv pip install --prerelease=allow -e '.[ci_cpu]' --system
+        uv pip install -e '.[ci_cpu]' --system
 
     - name: Cache pre-commit
       uses: actions/cache@v4
--- a/.github/workflows/doctests.yaml
+++ b/.github/workflows/doctests.yaml
@@ -32,7 +32,7 @@ jobs:
       run: |
         # Install in system python as we're in a sandbox env
         # Install in verbose mode to see what's going on
-        uv pip install -e '.[ci_cpu]' --system
+        uv pip install --prerelease=allow -e '.[ci_cpu]' --system
 
     - name: Cache pre-commit
       uses: actions/cache@v4
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -35,7 +35,7 @@ jobs:
       run: |
         # Install in system python as we're in a sandbox env
         # Install in verbose mode to see what's going on
-        uv pip install -e '.[ci_gpu]' hf_transfer --system
+        uv pip install --prerelease=allow -e '.[ci_gpu]' hf_transfer --system
 
     - name: Download Test Data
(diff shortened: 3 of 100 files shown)


src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
```diff
@@ -31,7 +31,7 @@ envs:
 
 setup: |
   set -e
-  pip install uv && uv pip install oumi[gpu]
+  pip install uv && uv pip install --prerelease=allow oumi[gpu]
```
nikg4, line 34: there are a lot of repetitive and verbose changes here.  Can this be improved in the future ?

For example, we could make Launcher recognize some special command/macro  `oumi_pip_install  <targets>`  and 
 replace `oumi_pip_install ... `  with `pip install uv && uv pip install --prerelease=allow ...`
so we can avoid such sweeping changes in hundred configs.

nikg4 (approved):

src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
taenin: That definitely sounds like a doable feature ask. If we scope it out a bit, we can add this as a public issue.

taenin (approved):
+1 to Nikolai's comment. Consistently sending out PRs updating ~100 files isn't sustainable for reviewing. We should look into a solution to prevent changes like this if possible.

wizeng23:
A PR updating 100 files is unavoidable, as I'll have to do it with Nikolai's suggested change as well. I'm using find-replace to make this change, and am manually review all lines changed before sending it out for review. I understand the concern with this from a reviewing perspective, but IMO creating more convenience wrappers has its own downsides; pip is pretty well known, but if we create oumi_pip, users will now have to search through our code/docs to figure out what it does.

If you really want to avoid any massive code changes, I don't see another way besides doing a pip package release in Omegaconf.

src/experimental/configs/recipes/phi3/dpo/fsdp_gcp_job.yaml
wizeng23: This is no longer needed as I took an alternative approach Oussama suggested

wizeng23:
Update: Took an alternative approach Oussama suggested to fix this in the pyproject.toml

oelachqar (approved):

Issue #1351, opened 2025-02-03
[Feature] Add GRPO RL Support for Improved Preference RL Training


satish860:
It would be beneficial to include support for GRPO RL, enabling better preference-based reinforcement learning training.

oelachqar:
Hi @satish860, thank you for the request! This would be super valuable indeed.

If an RL expert from the community reads this, and is able to help in any capacity - please let us know!

nikg4:
OPE-993

PR #1482, opened 2025-02-25
GRPO trainer: Minimal initial integration

nikg4:
# Description

-- Define TRL_GRPO trainer type, and add the minimal plumbing to support it.
-- Add new registry type: `REWARD_FUNCTION`
-- Define a new base class for GRPO datasets
-- Define `trl-lib/tldr`  GRPO dataset  (it's used in GRPO example:  https://huggingface.co/docs/trl/main/en/grpo_trainer )
-- Define sample training config to train Qwen 0.5B model on the `trl-lib/tldr` dataset.
-- Register sample rewards functions to prefer completions of specific  length e.g.,  20 tokens.

Support for vLLM and for other configs params will be added as follow-up.

## Related issues

Towards OPE-993
Towards #1351

## Before submitting

## Reviewers


src/oumi/datasets/grpo/rewards/completion_length_rewards.py
```diff
@@ -0,0 +1,52 @@
...
+from oumi.core.registry import RegistryType, register
+
+
+def compute_soft_target_token_length_reward(num_tokens: int, *, target_tokens: int):
+    """Returns maximum reward for inputs that are `target_tokens` long.
+
+    The reward reduces smoothly if the actual number of tokens deviates
+    from `target_tokens`.
+    """
+    x = float(num_tokens) / target_tokens
```
taenin, line 31: nit: float division is the default in python 3 now, so this cast isn't necessary

src/oumi/datasets/grpo/rewards/completion_length_rewards.py
```diff
@@ -0,0 +1,52 @@
...
+
+import math
+import re
+
+from oumi.core.registry import RegistryType, register
+
+
+def compute_soft_target_token_length_reward(num_tokens: int, *, target_tokens: int):
+    """Returns maximum reward for inputs that are `target_tokens` long.
+
```
taenin, line 27: Nit: update the docstring to specify the args / return value. It could also be valuable to have the exact equation in the description: `num_tokens / (target_tokens * e^(num_tokens / target_tokens))`

taenin (approved):

src/oumi/datasets/grpo/rewards/completion_length_rewards.py
nikg4: updated

src/oumi/datasets/grpo/rewards/completion_length_rewards.py
nikg4: updated

src/oumi/datasets/grpo/rewards/completion_length_rewards.py
```diff
@@ -22,14 +23,19 @@ def _whitespace_tokenize(s: str) -> list[str]:
     return re.split(r"\s+", s)
 
 
+_NORMALIZING_SCALE_FACTOR: Final[float] = 1 / 0.36787944117
```
taenin, line 26: Should this be math.exp(-1) instead? Or are you concerned about this value being different on different hardware?

oelachqar (approved):

src/oumi/datasets/grpo/rewards/completion_length_rewards.py
nikg4: yes, it can be computed as  `1.0 / math.exp(-1)`. I just didn't want to invoke math computations  when initializing global constants (vs using a precomputed  value).

taenin: Edited my comment above: using `math.e` is probably more concise and should be a constant

nikg4: ah I see. Updated

PR #1490, opened 2025-02-27
Define `GrpoParams` under configs

nikg4:
# Description

-- Include initial set of params including  vLLM-related params.
-- Tested vLLM completions on 1 GPU.  Distributed  multi-GPU setup still needs work.

## Related issues

Towards OPE-993

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/grpo_params.py
+++ b/src/oumi/core/configs/params/grpo_params.py
@@ -1,0 +1,67 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from oumi.core.configs.params.base_params import BaseParams
+
+
+@dataclass
+class GrpoParams(BaseParams):
+    model_init_kwargs: Optional[dict[str, Any]] = None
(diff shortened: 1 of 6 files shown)


src/oumi/core/configs/params/grpo_params.py
```diff
@@ -0,0 +1,83 @@
...
+
+from oumi.core.configs.params.base_params import BaseParams
+
+
+@dataclass
+class GrpoParams(BaseParams):
+    model_init_kwargs: dict[str, Any] = field(default_factory=dict)
+    """Keyword arguments for `AutoModelForCausalLM.from_pretrained(...)`"""
+
+    use_vllm: bool = False
```
jgreer013, line 26: Could we re-use the inference config here?

src/oumi/core/configs/params/grpo_params.py
```diff
@@ -0,0 +1,83 @@
...
+    """Device where vLLM generation will run.
+
+    For example, "cuda:1". If set to `None`, the system will
+    automatically select the next available GPU after the last one used for training.
+    This assumes that training has not already occupied all available GPUs.
+    If only one device is available, the device will be shared between both training
+    and vLLM.
+    """
+
+    vllm_gpu_memory_utilization: float = 0.9
```
jgreer013, line 43: Normally this is set to 0.9 in inference, but in your config you set it to 0.3, is that because it's sharing a GPU? Should we make 0.3 the default?

jgreer013 (approved):

src/oumi/core/configs/params/grpo_params.py
nikg4: Right!   0.9 is also GRPO default .  In sample config, I set it to 0.3 because 1 GPU is shared for training and vLLM, which works for small (0.5B) model , but probably not the right default value in general.

taenin (approved):

src/oumi/core/configs/params/grpo_params.py
nikg4: there is some overlap between this config and InferenceConfig/Generation params, but it's not a perfect match:
1. Most of params in Inference config aren't applicable here
2. Not all params defined for GRPO are present in existing configs
3. Nesting InferenceConfig under TrainingConfig doesn't seem clean.
I think it's better not to do it

PR #1496, opened 2025-03-02
Fix `log_samples` not propagating from `eval_kwargs`

jgreer013:
Fix issue where `log_samples` kwarg is not able to be passed to lm_harness

# Description

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/evaluation/backends/lm_harness.py
+++ b/src/oumi/core/evaluation/backends/lm_harness.py
@@ -41,6 +41,8 @@ from oumi.core.distributed import is_world_process_zero
 from oumi.core.evaluation.evaluation_result import EvaluationResult
 from oumi.utils.logging import logger
 
+_LOG_SAMPLES_KEY = "log_samples"
+
 # Used to set the few-shot seed for lm_eval.api.task.Task. The value is consistent with
 # LM Harness `simple_evaluate`'s default `fewshot_random_seed` = 1234.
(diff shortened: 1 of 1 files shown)


src/oumi/core/evaluation/backends/lm_harness.py
```diff
@@ -345,11 +347,14 @@ def evaluate(
     lm = lm_class(**lm_harness_model_params)
 
     logger.info("Starting evaluation...")
+    log_samples = False
+    if _LOG_SAMPLES_KEY in task_params.eval_kwargs:
+        log_samples = task_params.eval_kwargs.pop(_LOG_SAMPLES_KEY)
```
kaisopos, line 352: Avoid poping, because we store the full input config and all input params for reproducibility.
Maybe instead just do `log_samples = task_params.eval_kwargs[_LOG_SAMPLES_KEY]`

kaisopos (approved):

src/oumi/core/evaluation/backends/lm_harness.py
jgreer013: If we leave the `log_samples` key in the kwargs, we get a duplicate argument error due to both being provided.

Alternatively we could do 2 different evaluate calls but that gets messy. Should we write the configs and samples before this point?

PR #1498, opened 2025-03-03
Add recurring tests to keep our test badges updated.

taenin:
# Description

This PR adds a new `ci_tests` workflow that triggers both our CPU and GPU tests in parallel on the `main` branch.

Currently our test badges in our README reflect the latest tests run from pull requests. This means that problematic PRs will cause our tests to be listed as failing even when tests pass on `main`.

Follow up work:
- Update our badges to look for only tests run on the `main` branch.

## Related issues

N/A

## Before submitting

## Reviewers

### DIFF
--- a/.github/workflows/nightly_tests.yaml
+++ b/.github/workflows/nightly_tests.yaml
@@ -1,0 +1,15 @@
+name: Nightly Tests
+
+on:
+  schedule:
+    - cron: '23 6 * * *' # 6:23 AM UTC ~ 11:23 PM PST
+  workflow_dispatch:
+
+jobs:
+  cpu-tests:
+    secrets: inherit
+    uses: ./.github/workflows/pretest.yaml
+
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly_tests.yaml
```diff
@@ -0,0 +1,15 @@
+name: Nightly Tests
+
+on:
+  schedule:
+    - cron: '23 6 * * *' # 6:23 AM UTC ~ 11:23 PM PST
```
nikg4, line 5: once a day seems too infrequent. Should it be every N hours?

nikg4 (approved):

oelachqar (approved):

.github/workflows/nightly_tests.yaml
taenin: SG, updated to once every 8 hours

PR #1499, opened 2025-03-03
[Evaluation] Adding support for logging model samples for all backends

kaisopos:
# Description

Adding support for logging model samples for all backends.
Generalizing Jeremy's PR-1496 for all backends

## Related issues

Towards OPE-1097

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/evaluation_params.py
+++ b/src/oumi/core/configs/params/evaluation_params.py
@@ -102,6 +102,15 @@ class EvaluationTaskParams(BaseParams):
     If set, this must be a positive integer.
     """
 
+    log_samples: Optional[bool] = False
+    """Whether to log the samples used for evaluation.
+
+    If not set (False): the model samples used for evaluation will not be logged.
+    If set to True: the model samples generated during inference and used for
+    evaluation will be logged in `backend_config.json`. The backend may also log
+    other intermediate results related to inference.
+    """
+
     eval_kwargs: dict[str, Any] = field(default_factory=dict)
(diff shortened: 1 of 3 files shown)


src/oumi/core/configs/params/evaluation_params.py
```diff
@@ -102,6 +102,15 @@ def my_evaluation(task_params, config):
     If set, this must be a positive integer.
     """
 
+    log_samples: Optional[bool] = False
```
nikg4, line 105: Consider adding a check in __post_init__  that log_samples is not re-defined in `eval_kwargs` as well (it can lead to confusing  and potentially error-prone behavior).  Example: https://github.com/oumi-ai/oumi/blob/0748703fceacc78e01e04f78d657058a0c6b6fa9/src/oumi/core/configs/params/data_params.py#L169

nikg4 (approved):

src/oumi/core/configs/params/evaluation_params.py
kaisopos: The same can happen for all parameters (not just `log_samples `).
I have code in place checking for this [here](https://github.com/oumi-ai/oumi/blob/0748703fceacc78e01e04f78d657058a0c6b6fa9/src/oumi/core/evaluation/evaluator.py#L276). 
And I also introduced a [test to check for this](https://github.com/oumi-ai/oumi/blob/0748703fceacc78e01e04f78d657058a0c6b6fa9/tests/unit/core/evaluation/test_evaluator.py#L478).

PR #1507, opened 2025-03-04
Update TRL to 0.15 and fix Liger/dataset code

wizeng23:
# Description

This includes 0.15.2 released two weeks ago. We want to upgrade to 0.15 to receive multiple new fields in GRPOParams, such as `vllm_max_model_len`.

The challenge is that `SFTTrainer`, which is the primary trainer we use in Oumi, significantly updated its dataset preparation logic: https://github.com/huggingface/trl/blob/fc2b041b58f6fbe766dceaec819bc5a8f9d209da/trl/trainer/sft_trainer.py#L356. The core problem is that `SFTTrainer` only accepts HF datasets for the most part, and we pass in different types of classes (`PretrainingAsyncTextDataset`, Datapipes, etc.).

This PR does a fairly simple fix for this. For the `TRL_SFT` trainer type, we look at the first example in the dataset, and if it includes the `input_ids` field, it means it's already processed aka tokenized, and we skip the dataset processing logic in  `trl`. `trl`'s dataset logic includes text processing tokenization, packing, and truncation, which our dataset logic should already be doing.

Finally, this PR sets HF's Liger arguments where possible instead of patching Liger ourselves to avoid a bug that's present  with Liger in trl 0.15: https://github.com/huggingface/trl/pull/2940.

I tested a variety of configs to make sure various dataset paths didn't have issues:
- `tests/scripts/gcp_e2e_tests_job.yaml`
- `configs/examples/fineweb_ablation_pretraining/ddp/gcp_job.yaml`
- `configs/examples/fineweb_ablation_pretraining/fsdp/gcp_job.yaml`
- `configs/examples/grpo_tldr/gcp_job.yaml`
- `configs/recipes/llama3_1/sft/8b_full/gcp_job.yaml`
- `configs/recipes/llama3_1/sft/8b_lora/gcp_job.yaml`
- `configs/recipes/llama3_1/sft/8b_qlora/gcp_job.yaml`
- `configs/recipes/qwq/sft/lora_gcp_job.yaml`
- `configs/recipes/smollm/sft/135m/quickstart_gcp_job.yaml`
- `configs/recipes/vision/llama3_2_vision/sft/11b_full/gcp_job.yaml`
- `configs/recipes/vision/llama3_2_vision/sft/11b_lora/gcp_job.yaml`
- `configs/recipes/vision/qwen2_vl_2b/sft/gcp_job.yaml`

## Detailed notes

- Non-HF datasets can't be used anymore, as specified by the type hints in the function. Note they technically violate their own type hints since `ConstantLengthDataset` is not a HF dataset. The specific attributes accessed in the `dataset` param is `column_names` and `map()`, which should be present for HF datasets.
    - `PretrainingAsyncTextDataset` (used for packed SFT datasets) is a non-HF dataset and will error. We subclass `ConstantLengthDataset` to inherit its special treatment.
    - There is another bug in the `SFTTrainer` code here: https://github.com/huggingface/trl/blob/fc2b041b58f6fbe766dceaec819bc5a8f9d209da/trl/trainer/sft_trainer.py#L423. They assume `column_names` is iterable, but it can in fact be null: https://github.com/huggingface/datasets/blob/b37230c09f29aea04799ed4a833f91ccf779a186/src/datasets/iterable_dataset.py#L2935. This was discovered with our `test_train.py:test_train_pack_with_pretraining_dataset` test. For some reason, the call to `maybe_convert_to_chatml` sets the `column_names` back to None.
    - We generally convert our `BaseMapDataset` and `BaseIterableDataset` classes to HF datasets before passing them into `SFTTrainer`. The only exception is when we want to pack a non-packed dataset (i.e. not a `BasePretrainingDataset`), we wrap the dataset as a `PretrainingAsyncTextDataset` which is not a HF dataset.

## Related issues

Towards OPE-1108, OPE-1117

## Before submitting


oelachqar (approved):
Thank you for debugging this Will !

taenin (approved):
Could you log a bug on HF for the inconsistencies you described? Curious if they're aware of them.

pyproject.toml
```diff
@@ -67,7 +67,8 @@ dependencies = [
     # Llama Vision attention is broken as late as 4.48.2 if gradient checkpointing is
     # enabled. See OPE-875 and https://github.com/huggingface/transformers/issues/36040.
     "transformers>=4.48.0,<4.49",
-    "trl>=0.14.0,<0.15",
+    # >=0.14.0 is needed for GRPOTrainer.
+    "trl>=0.15.0,<0.16",
```
nikg4, line 71: Please test configs with `use_torchdata: True` . For example

oumi launch up --config configs/recipes/vision/qwen2_vl_2b/sft/gcp_job.yaml --cluster qwen2-vl-vision

Don't forget to add  this: `pip install -e ".[gpu]"` !


I get this error:

_prepare_dataset
(qwen2-vl-sft-oumi-train, pid=6232) [rank0]:     if "prompt" in dataset.column_names and "completion" in dataset.column_names:
(qwen2-vl-sft-oumi-train, pid=6232) [rank0]:   File "/home/gcpuser/miniconda3/lib/python3.10/site-packages/torch/utils/data/datapipes/datapipe.py", line 145, in __getattr__
(qwen2-vl-sft-oumi-train, pid=6232) [rank0]:     raise AttributeError(
(qwen2-vl-sft-oumi-train, pid=6232) [rank0]: AttributeError: 'MapToIterConverterIterDataPipe' object has no attribute 'column_names

nikg4 (changes requested):
The change seems to be breaking VLM configs, let's verify they are OK before submit.

pyproject.toml
wizeng23: Fixed

src/oumi/train.py
```diff
@@ -298,6 +295,31 @@ def train(config: TrainingConfig, **kwargs) -> None:
     if len(config.data.get_split(DatasetSplit.VALIDATION).datasets) != 0:
         eval_dataset = build_dataset_mixture(config, tokenizer, DatasetSplit.VALIDATION)
 
+    # trl's SFTTrainer has its own dataset processing code. We should skip it if
+    # the dataset is already processed, i.e. it's tokenized and has an `input_ids` col.
+    # This generally occurs if the dataset is:
+    # 1. In the Oumi registry and thus is processed by the `BasePretrainingDataset` or
+    # `BaseSftDataset` classes
+    # 2. Packing is requested, and thus is processed by the
+    # `PretrainingAsyncTextDataset` class
```
nikg4, line 304: nit: Consider adding a reference to Linear task  ( in case we need to find it later)

nikg4 (approved):

src/oumi/core/datasets/pretraining_async_text_dataset.py
```diff
@@ -135,6 +135,11 @@ def __init__(
         else:
             self.formatting_func = lambda x: x
 
+    @property
+    def column_names(self):
```
nikg4, line 139: nit: add return type annotation?

PR #1509, opened 2025-03-04
Update sample GRPO script to validate num_generations

nikg4:
# Description

-- Check if global batch size is divisible by num_generations . Otherwise, training may fail (depending on `trl` version)

## Related issues

Fixes OPE-993

## Before submitting

## Reviewers

### DIFF
--- a/configs/examples/grpo_tldr/gcp_job.yaml
+++ b/configs/examples/grpo_tldr/gcp_job.yaml
@@ -38,16 +38,14 @@ run: |
   set -e  # Exit if any command failed.
   source ./configs/examples/misc/sky_init.sh
 
-  export PER_DEVICE_BATCH_SIZE=2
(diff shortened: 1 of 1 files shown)


wizeng23 (approved):
Do we want to move this logic into our code? This logic likely won't be copied into every future GRPO config. IMO not necessary though, as the current HF error for this is descriptive.

nikg4:
> Do we want to move this logic into our code? This logic likely won't be copied into every future GRPO config. IMO not necessary though, as the current HF error for this is descriptive.

Oumi has this logic, but it's currently treated as a warning https://github.com/oumi-ai/oumi/blob/e31dd220360bc01a3a09456571cc7fbb98538f78/src/oumi/train.py#L173
 Let's consider upgrading it to error if it fails for new `trl` versions.

PR #1522, opened 2025-03-08
Evaluation: Inference optimizations

kaisopos:
# Description

Enable re-using the same inference engine across all tasks. 

**_Exception_**: In the case of LM Harness: We need to destroy the engine, since LM Harness uses its own inference engine; a 2nd engine will allocate unnecessary resources, potentially increasing the memory footprint. 

## Related issues

Towards OPE-1097

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/evaluation/backends/alpaca_eval.py
+++ b/src/oumi/core/evaluation/backends/alpaca_eval.py
@@ -24,14 +24,13 @@ except ImportError:
 
 import pandas as pd
 
-from oumi.builders.inference_engines import build_inference_engine
 from oumi.core.configs import (
     AlpacaEvalTaskParams,
     EvaluationConfig,
-    InferenceConfig,
 )
 from oumi.core.distributed import is_world_process_zero
 from oumi.core.evaluation.evaluation_result import EvaluationResult
+from oumi.core.inference import BaseInferenceEngine
 from oumi.datasets.evaluation import AlpacaEvalDataset, utils
 from oumi.utils.logging import logger
 
@@ -39,6 +38,7 @@ from oumi.utils.logging import logger
 def evaluate(
     task_params: AlpacaEvalTaskParams,
     config: EvaluationConfig,
+    inference_engine: BaseInferenceEngine,
 ) -> EvaluationResult:
     """Evaluates a model using the Alpaca Eval framework.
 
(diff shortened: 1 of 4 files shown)


src/oumi/core/evaluation/evaluator.py
```diff
@@ -117,6 +121,10 @@ def evaluate_task(
             lm_harness_task_params = Evaluator._get_backend_task_params(task_params)
             assert isinstance(lm_harness_task_params, LMHarnessTaskParams)
 
+            # Destroy the inference engine, if created by a previous task. LM Harness
+            # uses its own inference engine, which is created internally.
+            self._inference_engine = None
```
kaisopos, line 126: Will this really destroy the engine and deallocate its resources? Is there a better way?
@taenin

src/oumi/core/evaluation/backends/alpaca_eval.py
```diff
@@ -96,25 +97,13 @@ def evaluate(
...
-        model_params=config.model,
-        remote_params=config.inference_remote_params,
-    )
-    responses = inference_engine.infer(
-        input=alpaca_dataset, inference_config=inference_config
+        "\tAlpacaEval inference `model_params`:\n"
+        f"{pformat(inference_engine._model_params)}\n"
+        "\tAlpacaEval inference `generation_params`:\n"
+        f"{pformat(inference_engine._generation_params)}"
     )
```
nikg4, line 105: I think there may be an ambiguity in `InferenceEngine` design:   you can pass  `inference_config` both in each  `infer(...)` call **and**  in engine's constructor. We should probably standardize on only one way or another , and remove the alternative way of passing `inference_config` @taenin WDYT ?

src/oumi/core/evaluation/evaluator.py
kaisopos: OK updated

nikg4 (approved):

src/oumi/core/evaluation/backends/alpaca_eval.py
kaisopos: I think @taenin mentioned that the purpose of (optionally) passing `inference_config` in each `infer(...)` allows us to update the generation params, after the engine has been instantiated (constructor).

taenin: Oussama and I chatted about this a bit. I agree that it's odd having them in both places. The reason that we do, at least for now, is that the inference config contains the generation parameters which include specific values like temperature, top_k, top_p, etc. When I'm playing with a model, I'd often like to change these fields on each query.

As @kaisopos  mentioned, if you don't supply the inference config in `infer` it will fall back to the one used during instantiation.

PR #1526, opened 2025-03-10
[Evaluation] Renaming `evaluation_platform` → `evaluation_backend`

kaisopos:
# Description

Renaming `evaluation_platform` → `evaluation_backend` for compatibility with PR-1484. Now that we released a new PyPI package (0.1.8) we can start using the new field name in our yaml files (ie., `evaluation_backend`) that is compatible with the latest release. 

## Related issues

Towards OPE-1097

## Before submitting

## Reviewers

### DIFF
--- a/configs/projects/aya/evaluation/eval.yaml
+++ b/configs/projects/aya/evaluation/eval.yaml
@@ -22,59 +22,59 @@ model:
   attn_implementation: "sdpa"
 
 tasks:
-  - evaluation_platform: lm_harness
+  - evaluation_backend: lm_harness
     task_name: m_mmlu_ar
     eval_kwargs:
       num_fewshot: 5
-  - evaluation_platform: lm_harness
+  - evaluation_backend: lm_harness
     task_name: m_mmlu_zh
     eval_kwargs:
       num_fewshot: 5
-  - evaluation_platform: lm_harness
+  - evaluation_backend: lm_harness
     task_name: m_mmlu_en
     eval_kwargs:
       num_fewshot: 5
-  - evaluation_platform: lm_harness
+  - evaluation_backend: lm_harness
     task_name: m_mmlu_fr
     eval_kwargs:
       num_fewshot: 5
-  - evaluation_platform: lm_harness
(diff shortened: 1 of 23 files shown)


wizeng23 (approved):
1. Could you check if there's any remaining references to evaluation_platform throughout our codebase? For example, I still see it here: tests/unit/core/evaluation/test_save_utils.py
2. We should either delete all references to evaluation_platform in src/oumi/core/configs/params/evaluation_params.py since we've had the deprecation warning out for a version, or update that deprecation warning to mention which version we will remove this parameter (I'd recommend 0.2.0). Feel free to do in another PR, but we should proceed with one of these IMO

wizeng23:
Could you please also run a sanity check on GCP to verify this?

kaisopos:
> 1. Could you check if there's any remaining references to evaluation_platform throughout our codebase? For example, I still see it here: tests/unit/core/evaluation/test_save_utils.py

This is by design, until we **_stop supporting both fields_** (in `evaluation_params.py`):
```
        if "evaluation_platform" in task_params_dict:
            task_params_dict.pop("evaluation_platform")  # deprecated field.
``` 
I checked again - I dont think there are any other references (expect `evaluation_params.py`, `test_save_utils.py`)

> 2. We should either delete all references to evaluation_platform in src/oumi/core/configs/params/evaluation_params.py since we've had the deprecation warning out for a version, or update that deprecation warning to mention which version we will remove this parameter (I'd recommend 0.2.0). Feel free to do in another PR, but we should proceed with one of these IMO

It doesn't really hurt us to support both parameters for a while; it's just 10 lines of code that we can remove anytime. Not sure what the right version should be; I will write 0.2.0 (as you suggested) but if this comes too soon, maybe we can delay more.

kaisopos:
> Could you please also run a sanity check on GCP to verify this?

Tested locally various yaml files (e.g. `oumi evaluate -c configs/recipes/smollm/evaluation/135m/quickstart_alpaca_v2_eval.yaml`)
Why GCP?

PR #1527, opened 2025-03-10
Add s1k dataset implementation

khushalid:
# Description

Added s1K dataset for the "How Reasonable Is Your Reasoning Model?" project.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+class S1KDataset(BaseSftDataset):
+    """Dataset class for the simplescaling/s1K dataset."""
+    system_prompt = (
+        "Below is the type of question, "
+        "paired with the question that describes the problem. "
+        "Write a response that gives appropriate answer to the question."
+    )
+    
+    default_dataset = "simplescaling/s1K"
+    QUERY_TEMPLATE_NOANSWER = """{Question}""".strip()
```
kaisopos, line 33: Is this used anywhere? If not, should we drop it?

src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+import pandas as pd
+
+from oumi.core.datasets import BaseSftDataset
+from oumi.core.registry import register_dataset
+from oumi.core.types.conversation import Conversation, Message, Role
+
+@register_dataset("simplescaling/s1K")
+class S1KDataset(BaseSftDataset):
+    """Dataset class for the simplescaling/s1K dataset."""
+    system_prompt = (
```
jgreer013, line 26: System prompt is missing from the `messages`

src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+        thinking_trajectory: str = example.get("thinking_trajectories", None) or ""
+        cot_type: str = example.get("cot_type", None) or ""
+        question: str = example.get("question", None) or ""
+        answer: str = example.get("attempt", None) or ""
+        user_prompt = f"### Question Type:\n{cot_type}\n\n### Question:\n{question}"
+        answer = "Answer: " + answer if "Answer:" not in answer else answer
+        
+        messages = [
+            Message(role=Role.USER, content=user_prompt),
+            Message(role=Role.ASSISTANT, content="<|im_start|>think\n" + "\n".join(thinking_trajectory).strip() + "\n<|im_start|>answer\n" + answer.strip())
```
jgreer013, line 47: I would avoid using reserved tokens like these when formatting your assistant turn.

Instead try:
`<think>\n...</think>\n...`

kaisopos, line 47: We should avoid using tags here; can you use text instead, like 
`<think>` and `</think>`?

src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+    QUERY_TEMPLATE_NOANSWER = """{Question}""".strip()
+
+    def transform_conversation(self, example: Union[dict, pd.Series]) -> Conversation:
+        """Transform a dataset example into a Conversation object."""
+
+        thinking_trajectory: str = example.get("thinking_trajectories", None) or ""
+        cot_type: str = example.get("cot_type", None) or ""
+        question: str = example.get("question", None) or ""
+        answer: str = example.get("attempt", None) or ""
+        user_prompt = f"### Question Type:\n{cot_type}\n\n### Question:\n{question}"
```
jgreer013, line 42: You could probably skip adding this particular format and just use `question` as the user prompt, unless you want this format to indicate that the model should think when prompts are formatted this way only.

src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+        answer = "Answer: " + answer if "Answer:" not in answer else answer
```
kaisopos, line 43: I noticed you used `attempt ` as the answer, while there's also an (unused) `solution` field. 
Should we use `solution` as well for something? thoughts?

jgreer013, line 43: Same as above, could skip adding "Answer:" to the answer.

src/oumi/datasets/sft/s1k.py
```diff
@@ -0,0 +1,50 @@
...
+    default_dataset = "simplescaling/s1K"
+    QUERY_TEMPLATE_NOANSWER = """{Question}""".strip()
+
+    def transform_conversation(self, example: Union[dict, pd.Series]) -> Conversation:
+        """Transform a dataset example into a Conversation object."""
+
+        thinking_trajectory: str = example.get("thinking_trajectories", None) or ""
+        cot_type: str = example.get("cot_type", None) or ""
+        question: str = example.get("question", None) or ""
+        answer: str = example.get("attempt", None) or ""
```
jgreer013, line 41: Just confirming that your goal is to train on the Gemini produced `attempt`, not the ground truth `solution` for this dataset? Otherwise I'd advise using `solution`, although in that case the reasoning trace likely won't make sense.

jgreer013 (changes requested):

src/oumi/datasets/sft/s1k.py
kaisopos: Per @penfever, how the data was processed during S1 training can be found [here](https://github.com/simplescaling/s1/blob/bbb6b50138cbd07dd818cce0060eedafebdbd8fd/train/sft.py#L52).

src/oumi/datasets/sft/s1k.py
khushalid: I will drop this one.

src/oumi/datasets/sft/s1k.py
khushalid: I followed the code [here](https://github.com/simplescaling/s1/blob/bbb6b50138cbd07dd818cce0060eedafebdbd8fd/data/tokenization.py#L31)
Since they were using <|im_start|> tags, I kept it like that. Should I change it to this:
`Message(role=Role.ASSISTANT, content="<think>\n" + "\n".join(thinking_trajectory).strip() + "\n</think>\n<answer>\n" + answer.strip() + "\n</answer>")`

src/oumi/datasets/sft/s1k.py
khushalid: attempt contains the full generated text of LLM.
solution only has the final answer.

I am trying to keep the format same as the text field in this [dataset](https://huggingface.co/datasets/simplescaling/s1K_tokenized), since they are using the [text](https://github.com/simplescaling/s1/blob/bbb6b50138cbd07dd818cce0060eedafebdbd8fd/train/sft.py#L67) field as the input text for training.

src/oumi/datasets/sft/s1k.py
khushalid: Will add system_prompt to the message.

src/oumi/datasets/sft/s1k.py
khushalid: I think this format is not important and I'll just use question as the user prompt.

src/oumi/datasets/sft/s1k.py
khushalid: Will remove the format.

src/oumi/datasets/sft/s1k.py
khushalid: Yes as per the s1 repo given, we are using attempt and not solution. I will go through the training sft part again of s1k and see if we need solution anywhere.

src/oumi/datasets/sft/s1k.py
nikg4: @khushalid  Please make and push the changes to unblock further review

wizeng23:
Hi @khushalid , please fix the errors in this PR; I can approve and merge it afterwards. You can run `pre-commit run --all-files` locally to reproduce the errors.

wizeng23:
Hi @khushalid, could you please address the above comments? Otherwise I'll close this PR.

wizeng23:
Closing due to inactivity. If you are still working on this, please address all previous comments and send out again for review.

PR #1536, opened 2025-03-12
Multimodal: Limit max number of images per Conversation

nikg4:
# Description

-- Using large number of images can lead to _much_ higher (model-dependent)  VRAM usage during training  and OOMs
-- Support `max_images` param in `VisionLanguageSftDataset`  to address this issue (a more elaborate solution may be needed, but it's a starting point)
-- Add a util to remove excessive images in `conversation_utils`

## Related issues

Towards OPE-355

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/datasets/vision_language_dataset.py
+++ b/src/oumi/core/datasets/vision_language_dataset.py
@@ -89,8 +89,9 @@ class VisionLanguageSftDataset(BaseSftDataset, ABC):
             trust_remote_code: Whether to trust remote code execution for the processor.
             return_conversations: Whether to return raw `Conversation` objects.
             max_images: The maximum number of images per conversation.
-                If the limit is exceeded, the first N images are used.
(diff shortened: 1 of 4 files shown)


wizeng23 (approved):
This could lead to issues in training if you have a message where the text is referencing a non-existent image, right? What about an alternative approach where the entire message is deleted if an image is removed from the message to avoid broken messages?

nikg4:
> This could lead to issues in training if you have a message where the text is referencing a non-existent image, right? What about an alternative approach where the entire message is deleted if an image is removed from the message to avoid broken messages?

yep, it can lead to potential quality issues so it should be used mindfully (e.g., if only affects a small fraction of data)

The alternative approach that you suggested (skip such conversations)  would be more difficult to implement ("ignore example" operation isn't supported in by Datasets API), so can be non-perfect too e.g., if you drop 80% of the data.  Anyhow, we should definitely consider adding  more options how to handle that.

PR #1539, opened 2025-03-13
Add letter counting GRPO example

wizeng23:
# Description

This uses [Jeremy's dataset](https://huggingface.co/datasets/oumi-ai/oumi-letter-count) to GRPO-train a model to actually be able to count how many R's are in strawberry.

- Register `oumi-ai/oumi-letter-count` dataset
- Register a simple letter count reward function. This function simply assumes the last set of digits in the completion is the answer. A future PR will improve this.
- Add GCP and train jobs for letter counting model
- Properly added `TldrGrpoDataset` to the registry, as it wasn't there before
- There is a bug with enabling `include_performance_metrics`. See #1540. Filed a TRL bug at https://github.com/huggingface/trl/issues/3081.

Tested that this runs on GCP, and that the loss is going down.

## Related issues

Towards OPE-1122
Towards #1540 

## Before submitting


configs/examples/grpo_tldr/gcp_job.yaml
```diff
@@ -51,11 +51,11 @@ run: |
   oumi distributed torchrun \
       -m oumi train \
       -c configs/examples/grpo_tldr/train.yaml \
-      --training.run_name="${OUMI_RUN_NAME}.${SKYPILOT_TASK_ID}" \
-      --training.per_device_train_batch_size=${PER_DEVICE_BATCH_SIZE} \
-      --training.grpo.num_generations=${GRPO_NUM_GENERATIONS} \
-      --training.max_steps=20 \
-      --training.save_steps=0 \
-      --training.save_final_model=true
+      --training.run_name "${OUMI_RUN_NAME}.${SKYPILOT_TASK_ID}" \
```
nikg4, line 54: nit:Why this change ?  IMO, the `foo=x` syntax is more direct, less ambiguous

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+
+def _find_last_number(s: str) -> int:
+    """Finds the last number (aka adjacent numeric digits) in a string."""
+    return int(re.findall(r"\d+", s)[-1])
+
+
+def compute_letter_count_reward(completion: str, target_count: int) -> int:
+    """Computes the rewards for counting the letters in a string."""
+    # This assumes that the last group of digits in the completion is the count.
+    # It also assumes that it is counting the right letter.
```
nikg4, line 28: Could you add some examples in a comment?  Currently, it's  not very clear what this function is trying to do.

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+
+import re
+
+from oumi.core.registry import RegistryType, register
+
+
+def _find_last_number(s: str) -> int:
+    """Finds the last number (aka adjacent numeric digits) in a string."""
+    return int(re.findall(r"\d+", s)[-1])
```
nikg4, line 22: Instead of raising an exception , can we detect "NOT_FOUND" condition and return None?

nikg4, line 22: if we care about the last occurrence, should it be something like this `\d+$` ? (not 100% sure about exact regex syntax for this)

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+    # It also assumes that it is counting the right letter.
+    try:
+        count = _find_last_number(completion)
+    except Exception:
+        count = 0
+    return -abs(count - target_count)
+
+
+@register("count_letters", RegistryType.REWARD_FUNCTION)
+def _count_letters(completions, letter_count, **kwargs):
```
nikg4, line 37: nit: add type hints?

nikg4, line 37: The dataset also contains "letter_count_string" column (values like "one", "two" ,.. instead of digits)

configs/examples/grpo_tldr/gcp_job.yaml
taenin: +1

configs/examples/grpo_letter_counting/gcp_job.yaml
```diff
@@ -0,0 +1,48 @@
...
+  pip install -U flash-attn --no-build-isolation
+
+run: |
+  set -e  # Exit if any command failed.
+  source ./configs/examples/misc/sky_init.sh
+
+  set -x
+  oumi distributed torchrun \
+      -m oumi train \
+      -c configs/examples/grpo_letter_counting/train.yaml \
```
taenin, line 45: We may consider starting to prefix these scripts with `oumi://` so they can use this example without having the github repo downloaded locally.
```suggestion
      -c oumi://configs/examples/grpo_letter_counting/train.yaml \
```

configs/examples/grpo_letter_counting/gcp_job.yaml
```diff
@@ -0,0 +1,48 @@
+# Job config to GRPO tune a letter counting model.
+#
+# Requirements:
+#   - Set up SkyPilot GCP: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html#setup
+#   - Log into WandB (`wandb login`) or disable `enable_wandb`
+#
+# Usage:
+#   oumi launch up -c configs/examples/grpo_letter_counting/gcp_job.yaml --cluster letter-counting-grpo
```
taenin, line 8: Ditto here for usage
```suggestion
#   oumi launch up -c oumi://configs/examples/grpo_letter_counting/gcp_job.yaml --cluster letter-counting-grpo
```

src/oumi/core/configs/params/grpo_params.py
```diff
@@ -128,7 +128,7 @@ def __post_init__(self):
             )
 
     def to_hf_trainer_kwargs(self) -> dict[str, Any]:
-        """Converts GRPO training params GRPOTrainer kwargs."""
+        """Converts GrpoParams to TRL's GRPOConfig kwargs."""
```
taenin, line 131: Shouldn't GRPO always be in caps?

src/oumi/datasets/grpo/letter_count.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+import pandas as pd
+from typing_extensions import override
+
+from oumi.core.datasets.base_grpo_dataset import BaseExperimentalGrpoDataset
+from oumi.core.registry import register_dataset
+
+
+@register_dataset("oumi-ai/oumi-letter-count")
```
taenin, line 22: Isn't it a bit redundant to use `oumi` twice here? Personally I would expect the dataset to be called something like `oumi-ai/letter-counting`

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+from oumi.core.registry import RegistryType, register
+
+
+def _find_last_number(s: str) -> int:
+    """Finds the last number (aka adjacent numeric digits) in a string."""
+    return int(re.findall(r"\d+", s)[-1])
+
+
+def compute_letter_count_reward(completion: str, target_count: int) -> int:
+    """Computes the rewards for counting the letters in a string."""
```
taenin, line 26: To Nikolai's comment here: let's expand this doc string to properly cover the input params and output, as well as what the function is doing.

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,41 @@
...
+    try:
+        count = _find_last_number(completion)
+    except Exception:
+        count = 0
+    return -abs(count - target_count)
+
+
+@register("count_letters", RegistryType.REWARD_FUNCTION)
+def _count_letters(completions, letter_count, **kwargs):
+    """Reward function for counting letters in a string."""
```
taenin, line 38: Needs a proper docstring

tests/unit/datasets/grpo/rewards/test_count_letters_rewards.py
```diff
@@ -0,0 +1,14 @@
...
+    assert compute_letter_count_reward("foo bar 1", target_count=1) == 0
+    assert compute_letter_count_reward("foo bar1", target_count=1) == 0
+    assert compute_letter_count_reward("foo bar one", target_count=1) == -1
+    assert compute_letter_count_reward("11 1", target_count=1) == 0
+    assert (
+        compute_letter_count_reward(
+            "The number of 'r's in strawberry is 10.", target_count=3
+        )
+        == -7
+    )
```
taenin, line 14: Even though each test is only one line, I'd recommend breaking them up into multiple tests. This way your test name can be descriptive, and it's clear what's going wrong if a test fails.

configs/examples/grpo_tldr/gcp_job.yaml
wizeng23: Consistency with our other configs, which basically all don't use "="

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
wizeng23: We don't necessarily want that, as the answer likely won't be the very last thing in the string. There's likely at least a period (ex. "The answer is 3.")

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
wizeng23: Done. Add them to `src/oumi/datasets/grpo/rewards/completion_length_rewards.py` as well

wizeng23: The reward function is meant to be a very rough heuristic for now, and I plan to improve it in a future PR (ex. modify the prompt to require a numeric output). I want to check in what I have first to clear my branch, and since there's a bug fix for the registry as well.

I may end up taking your suggestion in a future PR if I can't prompt the model to always respond with a numeric answer. However, it may end up being pretty complicated to catch all the ways the numbers can be said. Off the top of my head, there's also "once", "twice", and "thrice".

configs/examples/grpo_letter_counting/gcp_job.yaml
wizeng23: Is there an issue filed for this? IMO that change should be done in a separate PR which modifies all config files at once (your favorite type of PR :p ). Much easier to track where we've changed it or not

configs/examples/grpo_letter_counting/gcp_job.yaml
wizeng23: See above comment

src/oumi/core/configs/params/grpo_params.py
wizeng23: I don't think we have a consistent convention for this internally. We have "FSDPParams", yet we also have "PeftParams" and "LoraParams" instead of their capitalized variants. I don't have a strong opinion here, but I don't think it's in the scope of this PR; the class was already named GrpoParams before this PR.

src/oumi/datasets/grpo/letter_count.py
wizeng23: Agreed, but this is what Jeremy uploaded. I was running into issues reuploading it (and reformatting it to be more aligned with other GRPO datasets like TLDR), so opted not to do it in this PR. It's just a refactor without functional change, so pretty low-priority IMO.

configs/examples/grpo_letter_counting/gcp_job.yaml
taenin: No need to update all configs right now! I'd rather you use it here if we plan on making a blog post for this config though.

tests/unit/datasets/grpo/rewards/test_count_letters_rewards.py
wizeng23: Done, but IMO it's a bit overkill here :p test results show which line fails, and one-liner tests on a simple function are pretty self-descriptive. The only minor benefit I can see is that if multiple test cases break, they'll all be visible instead of being blocked by one failing case.

configs/examples/grpo_letter_counting/gcp_job.yaml
wizeng23: I can quickly submit a PR to update them all before the blog post (we should get better results for this before posting). I'd rather have configs changes be all-or-none, so people don't find our configs in an inconsistent state.

src/oumi/datasets/grpo/letter_count.py
wizeng23: https://linear.app/oumi/issue/OPE-1129

configs/examples/grpo_letter_counting/train.yaml
```diff
@@ -0,0 +1,56 @@
...
+  enable_gradient_checkpointing: False
+  gradient_checkpointing_kwargs:
+    use_reentrant: False
+  ddp_find_unused_parameters: False
+  optimizer: "adamw_torch_fused"
+  compile: True # Set to False if `grpo.use_vllm` is enabled: https://github.com/vllm-project/vllm/issues/12783
+
+  grpo:
+    num_generations: 6
+    use_vllm: False
```
nikg4, line 45: if vllm is disabled, should we remove all vllm params from teh config ? 
until all issues  are figured out

nikg4 (approved):

src/oumi/core/configs/params/grpo_params.py
taenin: Really a bummer since we end up with two different ways of capitalizing GRPO in the same comment. Alas.

taenin (approved):

tests/unit/datasets/grpo/rewards/test_count_letters_rewards.py
wizeng23: Ended up switching to parameterized tests. Totally forgot about them when I wrote this :p

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,81 @@
...
+        completion: The completion string from the LLM.
+        target_count: The target count of letters.
+
+    Returns:
+        The reward value, calculated as the negative of the absolute difference between
+        the count and the target count. The count is assumed to be the last group of
+        consecutive digits in the completion string.
+    """
+    count = _find_last_number(completion)
+    if count is None:
```
jgreer013, line 55: Maybe add a different negative reward for no number found?

jgreer013 (approved):

jgreer013:
> @wizeng23 Sanity check: Does it actually make the model count letters correctly?

We should definitely test with an experiment, but we're technically just rewarding the final answer, not the steps it takes to get there.

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
```diff
@@ -0,0 +1,81 @@
...
+        count = 0
+    return -abs(count - target_count)
```
jgreer013, line 57: Note: scaling the rewards based on how far away the answer is could be good, could be bad.

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
wizeng23: I plan to submit this for now and to do follow-up improvements in a different PR.

To address your comment, should it be a fixed negative reward for no number found, or should I have a stronger penalization than what I have currently (ex. set count to -1)?

src/oumi/datasets/grpo/rewards/count_letters_rewards.py
wizeng23: Yep, could experiment with a binary reward (1 for correct, 0 for incorrect, -1 for not found maybe). I'm guessing the target counts in the dataset almost all be smaller than ~5, but the model guessing a very high number could definitely mess things up.

PR #1545, opened 2025-03-16
Augmenting logging training/model statistics 

optas:
# Description

Before training begins, here we further log:
 - a model's total number of parameters, 
 - those trainable (and percent frozen), 
 - those corresponding to torch.nn.Embedding (embeddings).

Example output:
Screenshot 2025-03-16 at 1 31 22 AM

The `ModelParameterCount` was updated and relevant unit tests were added. 

## Related issues

Fixes # OPE-1130

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/torch_utils.py
+++ b/src/oumi/utils/torch_utils.py
@@ -200,15 +200,15 @@ class ModelParameterCount:
 
     @property
     def trainable_params_percent(self) -> float:
-        """Percentage of trainable parameters [0.0, 100.0], rounded to 2 decimals."""
+        """Percentage of trainable parameters [0.0, 100.0]."""
         if self.all_params == 0:
             return 0.0
-        return round(100 * self.trainable_params / self.all_params, 2)
+        return 100 * self.trainable_params / self.all_params
 
     @property
     def frozen_params_percent(self) -> float:
-        """Percentage of frozen parameters [0.0, 100.0], rounded to 2 decimals."""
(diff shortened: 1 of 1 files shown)


kaisopos:
2 high-level comments: 
- `model_summary.txt`: Instead of saving individual params in a `txt` file, I am wondering if we should save the entire training config (similarly to what we do for evaluations) as a `json` or `yaml` file
- timing: It's more common to save after a run (vs. before), together with the results. The reason is that if the runs keep on failing, we ll aggregate lots of configuration settings that don't work (which are noise). Saving after the run is useful for repro purposes.

src/oumi/utils/torch_utils.py
```diff
@@ -171,11 +172,38 @@ def get_device_name() -> str:
...
+            raise ValueError("`trainable_params` cannot be greater than `all_params`.")
+        if self.embedding_params > self.all_params:
+            raise ValueError("`embedding_params` cannot be greater than `all_params`.")
+
+    @property
+    def trainable_params_percent(self) -> float:
+        """Percentage of trainable parameters [0.0, 100.0], rounded to 2 decimals."""
+        if self.all_params == 0:
+            return 0.0
+        return round(100 * self.trainable_params / self.all_params, 2)
```
nikg4, line 200: I don't think we should  be `round()`-ing here . I suggest to return unrounded ratio.

We can round as needed in formatting function:  `{params.frozen_params_percent:.2f}%`

nikg4 (approved):

PR #1561, opened 2025-03-20
[Evaluation][Custom] Removing restrictions and better error checking

kaisopos:
# Description

1. Removing restrictions: Making the arguments `task_params` and `config` of the custom evaluation function optional. If users choose to define them, these will be populated but our Evaluator. If they don't need these, we should not make this function unnecessarily more complex by forcing them to define these.
2. Adding error checking for keyword arguments and providing better explanations when custom evals fail. Specifically: 
- Ensuring that user-provided keyword arguments, which are passed into method `Evaluator.evaluate` do NOT contain any reserved keys.
- Ensuring that user-provided keyword arguments match the expected input parameters of the custom evaluation function. If not, display to the user which specific arguments are unrecognized (provided by the user, but not known to the evaluation function) or missing (not provided by the user but required by the evaluation function).

## Related issues

Towards OPE-1097

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/evaluation/evaluator.py
+++ b/src/oumi/core/evaluation/evaluator.py
@@ -145,11 +145,12 @@ class Evaluator:
         elif evaluation_backend == EvaluationBackend.CUSTOM:
             evaluation_fn = Evaluator._get_custom_evaluation_fn(task_params.task_name)
             self._add_inference_engine_if_needed(evaluation_fn, kwargs, config)
+            custom_kwargs = Evaluator._merge_kwargs(kwargs, task_params.eval_kwargs)
 
             evaluation_result = evaluation_fn(
                 task_params=task_params,
                 config=config,
-                **kwargs,
+                **custom_kwargs,
             )
             if not isinstance(evaluation_result, EvaluationResult):
                 raise ValueError(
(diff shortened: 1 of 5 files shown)


src/oumi/core/evaluation/evaluator.py
```diff
@@ -143,15 +155,22 @@ def evaluate_task(
...
             )
         elif evaluation_backend == EvaluationBackend.CUSTOM:
-            evaluation_fn = Evaluator._get_custom_evaluation_fn(task_params.task_name)
-            self._add_inference_engine_if_needed(evaluation_fn, kwargs, config)
+            evaluation_fn_name = task_params.task_name or ""
+            evaluation_fn = Evaluator._get_custom_evaluation_fn(evaluation_fn_name)
             custom_kwargs = Evaluator._merge_kwargs(kwargs, task_params.eval_kwargs)
-
-            evaluation_result = evaluation_fn(
+            Evaluator._validate_custom_kwargs(
```
nikg4, line 161: nit: I think Python allows you to use `self.static_method()` syntax even for static methods

src/oumi/core/evaluation/evaluator.py
```diff
@@ -320,6 +339,88 @@ def _merge_kwargs(
...
+                "respectively."
+            )
+
+        # Ensure that user-provided keyword arguments, which are passed into method
+        # `Evaluator.evaluate`, match the expected input parameters of the custom
+        # evaluation function `evaluation_fn`.
+        fn_signature = inspect.signature(evaluation_fn)
+        fn_input_params = [param.name for param in fn_signature.parameters.values()]
+
+        provided_keys: set = custom_kwargs.keys() - set(RESERVED_KEYS)
```
nikg4, line 377: nit: `set[str]` annotation ?

nikg4 (approved):

src/oumi/core/evaluation/evaluator.py
kaisopos: Thanks didn't know that! Fixed across this file.

src/oumi/core/evaluation/evaluator.py
kaisopos: fixed

PR #1562, opened 2025-03-21
[Remote Inference] Supported params are ignored

kaisopos:
# Description

Updates:
1) ~~Validating that all params are supported, before the remote API call~~. 
2) Adding 2 missing params (`min_p`, `stop_token_ids`) in `_convert_conversation_to_api_input` (base remote engine)
3) Open AI: Dropping `logit_bias` for "o1-preview" (this model was not working before)
4) Instead of always using the model params that were used when initializing the inference engine (`self._model_params`), this change allows users to update the model with a new inference config (`inference_config.model`) after the engine is instantiated. This allows users to use the same OpenAI engine for gpt-4, but later (without re-instantiating the engine) they can pass in a new config and run inference for OpenAI's `o1-preview`.

## Related issues

Fixes OPE-1123

## Before submitting

## Reviewers


wizeng23:
Adding our oncall Nikolai. Nikolai, feel free to loop me in again if you need an additional review.

src/oumi/inference/remote_inference_engine.py
```diff
@@ -531,15 +542,17 @@ def infer_from_file(
...
-            "stop_strings",
             "temperature",
             "top_p",
+            "frequency_penalty",
+            "presence_penalty",
+            "logit_bias",
+            "stop_strings",
+            "stop_token_ids",
+            "min_p",
+            "guided_decoding",
```
kaisopos, line 555: Note: `min_p` and `stop_token_ids` were newly added here

src/oumi/inference/openai_inference_engine.py
```diff
@@ -33,3 +36,27 @@ def base_url(self) -> Optional[str]:
     def api_key_env_varname(self) -> Optional[str]:
         """Return the default environment variable name for the OpenAI API key."""
         return "OPENAI_API_KEY"
+
+    @override
+    def _convert_conversation_to_api_input(
```
kaisopos, line 41: This function could be omitted, since `logit_bias` is empty by default & the super function has been updated to NOT include `logit_bias` in the API call if empty. This fn will only protect the user if they unintentionally set this. It is questionable though if it's a better experience to silently ignore this or just fail.

src/oumi/inference/remote_inference_engine.py
```diff
@@ -243,7 +243,10 @@ def _get_list_of_message_json_dicts(
         )
 
     def _convert_conversation_to_api_input(
-        self, conversation: Conversation, generation_params: GenerationParams
+        self,
+        conversation: Conversation,
+        generation_params: GenerationParams,
+        model_params: ModelParams,
```
wizeng23, line 249: Could you explain in your PR description why model_params is being passed in now, instead of reading the one from class initialization (self._model_params)? It's a pretty significant change in this PR that's not documented

src/oumi/inference/remote_inference_engine.py
```diff
@@ -531,15 +548,17 @@ def infer_from_file(
...
-            "frequency_penalty",
-            "guided_decoding",
-            "logit_bias",
             "max_new_tokens",
-            "presence_penalty",
             "seed",
-            "stop_strings",
             "temperature",
             "top_p",
+            "frequency_penalty",
```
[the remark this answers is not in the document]
kaisopos: done

src/oumi/inference/remote_inference_engine.py
kaisopos: explained

PR #1570, opened 2025-03-25
[GRPO] Reorganize letter counting configs

wizeng23:
# Description

- Changed directory structure in preparation for future PR to add letter counting evaluation.
- Changed model for GRPO letter counting to Deepseek R1 distilled Qwen 1.5B, as reasoning models should have better performance.
- Removed some unnecessary `shard_for_eval` params for smaller models.
- Fixed broken documentation link.

## Related issues

Towards OPE-1122

## Before submitting


configs/examples/letter_counting/grpo/train.yaml
```diff
@@ -13,7 +13,7 @@
 #   - Other training configs: configs/**/pretraining/, configs/**/sft/, configs/**/dpo/
 
 model:
-  model_name: "Qwen/Qwen2-0.5B-Instruct"
+  model_name: "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
```
nikg4, line 16: Does this model train w/o errors ?  How much slower it is compared to 0.5B model ?

nikg4 (approved):

kaisopos (approved):

configs/examples/letter_counting/grpo/train.yaml
wizeng23: There was an unrelated error for model training, which I'm pretty sure didn't exist when i submitted my last PR. After fixing that, it trains. It's a bit slower, at 10 min for 5 steps instead of 6 min. The training speed seems rather variable though; for the 1.5B model, it goes from 5 steps after 10 min to 200 steps after 43 min. It's not just the first step that's slow (like with compilation), but the first couple.

taenin (approved):

PR #1574, opened 2025-03-27
[GRPO][Eval] Add letter counting eval

wizeng23:
# Description

- Add eval config and GCP job to count letters
- Add functions to enable GRPO datasets to output conversations
- Delete unneeded args for GRPO dataset init
- Add custom evaluation function, and registered it
  - Created `src/oumi/evaluation/registry` directory for holding registered eval fns
- Add system prompt to improve eval performance
- Add version restrictions to vllm installs. The latest version 0.8.2 requires torch 2.6.0 while we're on 2.5.

Tested on GCP.
- Accuracy for `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B`: 74.9%
  - Without system prompt: 32.7%
- Accuracy for `deepseek-ai/DeepSeek-R1-Distill-Llama-8B`: 93.4%
  - Without system prompt: 42.4%

Future work
- Test remote inference
- Support berry bench dataset
- Run both datasets against most popular models

## Related issues

Towards OPE-1122

## Before submitting

### DIFF
--- a/configs/examples/letter_counting/evaluation/eval.yaml
+++ b/configs/examples/letter_counting/evaluation/eval.yaml
@@ -20,13 +20,15 @@ model:
   trust_remote_code: True
 
 generation:
+  max_new_tokens: 2048
+  # This isn't used by vLLM, but is used for the native inference engine.
   batch_size: 4
 
 tasks:
   - evaluation_backend: custom
     task_name: count_letters
-    num_samples: 10
+    num_samples: 1000
 
-inference_engine: VLLM
+inference_engine: VLLM # Can also use NATIVE
 
 output_dir: "output/letter_counting/evaluation"
--- a/configs/examples/letter_counting/evaluation/eval.yaml
+++ b/configs/examples/letter_counting/evaluation/eval.yaml
@@ -13,17 +13,18 @@
 #   - Other eval configs: configs/**/evaluation/
 
 model:
-  model_name: "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
+  model_name: "HuggingFaceTB/SmolLM2-135M-Instruct"
+  # model_name: "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
   model_max_length: 131072
   torch_dtype_str: "bfloat16"
   attn_implementation: "sdpa"
   trust_remote_code: True
 
-data:
-  train:
-    datasets:
-      - dataset_name: "oumi-ai/oumi-letter-count"
-        split: "test"
+# data:
+#   train:
+#     datasets:
+#       - dataset_name: "oumi-ai/oumi-letter-count"
+#         split: "test"
 
 generation:
   batch_size: 4
--- a/configs/examples/letter_counting/evaluation/eval.yaml
+++ b/configs/examples/letter_counting/evaluation/eval.yaml
@@ -1,7 +1,8 @@
 # Config to eval an LLM's ability to count letters in words.
 #
 # Requirements:
-#   - Log into WandB (`wandb login`) or disable `enable_wandb`
+#   - Run `pip install vllm`
+#   - Log into HF: `huggingface-cli login`
 #
 # Usage:
 #   oumi evaluate -c oumi://configs/examples/letter_counting/evaluation/eval.yaml
@@ -21,7 +22,7 @@ model:
 
 generation:
   max_new_tokens: 2048
-  # This isn't used by vLLM, but is used for the native inference engine.
+  # This isn't used by vLLM, but is used for the NATIVE inference engine.
   batch_size: 4
 
 tasks:
@@ -29,6 +30,6 @@ tasks:
     task_name: count_letters
     num_samples: 1000
 
-inference_engine: VLLM # Can also use NATIVE
+inference_engine: VLLM # Can also use NATIVE if not running on GPUs
 
 output_dir: "output/letter_counting/evaluation"
--- a/configs/examples/letter_counting/evaluation/eval.yaml
+++ b/configs/examples/letter_counting/evaluation/eval.yaml
@@ -1,10 +1,10 @@
-# Eval config for Deepseek R1 Distill Qwen2.5 1.5B.
+# Config to eval an LLM's ability to count letters in words.
 #
 # Requirements:
 #   - Log into WandB (`wandb login`) or disable `enable_wandb`
 #
 # Usage:
-#   oumi evaluate -c configs/recipes/deepseek_r1/evaluation/distill_qwen_1_5b/eval.yaml
+#   oumi evaluate -c configs/examples/letter_counting/evaluation/eval.yaml
 #
 # See Also:
 #   - Documentation: https://oumi.ai/docs/en/latest/user_guides/evaluate/evaluate.html
@@ -19,14 +19,20 @@ model:
   attn_implementation: "sdpa"
   trust_remote_code: True
 
+data:
+  train:
+    datasets:
+      - dataset_name: "oumi-ai/oumi-letter-count"
+        split: "test"
+
 generation:
   batch_size: 4
 
 tasks:
-  # For all available tasks, see https://oumi.ai/docs/en/latest/user_guides/evaluate/evaluate.html
-  - evaluation_backend: lm_harness
-    task_name: mmlu_college_computer_science
-    eval_kwargs:
-      num_fewshot: 5
+  - evaluation_backend: custom
+    task_name: count_letters
+    num_samples: 10
 
-enable_wandb: True
+inference_engine: NATIVE
+
+output_dir: "output/letter_counting/evaluation"
--- a/configs/examples/letter_counting/evaluation/eval.yaml
+++ b/configs/examples/letter_counting/evaluation/eval.yaml
@@ -4,7 +4,7 @@
 #   - Log into WandB (`wandb login`) or disable `enable_wandb`
 #
 # Usage:
-#   oumi evaluate -c configs/examples/letter_counting/evaluation/eval.yaml
+#   oumi evaluate -c oumi://configs/examples/letter_counting/evaluation/eval.yaml
 #
 # See Also:
 #   - Documentation: https://oumi.ai/docs/en/latest/user_guides/evaluate/evaluate.html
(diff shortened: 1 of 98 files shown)


src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,66 @@
...
+        if not response or not isinstance(response.content, str):
+            continue
+        prediction = _extract_prediction(response.content)
+        if (
+            prediction is not None
+            and prediction == conversation.metadata["letter_count_integer"]
+        ):
+            count += 1
+
+    return EvaluationResult(
```
kaisopos, line 60: after the recent change, just return `{"count_letters": {"accuracy": count / total}}`

notebooks/Oumi - Build your own Custom Evaluation (Hallucination Classifier).ipynb
```diff
@@ -296,7 +296,10 @@
     "    # Compute any relevant metrics (such as Balanced Accuracy).\n",
     "    bacc = balanced_accuracy_score(y_true, y_pred)\n",
     "\n",
-    "    return EvaluationResult(task_result={\"bacc\": bacc})"
+    "    return EvaluationResult(\n",
+    "        task_name=\"my_custom_evaluation\",\n",
+    "        task_result={\"bacc\": bacc},\n",
+    "    )"
```
kaisopos, line 302: you can revert this; now it's automatically handled

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        return int(number_str)
+    except ValueError:
+        return None
+
+
+@register_evaluation_function("count_letters")
+def count_letters(
+    task_params: EvaluationTaskParams,
+    config: EvaluationConfig,
```
kaisopos, line 29: config is an optional parameter; if you are not using it remove it

src/oumi/evaluation/registry/count_letters_task.py
wizeng23: Should just be `{"accuracy": count / total}`, right?

src/oumi/evaluation/registry/count_letters_task.py
penfever, line 1: @wizeng23 @kaisopos , I think it would be nice to have a default behavior for this / future custom evals to (1) utilize standard metrics (most of the common ones are available in sklearn) and (2) return both balanced and unbalanced metrics by default (balanced accuracy, certain kinds of F1 scores are reasonable choices in most cases). For this particular eval, we could consider grouping the words according to difficulty (this is already done for BerryBench) and see if models fail more often on difficult words. Also for this particular eval, it makes sense to  also report distance-sensitive scores; cosine similarity and MAE seem like reasonable choices.

configs/examples/letter_counting/evaluation/eval.yaml
```diff
@@ -0,0 +1,34 @@
+# Config to eval an LLM's ability to count letters in words.
+#
+# Requirements:
+#   - Log into WandB (`wandb login`) or disable `enable_wandb`
```
kaisopos, line 4: It is disabled by default, right? So maybe you should say sth like:
If you want to use WandB, log in (`wandb login`) and enable it with `enable_wandb`

configs/examples/letter_counting/evaluation/eval.yaml
```diff
@@ -0,0 +1,34 @@
...
+  max_new_tokens: 2048
+  # This isn't used by vLLM, but is used for the native inference engine.
+  batch_size: 4
+
+tasks:
+  - evaluation_backend: custom
+    task_name: count_letters
+    num_samples: 1000
+
+inference_engine: VLLM # Can also use NATIVE
```
kaisopos, line 32: Can also use NATIVE --> Use `NATIVE` instead if you .... don't have GPUs?

configs/examples/letter_counting/evaluation/eval.yaml
```diff
@@ -0,0 +1,34 @@
...
+model:
+  model_name: "deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B"
+  model_max_length: 131072
+  torch_dtype_str: "bfloat16"
+  attn_implementation: "sdpa"
+  trust_remote_code: True
+
+generation:
+  max_new_tokens: 2048
+  # This isn't used by vLLM, but is used for the native inference engine.
```
kaisopos, line 24: nit: native -> `NATIVE` ? (so people can understand it's the `inference_engine ` below)?

configs/examples/letter_counting/evaluation/gcp_job.yaml
```diff
@@ -0,0 +1,59 @@
+# Job config to eval an LLM's ability to count letters in words.
+#
+# Requirements:
+#   - Set up SkyPilot GCP: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html#setup
+#   - Log into WandB (`wandb login`) or disable `enable_wandb`
```
kaisopos, line 5: Same comment here as above

src/oumi/core/datasets/base_grpo_dataset.py
```diff
@@ -95,4 +86,37 @@ def transform_grpo_example(self, example: Union[dict, pd.Series]) -> dict:
     @override
     def transform(self, sample: pd.Series) -> dict:
         """Validate and transform the sample into Python `dict`."""
-        return self.transform_grpo_example(sample)
+        return self._transform_grpo_example(sample)
+
+    def conversation(self, idx: int) -> Conversation:
```
kaisopos, line 91: conversation and conversations seem very generic.
How come they are not defined in the `BaseMapDataset`?

src/oumi/core/evaluation/evaluator.py
```diff
@@ -235,6 +235,8 @@ def _get_custom_evaluation_fn(task_name: Optional[str]) -> Callable:
                 "task name, which should be corresponding to a registered evaluation "
                 "function, using the decorator `@register_evaluation_function`."
             )
+        # Import to ensure custom evaluation functions are added to REGISTRY.
+        import oumi.evaluation.registry as evaluation_registry  # noqa: F401
```
kaisopos, line 239: How come this is not on the top of the file together with all the other imports?
Does this take too much time and we want a delayed initialization or something like that? 
Is there any other reason? (it's weird to have an import inside a private function)

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,71 @@
...
+
+from oumi.core.configs.params.evaluation_params import EvaluationTaskParams
+from oumi.core.inference.base_inference_engine import BaseInferenceEngine
+from oumi.core.registry import register_evaluation_function
+from oumi.datasets.grpo.letter_count import LetterCountGrpoDataset
+from oumi.utils.logging import logger
+
+
+def _extract_prediction(response: str) -> Optional[int]:
+    r"""Returns the numeric answer extracted from `\boxed{...}`, or returns None."""
```
kaisopos, line 26: do you need this `r` in the beginning of the string of the description?

kaisopos: super super nit: "Returns A or B" (instead of "Returns A, or returns B")

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,71 @@
...
+    regex_result = re.findall(r"\\boxed\{(\d+)\}", response)
+    if not regex_result or len(regex_result) != 1:
+        return None
+    number_str = regex_result[0]
+    # Except clause shouldn't trigger because the regex should only find ints.
+    try:
+        return int(number_str)
```
kaisopos, line 33: Does this work if `number_str` is a bool?
Does int(`False`) throw or is it 0?

kaisopos: I would double-check this works for corner cases.

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,71 @@
...
+    conversations = inference_engine.infer(input_conversations)
+    logger.info(f"Finished inference on {len(conversations)} conversations!")
+    if len(conversations) > 0:
+        logger.info(f"Sample conversation: {conversations[0]}")
+
+    count = 0
+    total = 0
+    for i, conversation in enumerate(conversations):
+        total += 1
+        response = conversation.last_message()
```
kaisopos, line 61: super nit: Add a comment to explain this, if you think it's needed

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -0,0 +1,71 @@
...
+        if not response or not isinstance(response.content, str):
+            continue
+        prediction = _extract_prediction(response.content)
+        if (
+            prediction is not None
+            and prediction == conversation.metadata["letter_count_integer"]
+        ):
+            count += 1
+
+    return {"accuracy": count / total}
```
kaisopos, line 71: Very debatable but just FYI: 
if your letter counter fails to produce a prediction (prediction is None), then you count this as a model failure, while it could be a prompt failure (your prompt may work better for some models vs others).
An alternative (not sure which solution is more fair, maybe check with Jeremy too): If prediction is None ignore it (`count` does not increase but also `total`--)

configs/examples/letter_counting/evaluation/eval.yaml
wizeng23: I blindly copied that; it shouldn't be there, good catch! I did log a bug for wandb support for custom evals: https://linear.app/oumi/issue/OPE-1173

src/oumi/core/datasets/base_grpo_dataset.py
wizeng23: My take is that BaseMapDataset is a much more generic base class, which is why those functions were defined in BaseSftDataset. I'm copying these functions over almost line-for-line as the GRPO dataset class needs similar behavior. I'll leave it up to Oussama if it makes sense to move it into BaseMapDataset or not.

src/oumi/evaluation/registry/count_letters_task.py
wizeng23: AFAIK line 32 should always work, but I put the except clause there as a precaution. This is because the regex is explicitly only grabbing a sequence of numeric digits from the string, which should be able to convert into an int.

src/oumi/evaluation/registry/count_letters_task.py
wizeng23: I think there's value in both. While part of the responsibility is on the user to prompt the model well, some models may not be good enough at instruction following to properly format the example, which is a different type of failure than the failure to count letters.

I now log both metrics in addition to some other useful numbers. Feedback on the names welcome!

PR #1577, opened 2025-03-31
Upgrade `transformers` to `v4.49.*`

nikg4:
# Description

-- Upgrade to `v4.49.*` , which is needed for Qwen 2.5 VL support
-- Updated more eval configs to use `shard_for_eval: True` for compatibility (see OPE-931), and for consistency with larger model sizes.
-- Upgrade to `v4.50` may be less straightforward (observed some errors/issues) to be handled later separately.
-- Tested: e2e integration tests` tests/scripts/launch_tests.sh` + some manual tests, including inference

## Related issues

Towards OPE-984, OPE-931, OPE-1175

## Before submitting

## Reviewers

### DIFF
--- a/configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml
+++ b/configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml
@@ -18,6 +18,7 @@ model:
   torch_dtype_str: "bfloat16"
   attn_implementation: "sdpa"
   trust_remote_code: True
+  shard_for_eval: True
 
 generation:
   batch_size: 4
--- a/configs/recipes/deepseek_r1/evaluation/distill_llama_8b/gcp_job.yaml
+++ b/configs/recipes/deepseek_r1/evaluation/distill_llama_8b/gcp_job.yaml
@@ -52,8 +52,7 @@ run: |
   echo "Starting evaluation for ${MODEL_CHECKPOINT_DIR} ..."
   set -x
 
-  oumi distributed accelerate launch \
-    -m oumi evaluate \
+  oumi evaluate \
     -c configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml \
     --run_name "${OUMI_RUN_NAME}.${SKYPILOT_TASK_ID}" \
     --model.model_name "${MODEL_CHECKPOINT_DIR}"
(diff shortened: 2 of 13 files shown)


oelachqar (approved):

configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml
```diff
@@ -18,6 +18,7 @@ model:
   torch_dtype_str: "bfloat16"
   attn_implementation: "sdpa"
   trust_remote_code: True
+  shard_for_eval: True
```
wizeng23, line 21: My fault for the removal. Forgot about the context for why they were there, and thought the models were too small to require sharding

wizeng23 (commented):
Which configs were tested with this change? Version upgrades sometimes result in very niche breakages (ex. Llama vision attention, counting tokens were broken in past updates). At the very least, I usually test Llama and Llama multimodal configs

configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml
wizeng23: That would be helpful, thanks! Hopefully inference isn't as affected since there's less complexities with it than training; you probably don't need to run every model type even. Also, only the NATIVE engine needs to be tested.

nikg4:
> Which configs were tested with this change? Version upgrades sometimes result in very niche breakages (ex. Llama vision attention, counting tokens were broken in past updates). At the very least, I usually test Llama and Llama multimodal configs

I run e2e eternal tests and they passed . Also test few random models manually.

In general, I think we should get away from exhaustive manual ad-hoc testing (extremely time consuming, and also error prone). If there are important uses cases, they should be added to https://github.com/oumi-ai/oumi/tree/main/tests/e2e

configs/recipes/deepseek_r1/evaluation/distill_llama_8b/eval.yaml
nikg4: IMO manual ad-hoc testing doesn't scale. Instead, we should add e2e tests for inference: https://linear.app/oumi/issue/OPE-1177
Any volunteers XD?

nikg4:
Submitting in the interest of time... LMK if it leads to any issues.

PR #1625, opened 2025-04-09
[GRPO] Add notebook to demonstrate GRPO & evaluation for letter counting

wizeng23:
# Description

This notebook GRPO-trains and evaluates Llama 3.2 3B on the letter counting task. It also can support evaluating various frontier models on the task as well. It supports being run from Colab, the user's local machine, and cloud clusters, but cloud clusters with enough memory are required to fully run the notebook.

Other misc. changes:
- Removed unnecessary installs from job configs (confirmed with tests)
- Removed `accelerate launch` command from letter counting eval as it's only run on one GPU.
- Update halloumi notebooks to install Oumi from PyPi
- Add system prompt to GRPO letter counting training
- Update answer parsing regex for training to match that used for eval. Also add a slight reward for getting the formatting right.
- Place system prompt before user prompt. This is because trl during GRPO training doesn't allow the last message to be a system prompt.

## Related issues

Towards OPE-1122

## Before submitting


notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Prerequisites\n",
+    "\n",
+    "### Machine Requirements\n",
+    "\n",
+    "This notebook runs both model evaluation and GRPO training, which require 8GB and 40GB VRAM, respectively.\n",
+    "\n",
+    "❗**NOTICE:** If running this notebook on Colab using a T4 GPU, it is not possible to run training due to memory requirements. To run evaluation, some adjustments need to be made as vLLM doesn't support T4 GPUs. This will be explained in the evaluation section.\n",
```
kaisopos, line 44: Nit: If you are running this notebook on Colab ...

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Evaluation\n",
+    "\n",
+    "First, we'll evaluate how various model perform on the letter counting task. We'll evaluate various frontier models by calling their respective remote API, in addition to evaluating Llama 3.2 3B by running local inference on it using vLLM.\n",
+    "\n",
+    "We've already defined a custom evaluation function in Oumi which runs inference on the above dataset, extracts the answer from the model response, and calculates various metrics such as accuracy. This function is defined at `src/oumi/evaluation/registry/count_letters_task.py`, and we print its contents below for reference."
```
kaisopos, line 154: Provide a link to count_letters_task.py.

IMO this is an important part (maybe the most important) to be excluded and not discussed. 
I suggest you at least show the signature and some pseudocode (I would personally prefer to NOT have this fn registered; it's not that useful for other use cases and breaks the flow of this notebook)

```
def count_letters(
    task_params: EvaluationTaskParams,
    inference_engine: BaseInferenceEngine,
) -> dict[str, Any]:
    # some psuedo code

    return {
        "accuracy": <model accuracy>,
        # other metrics
    }

Comment (kaisopos) [reply notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb:154]: "we print its contents below for reference" is not that helpful if you don't show the output

File: notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb [hunk 3]
```diff
@@ -0,0 +1,511 @@
...
+    "\n",
+    "print(inspect.getsource(count_letters))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the following section, you can select which models you want to evaluate. You can lower `NUM_SAMPLES`  to reduce cost when calling remote APIs, with the downside of noisier results."
+   ]
```
kaisopos, line 175: No action needed here, just FYI:
While I felt that the previous part missed an important function (`count_letters`) that the user should be able to see, I feel the exact opposite here. This is boilerplate code that the user should not see - e.g. this is really useless for a user: 

```
      inference_remote_params:
        api_url: "https://{REGION}-aiplatform.googleapis.com/v1beta1/projects/{PROJECT_ID}/locations/{REGION}/endpoints/openapi/chat/completions"
        max_retries: 3
        num_workers: 10
        politeness_policy: 60
        connection_timeout: 300
 ```
 
 I ll try to find a way to fix this, since we ll need the same boilerplate code in another notebook too

Comment (kaisopos) [reply notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb:175]: (suggestions very welcome on how to hide this)

File: notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb [hunk 4]
```diff
@@ -0,0 +1,511 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# EvaluationConfig for various models.\n",
+    "# Note that Llama 3B uses the local VLLM inference engines, while the others use various\n",
+    "# remote engines.\n",
+    "configs = {\n",
+    "    \"llama_3b\": \"\"\"\n",
```
kaisopos, line 212: Nit: The order is different here vs. other configs.
E.g. the `tasks:` is last in other configs

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "        temperature: 0.0\n",
+    "\n",
+    "      tasks:\n",
+    "        - evaluation_backend: custom\n",
+    "          task_name: count_letters\n",
+    "      \"\"\",\n",
+    "}"
+   ]
+  },
+  {
```
kaisopos, line 317: Add a markup box to explain this?

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "\n",
+    "    # Run the evaluation.\n",
+    "    evaluator = Evaluator()\n",
+    "    evaluator_out = evaluator.evaluate(config)\n",
+    "\n",
+    "    # # Record the results.\n",
+    "    results[model_name] = evaluator_out[0].get_results()"
+   ]
+  },
+  {
```
kaisopos, line 342: Add a markup box to explain this?

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "for model_name, result in results.items():\n",
+    "    print(\"-\" * 80)\n",
+    "    print(f\"Model: {model_name}\")\n",
+    "    print(f\"Accuracy: {result['accuracy']:.2%}\")\n",
+    "    correct = result[\"num_correct_answers\"]\n",
+    "    incorrect = result[\"num_incorrect_answers\"]\n",
+    "    invalid = result[\"num_invalid_answers\"]\n",
+    "    print(f\"Num correct, incorrect, invalid: {correct}, {incorrect}, {invalid}\")"
+   ]
+  },
```
kaisopos, line 358: Won't you discuss the results at all? 
Most people wont run this, they will just read it. 

IMO the whole point of notebooks is to show something interesting and catch users' attention. 
What is the point of this notebook if it doesn't inform the user on something - why are we doing this?

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## GRPO\n",
+    "\n",
+    "Now, we train Llama 3.2 3B on the task of counting letters using the GRPO algorithm implemented by [HuggingFace's `trl` library](https://huggingface.co/docs/trl/en/index).\n",
```
kaisopos, line 365: Why are we doing this? Just to do it?
The results above should give us a signal and then we should say: "wait, what if I would train my own model on this to address this and that issue I observed?"

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "\n",
+    "Note that we can calculate a concrete reward for this task by comparing the answer extracted by the model with the correct answer. In the reward function defined in `src/oumi/datasets/grpo/rewards/count_letters_rewards.py`, we calculate the reward to be `-abs(predicted_count - target_count)`. We use simple heuristics to extract the predicted count.\n",
+    "\n",
```
kaisopos, line 368: Is this something that we should surface and explain?

(overall, I have a different view on what is important to show in notebooks.
This is something potentially worth surfacing, while all the configs on how to query remote models are worth hiding.)

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "    incorrect = result[\"num_incorrect_answers\"]\n",
+    "    invalid = result[\"num_invalid_answers\"]\n",
+    "    print(f\"Num correct, incorrect, invalid: {correct}, {incorrect}, {invalid}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## GRPO\n",
```
kaisopos, line 363: This could be a different notebook and we could just reference it here. 
IMO there is less value to have so much code in a single notebook and not explain it thoroughly, rather than break it down into 2 independent parts and have more real estate to really discuss what is happening (show the letter counting function etc)

src/oumi/evaluation/registry/count_letters_task.py
```diff
@@ -50,9 +50,9 @@ def count_letters(
         num_samples = len(dataset)
     input_conversations = [dataset.conversation(i) for i in range(num_samples)]
     conversations = inference_engine.infer(input_conversations)
-    logger.info(f"Finished inference on {len(conversations)} conversations!")
+    logger.debug(f"Finished inference on {len(conversations)} conversations!")
```
taenin, line 53: Any reason to change from `info` -> `debug`?

taenin (commented):
I don't think https://github.com/oumi-ai/oumi/pull/1614 has been pushed to prod yet, so running `pip install oumi[gpu]` will not install VLLM until we push a new version. Is this something you're planning to do this week?

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "\n",
+    "### Machine Requirements\n",
+    "\n",
+    "This notebook runs both model evaluation and GRPO training, which require 8GB and 40GB VRAM, respectively.\n",
+    "\n",
+    "❗**NOTICE:** If running this notebook on Colab using a T4 GPU, it is not possible to run training due to memory requirements. To run evaluation, some adjustments need to be made as vLLM doesn't support T4 GPUs. This will be explained in the evaluation section.\n",
+    "\n",
+    "If your local machine cannot run this notebook, you can instead run this notebook on a cloud platform. The following demonstrates how to open a VSCode instance backed by a GCP node with 4 A100 GPUs, from which the notebook can be run. It is possible to run this notebook on just 1 GPU, but you will need make some adjustments to training parameters, which will be explained in the training section.\n",
+    "\n",
+    "```bash\n",
```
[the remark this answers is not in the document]
wizeng23: Yep! We don't want all users to blindly run this cell. It's simply a suggestion that should be run outside this notebook

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "os.environ[\"VLLM_WORKER_MULTIPROC_METHOD\"] = \"spawn\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Dataset\n",
+    "\n",
+    "The dataset we'll use for this notebook is [`oumi-ai/oumi-letter-count`](https://huggingface.co/datasets/oumi-ai/oumi-letter-count), which can be found on HF Datasets. Its prompts ask to count the letters in various English words, with metadata in each example containing the correct count. We use the `train` split for training and the `test` split for evaluation. The following code shows an example prompt:"
```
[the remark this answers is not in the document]
wizeng23: Fixed the first, and added an explanation for the second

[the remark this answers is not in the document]
wizeng23: Right, but running it will. IMO this is pretty standard for Colab

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
```diff
@@ -0,0 +1,511 @@
...
+    "pprint(dataset.conversation(0).to_dict())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Evaluation\n",
+    "\n",
+    "First, we'll evaluate how various model perform on the letter counting task. We'll evaluate various frontier models by calling their respective remote API, in addition to evaluating Llama 3.2 3B by running local inference on it using vLLM.\n",
```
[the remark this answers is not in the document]
wizeng23: IMO the previous diction is fine. The second sentence is explaining how exactly we'll evaluate different types of models (frontier vs Llama 3B). I did shorten it though.

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
oelachqar: How about just loading an existing config?

src/oumi/datasets/grpo/letter_count.py
```diff
@@ -77,5 +78,5 @@ def transform_conversation(self, sample: pd.Series) -> Conversation:
         # Convert messages from np.ndarray to list.
         sample_dict["messages"] = sample_dict["messages"].tolist()
```
oelachqar, line 79: nit: I'm pretty sure `.tolist()` does a copy, but just to be safe we could create an empty list and append the system preompt + the message

oelachqar (approved):

kaisopos (approved):

taenin (approved):

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: IMO it's still somewhat useful. Users who read this can see recommended configs for calling various frontier models. If they don't care, they could skim it/skip after reading one or two examples. When i read notebooks, I don't necessarily read every single line.

I tried having a base config which contains the fields used by most of those params, and then having the dict contain only the fields to override, and using OmegaConf.merge() to combine the two, but  it doesn't actually work well (the second config in the merge will override all fields in the first). There definitely should be a way to do this, but IMO it doesn't even save that many lines of code, and adds more code complexity, so i put it off for now

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: Users will see the contents of the fn by running this cell. I'll include some cell outputs before saving

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: Good eye! Fixed

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: Ditto

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: IMO it's overkill to have a markdown box, as it's short and the earlier markdown box does go over what this cell does. I did add a code comment though.

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: I'll keep some cell outputs before publishing

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: Even if there aren't specific numbers, I think it's still pretty intuitive that training a model on the task will improve the model's performance on that task.

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: Done. Added a GitHub link as well

notebooks/Oumi - Train a Letter Counting Model using GRPO.ipynb
wizeng23: I'd rather keep it as a single notebook to show an E2E real-world use case; eval model's performance on some task, train it to be better, and re-evaluate it. A notebook for just training would be quite short (basically just an `oumi train` invocation), and a notebook for evaluation would have high overlap with Halloumi/custom eval notebooks).

PR #1632, opened 2025-04-11
[Evaluation] Convenience fn to retrieve standard configs

kaisopos:
# Description

Convenience function to instantiate `EvaluationConfig`s for custom evals
- only for the most common/popular hosted models
- objective is to be used from users/notebooks that do NOT want to understand our configuration options; they want something that works as-is out-of-the-box without any complexity (ie. single step: model name -> config)

Signature example: how to run evaluations on LLaMA 405B without having to research the platforms that support it and how to query? This convenience function chooses GCP behind the scenes and only asks the user to provide the REGION and PROJECT_ID.

Why not define yaml files?
My top 2 reasons (if these can be solved well in yaml, lets do yaml instead)
1. Setting REGION and PROJECT_ID (for GCP) in the api_url defined in yaml is messy; should we put placeholders and replace with env variables and how?
2. Better error messages. As part of having the simplest user experience possible, we want to provide meaningful error messages that explain to the users what key or variable they need to set, how, and why (vs. getting a generic message). 

## Related issues

Fixes N/A

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/evaluation/configs.py
+++ b/src/oumi/evaluation/configs.py
@@ -1,3 +1,17 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import os
 from typing import Optional
 
--- a/src/oumi/evaluation/configs.py
+++ b/src/oumi/evaluation/configs.py
@@ -1,0 +1,170 @@
+import os
+from typing import Optional
+
+from oumi.core.configs import EvaluationConfig
+from oumi.core.configs.inference_engine_type import InferenceEngineType
+from oumi.core.configs.params.evaluation_params import EvaluationTaskParams
+from oumi.core.configs.params.generation_params import GenerationParams
+from oumi.core.configs.params.model_params import ModelParams
+from oumi.core.configs.params.remote_params import RemoteParams
+
+MAP_ENGINE_TO_API_KEY = {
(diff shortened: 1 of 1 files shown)


jgreer013 (approved):

src/oumi/evaluation/configs.py
```diff
@@ -0,0 +1,170 @@
...
+from oumi.core.configs.params.remote_params import RemoteParams
+
+MAP_ENGINE_TO_API_KEY = {
+    InferenceEngineType.OPENAI: "OPENAI_API_KEY",
+    InferenceEngineType.GOOGLE_GEMINI: "GEMINI_API_KEY",
+    InferenceEngineType.ANTHROPIC: "ANTHROPIC_API_KEY",
+}
+
+
+def _get_gcp_api_url(
```
oelachqar, line 18: You can set this in the inference engine class for GCP itself

src/oumi/evaluation/configs.py
```diff
@@ -0,0 +1,170 @@
...
+            f"Current values are: REGION=`{region}` and PROJECT_ID=`{project_id}`"
+        )
+
+    return (
+        f"https://{region}-aiplatform.googleapis.com/v1beta1/projects/"
+        f"{project_id}/locations/{region}/endpoints/openapi/chat/completions"
+    )
+
+
+def _validate_api_key(
```
oelachqar, line 42: The API key is optional for remote inference engines, most have one by default. We can skip setting it here

src/oumi/evaluation/configs.py
kaisopos: Our GCP inference engine requires an `api_url`; it does not work at the granularity of `REGION` and `PROJECT_ID` (users provide the entire `api_url` that embeds these), so it's hard to check for these specifically.

src/oumi/evaluation/configs.py
kaisopos: If this function is called, it means that the user is asking evaluation for a model that requires `api_key` (one of:  `OPENAI_API_KEY`, `GEMINI_API_KEY`, `ANTHROPIC_API_KEY`). Calling this function is conditional to using one of the OpenAI, Gemini, or Anthropic engines.

oelachqar (changes requested):
Let's chat offline on what the best place to put the util functions!

PR #1634, opened 2025-04-14
Support custom `processor args` in `ModelParams`

nikg4:
# Description

-- Update `ModelParams` to include `processor_args`
-- Update `build_processor()` and most of its invocations,  except `vision_language_conversation_feature_generator.py`, which will be handled as a follow-up (it's a more involved change)

## Related issues

Towards OPE-1185

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/model_params.py
+++ b/src/oumi/core/configs/params/model_params.py
@@ -69,6 +69,14 @@ class ModelParams(BaseParams):
     covered by other fields in ModelParams.
     """
 
+    processor_kwargs: dict[str, Any] = field(default_factory=dict)
+    """Additional keyword arguments to pass into the processor's constructor.
+
+    Processors are used in Oumi for vision-language models to process image and
+    text inputs. This field is optional and can be left empty for text-only models,
+    or if not needed.
+    """
+
     model_max_length: Optional[int] = None
     """The maximum sequence length the model can handle.
 
--- a/src/oumi/core/configs/params/model_params.py
+++ b/src/oumi/core/configs/params/model_params.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import json
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, fields
 from pathlib import Path
 from typing import Any, Optional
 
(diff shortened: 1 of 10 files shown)


src/oumi/core/configs/params/model_params.py
```diff
@@ -69,6 +69,14 @@ class ModelParams(BaseParams):
     covered by other fields in ModelParams.
     """
 
+    processor_kwargs: dict[str, Any] = field(default_factory=dict)
+    """Additional keyword arguments to pass into the processor's constructor.
+
+    Processors are used in Oumi for vision-language models to process image and
+    text inputs. This field is optional and can be left empty for text-only models,
+    or if not needed.
```
wizeng23, line 77: nit: consider mentioning that they'll override model-specific default values for these kwargs, if present.

taenin (approved):

src/oumi/core/configs/params/model_params.py
```diff
@@ -190,6 +198,17 @@ def __post_init__(self):
...
 
+        if len(self.processor_kwargs) > 0:
+            conflicting_keys = {f.name for f in fields(self)}.intersection(
+                self.processor_kwargs.keys()
+            )
+            if len(conflicting_keys) > 0:
+                raise ValueError(
+                    "processor_kwargs attempts to override the following "
+                    f"reserved fields: {conflicting_keys}. "
+                    "Use properties of ModelParams instead."
```
wizeng23, line 209: As far as I can tell, is `trust_remote_code` the only ModelParams property which is used in processor creation? Is there a chance of any false errors here, where a kwarg key conflicts with a ModelParams property, but that property isn't used in processor creation?

wizeng23 (approved):

src/oumi/core/configs/params/model_params.py
nikg4: updated

src/oumi/core/configs/params/model_params.py
nikg4: This validation logic is a bit inexact but should work in practice (we can always modify it if it something unexpected happens )

This is also based on two existing precedents: https://github.com/search?q=repo%3Aoumi-ai%2Foumi%20conflicting_keys&type=code

PR #1635, opened 2025-04-15
Support BerryBench evaluation

wizeng23:
# Description

This is fairly similar to letter counting eval. We use a different answer format, to align with what most LLMs tend to output for JSON. Tested with Llama 3b (1/10 correct) and Anthropic (8/10 correct).

## Related issues

Towards OPE-1122

## Before submitting

### DIFF
--- a/configs/examples/berry_bench/evaluation/eval.yaml
+++ b/configs/examples/berry_bench/evaluation/eval.yaml
@@ -1,0 +1,36 @@
+# Config to evaluate an LLM on the Oumi BerryBench dataset.
+#
+# Requirements:
+#   - Run `pip install oumi[gpu]`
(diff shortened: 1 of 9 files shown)


configs/examples/letter_counting/evaluation/gcp_job.yaml
```diff
@@ -38,6 +38,7 @@ envs:
 setup: |
   set -e
   pip install uv && uv pip install oumi[gpu]
+  pip install "vllm>=0.7.3,<0.8.0"
```
nikg4, line 41: nit: add a comment why this vllm override is needed

nikg4 (approved):

PR #1637, opened 2025-04-16
Adding LoRA train config for Qwen-VL 2.0

optas:
# Description

Adds a LoRA training  config for Qwen-VL 2.0 tested on multi **and** single GPU A:100s on GCP.

Note: per the potential bug reported on # 1190 ([linear](https://linear.app/oumi/issue/OPE-1190)), I added some trivial bash logic to enable the user to switch between single vs. multi-GPU via the launch configuration .yaml file ([link](https://github.com/oumi-ai/oumi/compare/main...optas/qwen_vl_2_lora#diff-35788a4ce94b934ca3f8a29aae8c74cdfc8ba69cfeba6f27032382d98d06da9bR49)). 

**The default is set on a single GPU, which jibes better with LoRA (and compute-constrained) users.**

## Related issues

Fixes OPE-1101

## Before submitting

## Reviewers


taenin:
> Reviewers, please let me know what you think about using the env variable `MULTI_GPU` to make the distinction on how to launch the training. This issue with a single GPU and distributed ([see Linear](https://github.com/oumi-ai/oumi/pull/1637)) **might** be affecting more models, which we might not have tested thoroughly with single GPU environments. If I am missing something gross here, please let me know, and obviously, I will update the code to work with distributed torchrun no matter how many GPUs one has access to.

I think it's likely better to fix the root cause. I don't think torchrun should fail for the single-gpu scenario, so something is clearly amiss.

configs/recipes/vision/qwen2_vl_2b/sft/lora/gcp_job.yaml
```diff
@@ -0,0 +1,63 @@
...
+  pip install uv && uv pip install oumi[gpu] hf_transfer
+  # Install model from HF Hub. This tool increases download speed compared to
+  # downloading the model during training.
+  HF_HUB_ENABLE_HF_TRANSFER=1 huggingface-cli download Qwen/Qwen2-VL-2B-Instruct
+
+run: |
+  set -e  # Exit if any command failed.
+  source ./configs/examples/misc/sky_init.sh
+
+  if [[ "${MULTI_GPU,,}" == "true" ]]; then
```
nikg4, line 49: Can we do something like below: 
```
if test ${OUMI_TOTAL_NUM_GPUS} -gt 1; then
  ...
else
 ...
fi
```
which relies on a pre-existing OUMI variable `OUMI_TOTAL_NUM_GPUS` (haven't seen MLTI_GPU used before so not entirely sure how it works and who/when sets it)

I agree with other comment that it's better to find a root cause, however we can submit this PR as a temporary fix

nikg4 (approved):

configs/recipes/vision/qwen2_vl_2b/sft/lora/gcp_job.yaml
taenin: +1 using `OUMI_TOTAL_NUM_GPUS` makes more sense. Be aware, though, that in a non-dev install `./configs/examples/misc/sky_init.sh` will not be available

taenin (approved):

configs/recipes/vision/qwen2_vl_2b/sft/lora/gcp_job.yaml
taenin: Yes, the job won't work if a user installs oumi via `pip install oumi`. The file `sky_init.sh` won't be found. This problem exists in other configs though, feel free to submit and we can address it separately.

PR #1638, opened 2025-04-16
[Remote Inference] Error checking for `api_key`

kaisopos:
# Description

Changes
- Do NOT include the AUTH BEARER header in the request, if API KEY is NOT available
- Fast fail if the API KEY is NOT available but required (`api_key_env_varname` is set)

## Related issues

Fixes N/A

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -437,8 +437,7 @@ class RemoteInferenceEngine(BaseInferenceEngine):
                     "An API key is required for remote inference with the "
                     f"`{self.__class__.__name__}` inference engine. "
                     "Please set the environment variable "
-                    f"`{remote_params.api_key_env_varname}` as follows: "
-                    f"`os.environ['{remote_params.api_key_env_varname}'] = <value>`"
+                    f"`{remote_params.api_key_env_varname}`."
                 )
         async with semaphore:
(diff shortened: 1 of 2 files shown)


src/oumi/inference/remote_inference_engine.py
```diff
@@ -428,6 +431,15 @@ async def _query_api(
...
         if not remote_params.api_url:
             raise ValueError("API URL is required for remote inference.")
+        if not self._get_api_key(remote_params):
+            if remote_params.api_key_env_varname:
+                raise ValueError(
+                    "An API key is required for remote inference with the "
+                    f"`{self.__class__.__name__}` inference engine. "
+                    "Please set the environment variable "
+                    f"`{remote_params.api_key_env_varname}` as follows: "
+                    f"`os.environ['{remote_params.api_key_env_varname}'] = <value>`"
```
taenin, line 441: This is only true if they're running in a python environment. This message will not be correct if they're using the CLI.

I'd consider ending the error message after 
```
f"Please set the environment variable `{remote_params.api_key_env_varname}"
```

They should be able to figure out fairly quickly how to set an env var for their environment.

taenin (approved):

src/oumi/inference/remote_inference_engine.py
kaisopos: Sg

PR #1644, opened 2025-04-17
[Evaluation] Convenience function for standard config retrieval

kaisopos:
# Description

Second attempt to implement this convenience function (previous one [here](https://github.com/oumi-ai/oumi/pull/1632))

Objective: Convenience function to instantiate `EvaluationConfig`s for custom evals
- only for the most common/popular hosted models
- objective is to be used from users/notebooks that do NOT want to understand our configuration options; they want something that works as-is out-of-the-box without any complexity (ie. single step: model name -> config)

After iterating on this with Matt: the fn will be implemented with the code snippet instead:
```
remote_config_path = try_get_config_name_for_alias(model_name, AliasType.EVAL)
local_config_path = cli_utils.resolve_and_fetch_config(remote_config_path)
config = EvaluationConfig.from_yaml(local_config_path)
config.tasks[0].task_name="my_custom_fn"
```

## Related issues

Fixes N/A

## Before submitting

## Reviewers


configs/projects/custom_evals/claude_3_5_sonnet.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+inference_engine: ANTHROPIC
+
+generation:
+  max_new_tokens: 8192
+  temperature: 0.0
+
+tasks:
+  - evaluation_backend: custom
+    task_name: my_custom_eval_function
```
taenin, line 12: Food for thought here: would it make more sense for all of these configs to be InferenceConfigs since this task won't exist for most users?

kaisopos: Should we have both?
If these are InferenceConfigs, I still need a solution for custom evals in notebooks.

Since note: I am tentatively planning to have "my_custom_eval_function" as the registered function name in all notebooks, to be able to use the configs as they are.

taenin: Previously I created an inference config for close-source models here: https://github.com/oumi-ai/oumi/blob/main/configs/apis/anthropic/infer_claude3_7.yaml

That could be a good place to keep just inference configs. You could always massage them into Eval configs as needed.

However, if they set values that only make sense for evaluation and not for inference, it's fine to keep them separate.

configs/apis/anthropic/infer_claude3_7.yaml
kaisopos, line 1: @taenin renamed this - is this used anywhere? can't find it.

taenin: Only here :P https://www.linkedin.com/posts/oumi-ai_want-to-try-out-the-new-claude-model-from-activity-7299963818447687681-EkGv?utm_source=share&utm_medium=member_desktop&rcm=ACoAAEOKcV4BdaNVBKzPM2mhtqk_jnaWJa08M0k

src/oumi/cli/alias.py
```diff
@@ -46,6 +46,35 @@ class AliasType(str, Enum):
     "llama4-maverick": {
         AliasType.INFER: "oumi://configs/recipes/llama4/inference/maverick_instruct_together_infer.yaml",
     },
+    # Hosted models.
+    "claude_3_5_sonnet": {
```
taenin, line 50: Use hyphens instead of underscores in model names. Here and below.

kaisopos: fixed

configs/apis/anthropic/infer_claude3_7.yaml
kaisopos: SG

taenin (approved):

PR #1372, opened 2025-02-04
Update transformers version to 4.48

wizeng23:
# Description

Llama vision attention is still broken. Nikolai determined it occurs when gradient checkpointing is enabled, and filed an issue here: https://github.com/huggingface/transformers/issues/36040.

Since this can be worked around, and only affects one family of models, we still want to upgrade the version. 4.45 is 4 months old, and we need newer transformers versions to support new vision models.

We attempted this previously, and reverted it in #1111 due to the same issue.

Main code changes required are:
- Rename `tokenizer` to `processing_class`
- Update some attention class names due to HF Transformers' attention refactor https://github.com/huggingface/transformers/pull/35235

## Related issues

Fixes OPE-738, OPE-875
Towards OPE-1018

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [x] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [x] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

PR #1646, opened 2025-04-20
Add verl PPO trainer

wizeng23:
# Description

Note to reviewers: I'll write a unit test after the implementation of the trainer is approved.

This PR wraps the RayPPOTrainer class from verl for use in Oumi. Note that the name of the class is misleading, and it supports many other RL algorithms as well, such as GRPO. A future PR will integrate this class into the training script.

Some notable parts of the implementation:
- Unlike other trainers, a validation dataset is required (though it's not documented by verl).
- verl requires datasets in Parquet format on disk as input to training. We create this by saving the HF train/val datasets with the `to_parquet()` method to an Oumi cache directory. This could be extended in the future to have a separate cache directory per dataset, or to not rewrite if there weren't dataset changes, etc.
- The verl config is an Omegaconf config, with default values specified in https://github.com/volcengine/verl/blob/main/verl/trainer/config/ppo_trainer.yaml . For now, we copy a snapshot of the latest file from verl to our repo. Later, we can try to find a smarter way to always get the updated version of this default config.
- The code to set up the trainer mirrors what's done in https://github.com/volcengine/verl/blob/main/verl/trainer/main_ppo.py .

I've confirmed that this class works when integrated with the training script (official code for that to follow in a separate PR).

Future work
- Populate verl config values from Oumi config
- Add new trainer type and invoke this class
- Add job/train configs for countdown task

## Related issues

Towards OPE-1192

## Before submitting


pyproject.toml
```diff
@@ -74,6 +74,7 @@ dependencies = [
     "trl>=0.16.0,<0.17",
     "typer",               # Used by CLI
     "typing_extensions",   # Backports of typing updates to python 3.9
+    "verl>=0.3.0,<0.4", # Used for the VERL_PPO trainer.
```
oelachqar, line 77: Let's make verl optional for now, until we're ready to fully merge / support

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+from oumi.core.configs import TrainingConfig, TrainingParams
+from oumi.core.tokenizers import BaseTokenizer
+from oumi.core.trainers.base_trainer import BaseTrainer
+from oumi.utils.logging import logger
+
+
+class VerlPpoTrainer(BaseTrainer):
+    """verl PPO Trainer.
+
+    Note that the name of the class is misleading, as it can support other algorithms
```
oelachqar, line 44: We could call it VerlGrpoTrainer. Later on to support PPO we could move shared functions to a BaseVerlTrainer class

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+            processing_class: The tokenizer for the model.
+            args: Training parameters.
+            reward_funcs: List of reward functions to use.
+            train_dataset: Training dataset.
+            eval_dataset: Validation dataset. This is required by verl.
+            **kwargs: Additional keyword arguments.
+        """
+        self.processing_class = processing_class
+        self.params = copy.deepcopy(args)
+        if len(reward_funcs) > 1:
```
oelachqar, line 72: Is this a limitation from `verl`, or a temporary workaround ?

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+        self._create_dataset_files()
+
+        self._setup_verl_trainer()
+
+    def _create_dataset_files(self) -> None:
+        """Creates dataset files for verl in Parquet format.
+
+        The Parquet files are saved to the Oumi cache directory.
+        """
+        self.cache_dir = Path.home() / ".cache" / "oumi" / "verl_datasets"
```
oelachqar, line 88: nit: we should define `self.cache_dir` in the constructor, or make it private

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Args:
+            processing_class: The tokenizer for the model.
+            args: Training parameters.
+            reward_funcs: List of reward functions to use.
+            train_dataset: Training dataset.
+            eval_dataset: Validation dataset. This is required by verl.
+            **kwargs: Additional keyword arguments.
+        """
+        self.processing_class = processing_class
```
oelachqar, line 70: Maybe we can log that this is experimental and that the api might change ?

oelachqar (approved):

pyproject.toml
wizeng23: Moved under GPU for now, LMK if you think somewhere else is better

oelachqar: SG, you'll also need to replace the direct imports to make sure the library works when verl is not installed (e.g. vllm)

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+from verl.trainer.ppo.ray_trainer import (
+    RayPPOTrainer,
+    ResourcePoolManager,
+    Role,
+)
+from verl.workers.fsdp_workers import (
+    ActorRolloutRefWorker,
+    CriticWorker,
+)
+from verl.workers.reward_manager import NaiveRewardManager
```
taenin, line 33: If Verl's not a mandatory dependency you should wrap it like we do with LlamaCPP: https://github.com/oumi-ai/oumi/blob/main/src/oumi/inference/llama_cpp_inference_engine.py#L27

src/oumi/core/trainers/verl_ppo_trainer.py
taenin: +1

taenin (approved):
Should we add an "experimental" directory here similar what we've done for models? https://github.com/oumi-ai/oumi/tree/560820fe7a0e58f68f29402eee53f57d9b3770bf/src/oumi/models/experimental

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+    Note that the name of the class is misleading, as it can support other algorithms
+    like GRPO. Currently, we only support GRPO.
+
+    For documentation on the underlying verl PPO trainer, see
+    https://verl.readthedocs.io/en/latest/examples/config.html.
+    """
+
+    def __init__(
+        self,
+        processing_class: Optional[BaseTokenizer],
```
nikg4, line 53: this is a bit inconsistent with other code that calls this param "tokenizer". Also, currently in Oumi there exist two separate concepts form Tokenizer and Processor (https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/processors/base_processor.py ) 

 my suggestion is to rename the param to "tokenizer"

src/oumi/core/trainers/verl_ppo_trainer.py
```diff
@@ -0,0 +1,192 @@
...
+            or self.verl_config.actor_rollout_ref.actor.use_kl_loss
+        ):
+            role_worker_mapping[Role.RefPolicy] = ray.remote(ActorRolloutRefWorker)
+            mapping[Role.RefPolicy] = global_pool_id
+        resource_pool_manager = ResourcePoolManager(
+            resource_pool_spec=resource_pool_spec, mapping=mapping
+        )
+
+        # Create reward function manager
+        compute_score = self.reward_funcs[0] if self.reward_funcs else None
```
nikg4, line 146: optional: if there are multiple rewards functions, would it be possible to create   a private reward function that calls them all under the hood and returns an aggregated reward value (e.g., a sum) ?

src/oumi/core/trainers/verl_ppo_trainer.py
nikg4, line 70: nit: consider adding underscore to all member variables  to make everything (pseudo-)private by default

For example:
``self._processing_class`

nikg4 (approved):

src/oumi/core/trainers/verl_ppo_trainer.py
wizeng23: I need to investigate further. At a quick glance, I didn't see how to support multiple reward functions. Left a TODO for now.

wizeng23:
```
Should we add an "experimental" directory here similar what we've done for models? https://github.com/oumi-ai/oumi/tree/560820fe7a0e58f68f29402eee53f57d9b3770bf/src/oumi/models/experimental
```

To my understanding, this will be a first-class citizen trainer class we want to support, as it's a popular RL platform. The class will be integrated in the next PR, and I just defined it separately to reduce reviewer load. IMO it's no less experimental than the trl GRPOTrainer.

src/oumi/core/trainers/verl_ppo_trainer.py
wizeng23: Done. Also renamed verl_ppo_trainer.yaml to verl_trainer_config.yalm to reduce confusion.

src/oumi/core/trainers/verl_ppo_trainer.py
wizeng23: This is actually a change to all our trainers we had to make for https://github.com/oumi-ai/oumi/pull/1372. In summary, we renamed it because HF renamed it. See https://github.com/oumi-ai/oumi/blob/7bdb6ee13197b8cb97999dfbd155b2151e0cf7af/src/oumi/train.py#L194.

https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTTrainer.processing_class

I do agree it sounds quite similar to processor, which is a different thing.

src/oumi/core/trainers/verl_ppo_trainer.py
wizeng23: Good idea! I'm keeping it as a future extension for now to get this out sooner, and because most uses cases I've been seeing so far use just one reward.

wizeng23:
In the interest of time, I added some simple unit tests for now. I'll add more thorough ones later, which require a lot of mocking.

PR #1652, opened 2025-04-24
Integrate verl GRPO trainer into train script

wizeng23:
# Description

To reviewers: Feel free to provide on whether you think this is the best way to structure the trainer.

- Remove ray from `pyproject.toml`, as it's installed as a dependency of verl.
- Created new `VERL_GRPO` trainer type
- Add a training config validation check that a validation dataset is specified for verl training
- Create a separate function that initialized Ray and invokes the verl trainer.
    - The trainer has to run inside a ray remote function. To prevent the train script from erroring if ray isn't installed, the `@ray.remote` decorated function is nested inside the verl trainer function.
- Move setup shared by all trainers (creating datasets, tokenizer, etc.) before this function all, while setup not required by/breaking verl is done after. 
- Move `processing_class` out of the `_create_optional_training_kwargs` since it's required by all trainers.

Tested that verl GRPO training works, and also tested that regular training isn't affected by the ray import.
Finally, I ran e2e integration tests to confirm our other trainers don't break.

## Related issues

Towards OPE-1192

## Before submitting


src/oumi/train.py
```diff
@@ -184,31 +189,72 @@ def _finalize_training_config(config: TrainingConfig) -> TrainingConfig:
...
 def _create_optional_training_kwargs(
     config: TrainingConfig,
-    tokenizer: Optional[BaseTokenizer],
     trainer_type: TrainerType,
     metrics_function: Optional[Callable],
     reward_functions: list[Callable],
     collator: Optional[Callable],
     additional_trainer_kwargs: Optional[dict[str, Any]] = None,
 ) -> dict[str, Any]:
-    kwargs: dict[str, Any] = {"processing_class": tokenizer}
```
nikg4, line 194: why remove `"processing_class": tokenizer` ?

src/oumi/train.py
```diff
@@ -276,6 +301,86 @@ def train(
...
+
+    # We support running FSDP Oumi training without being invoked from the Accelerate
+    # launcher. We detect this with the following:
+    # 1. Accelerate's environment variables aren't set
+    # 2. We are running with a HF-family trainer (HF, TRL_SFT, TRL_DPO, TRL_GRPO)
+    # 3. FSDP is enabled in the Oumi config
+    # In this case, we mimic an Accelerate launcher run by setting the necessary
+    # environment variables.
+    # Note that normal Accelerate launcher runs won't be affected.
+    if (
```
nikg4, line 374: Nothing wrong with the changes as far as I can tell, but still it's a pretty big restructuring in `train()` function. Probably need to do some significant e2e testing that it doesn't lead to any unexpected side effects for misc models.

nikg4 (approved):

src/oumi/core/configs/training_config.py
```diff
@@ -203,3 +203,12 @@ def __post_init__(self):
...
                         **self.model.processor_kwargs
                     }
+
+        # Verl will error without a validation dataset.
+        if (
+            self.training.trainer_type == TrainerType.VERL_GRPO
+            and not self.data.validation.datasets
+        ):
+            raise ValueError(
+                "At least one validation dataset is required for VERL GRPO training."
```
taenin, line 213:
```suggestion
                "At least one validation dataset is required for VERL_GRPO training."
```
Minor nit for consistency with other messages

src/oumi/train.py
```diff
@@ -12,12 +12,17 @@
...
 import time
 from importlib.metadata import version
 from pathlib import Path
 from pprint import pformat
 from typing import Any, Callable, Final, Optional, Union
 
+try:
+    import ray  # pyright: ignore[reportMissingImports]
+except ModuleNotFoundError:
+    ray = None
```
taenin, line 25: Instead of importing `ray` at the top of `train.py`, I'd suggest importing it inside the body of `_verl_train`. That way it's clear that this is a `verl` dependency, not an `oumi` dependency.

src/oumi/train.py
```diff
@@ -184,31 +189,72 @@ def _finalize_training_config(config: TrainingConfig) -> TrainingConfig:
...
+        trainer = partial_trainer()
+        trainer.train(resume_from_checkpoint=checkpoint_location)
+
+        logger.info("Training is Complete.")
+
+    ray.get(_run_verl_train.remote(partial_trainer, checkpoint_location))
+    logger.info(
+        "\n\n» We're always looking for feedback. "
+        "What's one thing we can improve? https://oumi.ai/feedback"
+    )
```
taenin, line 255: Make this a helper method like `log_feedback_request` so you don't duplicate the message for each trainer.

taenin (commented):
This may be a good opportunity to start adding some unit tests to the trainer, namely around setup before training.

src/oumi/train.py
wizeng23: See my last bullet point in PR description. I moved it to the `create_trainer_fn()` call with the other required keyword arguments (model, dataset, etc.). IMO the optional keyword arguments are for the ones that differ between trainers, whereas the ones that are shared by all trainers shouldn't be here.

src/oumi/train.py
wizeng23: Let me test more models and note them in the PR description. I already did some light testing and it worked fine.  I looked at every line before I moved them, and I don't believe most of them have interactions with each other, so reordering should be fine.

wizeng23:
I do agree unit tests will be important, but it's a non-trivial task that may take me close to a day given the amount of mocking that needs to happen (I still need to flesh out the tests for the verl trainer too). Given the urgency, and the fact that it was already this way before this PR, I wanted to submit this as-is for now (though with thorough testing to make sure nothing else broke). I'll create a task to add this test right afterwards though.

taenin (approved):

src/oumi/train.py
```diff
@@ -184,31 +185,81 @@ def _finalize_training_config(config: TrainingConfig) -> TrainingConfig:
...
+    remote Ray function.
+    """
+    try:
+        import ray  # pyright: ignore[reportMissingImports]
+    except ModuleNotFoundError:
+        ray = None
+    if ray is None:
+        raise RuntimeError(
+            "ray is not installed. Please install it with `pip install 'oumi[gpu]'`."
+        )
```
taenin, line 234: Minor nit: can you simply delete `ray = None` and move the `raise` statement into the `except`?

PR #1653, opened 2025-04-24
Update e2e tests to run on lambda

wizeng23:
# Description

- Swapped e2e tests to run on Lambda
- Swapped A100-80GB:4 with H100:4 since the former isn't available in Lambda
- Fixed a broken qwen e2e test, since the config `configs/recipes/vision/qwen2_vl_2b/sft/train.yaml` was moved to `configs/recipes/vision/qwen2_vl_2b/sft/full/train.yaml`.
    - I updated the test to run multi-GPU since that's how the config is running it; wasn't sure if there's a specific reason it was set to be a 24GB 1 GPU test before.
- Added a test for the parallel `configs/recipes/vision/qwen2_vl_2b/sft/lora/train.yaml` config.

I've tested that "A100:1" and "H100:4"GPU tests work. I've been unable to schedule "A100:4" tests work since there hasn't been enough capacity on Lambda, but the tests are covered by the H100:4 tests.

## Before submitting

### DIFF
--- a/tests/e2e/test_train_e2e.py
+++ b/tests/e2e/test_train_e2e.py
@@ -427,7 +427,7 @@ def test_train_text_1gpu_24gb(
     _test_train_impl(test_config=test_config, tmp_path=tmp_path, use_distributed=False)
 
 
-@requires_gpus(count=4, min_gb=40.0)
+@requires_gpus(count=4, min_gb=39.0)
 @pytest.mark.parametrize(
     "test_config",
     [
@@ -475,7 +475,7 @@ def test_train_multimodal_4gpu_40gb(test_config: TrainTestConfig, tmp_path: Path
     )
 
 
-@requires_gpus(count=1, min_gb=40.0)
+@requires_gpus(count=1, min_gb=39.0)
 @pytest.mark.parametrize(
     "test_config",
     [
--- a/tests/e2e/test_train_e2e.py
+++ b/tests/e2e/test_train_e2e.py
@@ -427,18 +427,19 @@ def test_train_text_1gpu_24gb(
     _test_train_impl(test_config=test_config, tmp_path=tmp_path, use_distributed=False)
 
 
-@requires_gpus(count=1, min_gb=24.0)
+@requires_gpus(count=4, min_gb=40.0)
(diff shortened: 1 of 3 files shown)


taenin (commented):
Let's make sure we can actually run the tests e2e on lambda before submitting this change. If there's inadequate capacity we can still run just these tests on GCP

wizeng23:
Confirmed that I ran all tests and they succeeded. While I couldn't schedule the A100:4 machine today, its tests are covered by the H100:4 machine. To remove redundance, we could consider not running the A100:4 machine or having the H100:4 machine skip the 40GB memory requirement tests.

tests/e2e/test_train_e2e.py
```diff
@@ -427,32 +427,34 @@ def test_train_text_1gpu_24gb(
     _test_train_impl(test_config=test_config, tmp_path=tmp_path, use_distributed=False)
 
 
-@requires_gpus(count=1, min_gb=24.0)
+@requires_gpus(count=4, min_gb=40.0)
```
nikg4, line 430: I previously had to use value 39 (instead of 40) because not 100% of memory is available (and/or  1024 vs 1000 difference? don't remember exactly)

tests/scripts/launch_tests.sh
```diff
@@ -2,13 +2,15 @@
 set -e
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-E2E_TEST_CONFIG="${SCRIPT_DIR}/gcp_e2e_tests_job.yaml"
+E2E_TEST_CONFIG="${SCRIPT_DIR}/lambda_e2e_tests_job.yaml"
```
nikg4, line 5: Consider adding a comment with GCP config.

Ideally, would be good to add a command line argument to select either Lambda or GCP for these tests

nikg4 (approved):

taenin (approved):

tests/scripts/launch_tests.sh
wizeng23: Added argument

Issue #1519, opened 2025-03-06
[Feature] Add loading spinner during interactive inference


wizeng23:
### Feature request

Since we don't support streaming yet, there is no visual indication that progress is being made for inference, especially for outputs that take multiple seconds to minutes to generate. Adding a loading spinner can help alleviate this.

### Motivation / references

Could use this existing _print_and_wait() fn from the job launcher: https://github.com/oumi-ai/oumi/blob/f4e90e794f31bcf87f906b6e496536c3e97c6614/src/oumi/cli/launch.py#L75. It would need to be modified so that the result of the function call is piped through and returned.

### Your contribution

I'm happy to consult and provide help for whoever takes this on

chenruizhg:
I can help work on this one.

wizeng23:
Thanks Jack! Let me know if you have any questions.

shivam096:
Hi @wizeng23 ,

I can take it up if it is available again.

nikg4:
@shivam096 it's all yours!

shivam096:
Thanks @nikg4 for assigning the issue.

@wizeng23  can you help me with a brief about the requirement and also it would really help me if you could also provide any resource that could help me with the task.

wizeng23:
Hi Shivam, the description for this Feature at the top should have most of the details you need. This is the call where the spinner should be displayed: https://github.com/oumi-ai/oumi/blob/47df09e9b36344988d7ca653a888305d8d9a02d9/src/oumi/infer.py#L58

swapnilbanduke:
Hi team! I’d love to take this up if it’s still available. I’ve gone through the code references in infer.py and launch.py and I’m planning to wrap the infer() call with a modified _print_and_wait() to show a loading spinner during interactive inference.

Let me know if there’s anything specific you'd like me to keep in mind before I get started. Thanks!

wizeng23:
Sure! Assigned the task to you.

panizniksiar:
Hi @wizeng23 ,
I’d love to contribute to this issue! Since my background is in UX/UI design, I can create a simple, user-friendly loading spinner concept in Figma (e.g., animated dots or a circular loader) that aligns with Oumi’s style. I’ll share both a minimal and a more creative option for the team to review, so developers can easily implement the one that best suits their needs.
Please let me know if you’d like me to proceed. I’m excited to help improve the user experience here! 

Thanks,
Paniz

wizeng23:
[...]

PR #1656, opened 2025-04-25
Add loading spinner for interactive inference

shivam096:
# Description

This pull request implements a loading spinner during interactive inference to provide visual feedback while the inference process is running. The _print_and_wait function from the job launcher (oumi/src/oumi/cli/launch.py) was reused and integrated into the infer_interactive function in oumi/src/oumi/infer.py. This enhancement improves the user experience by addressing the lack of progress indication for long-running inference tasks.

# Changes Made
### Integration of `_print_and_wait`:

The `_print_and_wait` function was modified to ensure the result of the wrapped task is returned.
It was then used in the infer_interactive function to wrap the infer call, displaying a loading spinner with the message "Running inference...".

### Testing and Validation:

The changes were tested locally using a sample inference configuration (configs/recipes/smollm/inference/135m_infer.yaml).
Verified that the spinner appears during inference and the results are displayed correctly.

## Related issues
https://github.com/oumi-ai/oumi/issues/1519

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/infer.py
+++ b/src/oumi/infer.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional
+import time
+import oumi.cli.cli_utils as cli_utils
+
+from typing import Any, Callable, Optional
 
 from oumi.builders.inference_engines import build_inference_engine
 from oumi.core.configs import InferenceConfig, InferenceEngineType
@@ -24,6 +27,7 @@ from oumi.core.types.conversation import (
     Role,
     Type,
 )
+from multiprocessing.pool import Pool
 from oumi.utils.logging import logger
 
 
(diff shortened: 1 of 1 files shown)


shivam096:
@wizeng23 I have add the fix for - https://github.com/oumi-ai/oumi/issues/1519.

Can you please take a look ?

src/oumi/infer.py
```diff
@@ -55,15 +85,27 @@ def infer_interactive(
         except (EOFError, KeyboardInterrupt):  # Triggered by Ctrl+D/Ctrl+C
             print("\nExiting...")
             return
-        model_response = infer(
+        # model_response = infer(
```
wizeng23, line 88: Delete unused code?

shivam096: Updated the code

taenin (changes requested):
Thanks for the contribution, Shivam! Our implementation has changed a bit since the original feature request was added, so I have a suggestion that will make this a bit cleaner:

I'd recommend copying the `infer_interactive(...)` (https://github.com/oumi-ai/oumi/blob/main/src/oumi/infer.py#L43) method into the CLI infer.py file (https://github.com/oumi-ai/oumi/blob/main/src/oumi/cli/infer.py)

Then in that function, you can simply use the `status` context manager around individual requests to the model:
```
with cli_utils.CONSOLE.status(message):
   model_response = infer(...)
```

The main goal here is we want to isolate the CLI code from the rest of our code base. We don't want src/oumi/infer.py importing from our CLI.

Let me know if you have any questions!

shivam096:
Thank you for updating me on the latest @taenin .

A follow-up question - Should I just copy it there or move the code there entirely since this is a CLI-specific implementation? In that case, I would remove the `infer_interactive` method from the current file to the CLI file. 

Is my understanding correct?

taenin:
> Thank you for updating me on the latest @taenin .
> 
> A follow-up question - Should I just copy it there or move the code there entirely since this is a CLI-specific implementation? In that case, I would remove the `infer_interactive` method from the current file to the CLI file.
> 
> Is my understanding correct?
> 
> Asking since the `infer_interactive` would call the infer method in the `src/oumi/infer.py` file, and other associated method calls that reside in the `infer.py` file such as `_get_engine` method

I would recommend copying the method itself--it's fine if we have it live in two locations.
The `_get_engine` method is actually just a wrapper around `build_inference_engine` which isn't unique to that file. You can simply import `build_inference_engine` in the CLI.

One additional point to be aware of: in the CLI, we want to lazy-load all modules so the CLI remains fast. This means any imports you add should live inside of the functions that need them.

wizeng23:
Closing due to inactivity.

Issue #1322, opened 2025-01-31
[Feature] Unsloth fine-tuning support


brragorn:
### Feature request

AdOdd4004 and Amazing Q request from Reddit

### Motivation / references

toward universal support for fine-tuning

### Your contribution

linguistic

OPE-977

taenin:
Thanks for the feature request! Here's the reddit post referenced in the request for posterity: https://www.reddit.com/r/LocalLLaMA/comments/1id3ak8/comment/m9yq4v4/

brragorn:
Thanks @taenin 
Let's post the update back to that thread if this gets taken up and completed

georgedouzas:
Hey everyone,

I’d love to help out with this! Just wondering:

- Any key things to keep in mind for integration?
- Is there a preferred approach, or is it open-ended?
- Any blockers or decisions that need sorting first?

taenin:
@georgedouzas I would recommend starting by understanding how both Oumi and Unsloth ingest datasets and models.

Ultimately to add unsloth support, you'll need to update this method: https://github.com/oumi-ai/oumi/blob/main/src/oumi/builders/training.py#L49

This will involve adding a `TrainerType.UNSLOTH`. You'll also need to ensure that we can support UNSLOTH with our https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/trainers/base_trainer.py#L21 (or update this class as needed).

It would also be good to understand how our standard training configurations can be ported to inputs for Unsloth: https://oumi.ai/docs/en/latest/api/oumi.core.configs.html#oumi.core.configs.TrainingConfig

georgedouzas:
Started working on this, please assign it to me. Thanks!

aniruddh-alt:
Hi @oelachqar, I could work on this issue if it is still open. I can build off the work from PR [#1657](https://github.com/oumi-ai/oumi/pull/1657). Please let me know if you're looking for contributors on this. Thanks!

PR #1657, opened 2025-04-26
Add optional support for Unsloth

georgedouzas:
# Description

This PR adds minimal support for the unsloth library without making it a required dependency of Oumi. The idea is to let users load models and tokenizers through Unsloth’s API if it’s available, without changing the trainer abstraction or adding a new TrainerType. Since Unsloth patches `transformers` and `trl` under the hood, it can work with the existing setup without needing a brand-new trainer.

Feedback on whether this approach makes sense would be appreciated, or let me know if you would prefer a more explicit integration — like introducing a dedicated trainer type for Unsloth even though it’s patching everything internally. There are also some dependency conflicts between Unsloth and Oumi at the moment.

## Related issues

#1322 

## Before submitting

## Reviewers


src/oumi/__init__.py
```diff
@@ -69,10 +69,27 @@
 
 from __future__ import annotations
 
+import sys
 from typing import TYPE_CHECKING, Any
 
 from oumi.utils import logging
 
+
```
oelachqar, line 77: Do we need this import here?

src/oumi/train.py
```diff
@@ -387,11 +389,18 @@ def train(
     use_peft = config.training.use_peft and config.peft
 
     # Build model.
-    model = build_model(
-        model_params=config.model,
-        peft_params=config.peft if use_peft else None,
-        **(additional_model_kwargs or {}),
-    )
+    if is_unsloth_model(config.model.model_name):
```
oelachqar, line 392: Given `build_model` already handles unsloth models, do we need this check here?

pyproject.toml
```diff
@@ -222,6 +223,9 @@ ban-relative-imports = "all" # Disallow all relative imports.
     "UP035",  # UP035 `typing.List` is deprecated, use `list` instead
     "UP006",  # UP006 Use `list` instead of `List` for type annotation
 ]
+"src/oumi/__init__.py" = [
```
oelachqar, line 226: Is this change required?

src/oumi/builders/models.py
```diff
@@ -377,6 +384,11 @@ def build_tokenizer(
     Returns:
         tokenizer: The tokenizer object built from the configuration.
     """
+    # Unsloth
+    if is_unsloth_model(model_params.model_name):
```
oelachqar, line 388: Can we just use the regular tokenizer?

oelachqar:
@georgedouzas thank you for the contribution! At a high level, the approach looks great. 

If possible, it would be great to:
- `build_unsloth_model_tokenizer` -> `build_unsloth_model`
- Use `build_tokenizer` to load the tokenizer. I believe this should already be working
- Move the `unsloth` specific logic outside of `oumi/core`. This could go in a utils file:    `get_unsloth_model_args`,
    `get_unsloth_peft_model_args`, `is_unsloth_model`, etc
- Add a flag in the model params for the model implementation: `hf`, `unsloth`

PR #1659, opened 2025-04-28
[verl] Populate verl config from Oumi config

wizeng23:
# Description

- Add a `verl_config_overrides` field in TrainingParams to allow setting any verl config value. This is similar to what `trainer_kwargs` does for HF/TRL trainers, and gives users flexibility to pass in fields we don't have in Oumi configs.
- Replace `TrainingParams` with `TrainingConfig` in verl trainer initialization since values in the latter are needed for verl.
- Populate verl config from some Oumi config fields and user-provided overrides.

Future work
- Add more Oumi -> verl config field converters. This PR only adds a few, but adding more is time-consuming. There are 200 verl config values, and I need to read into the verl codebase for many as similarly-named params in Oumi/trl/verl are not the same. For example, `max_prompt_length` in Oumi/trl will truncate prompts longer than that value, but in verl, it will result in an error for longer prompts.
- Add new fields in Oumi configs for verl param values we believe are important enough to get their own Oumi param field, as opposed to having be passed through `verl_config_overrides`.
- Add countdown training/gcp job configs.

## Related issues

Towards OPE-1172

## Before submitting

### DIFF
--- a/src/oumi/core/trainers/verl_trainer_config.yaml
+++ b/src/oumi/core/trainers/verl_trainer_config.yaml
@@ -208,7 +208,7 @@ trainer:
   balance_batch: True
   total_epochs: 30
   total_training_steps: null
-  project_name: oum_verl
+  project_name: oumi_verl
   experiment_name: null
   logger: [ 'console', 'wandb' ]
   log_val_generations: 0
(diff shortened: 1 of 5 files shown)


src/oumi/core/trainers/verl_trainer_config.yaml
```diff
@@ -204,8 +208,8 @@ trainer:
   balance_batch: True
   total_epochs: 30
   total_training_steps: null
-  project_name: verl_examples
-  experiment_name: gsm8k
+  project_name: oum_verl
```
nikg4, line 211: typo: `oumI_verl` ?

nikg4 (approved):

PR #1661, opened 2025-04-29
Add Qwen3 32B configs

wizeng23:
# Description

Add configs for Qwen3 32B model for LoRA, eval, and inference, derived from QwQ configs. Also updated the README and aliases.

Note that the model does thinking by default, and a future PR will extend support for non-thinking models. More configs will be added in future PRs as well.

Tested that all these configs work.

## Related issues

Towards OPE-1209

## Before submitting


taenin (approved):
Let's add configs for the smaller flavors as well when time allows

PR #1665, opened 2025-04-30
Add Qwen3 30B A3B configs

wizeng23:
# Description

Similar to #1661, and additionally added an e2e to test multi-GPu training of an MoE model (a configuration we don't have yet). Tested that they all work.

## Related issues

Towards OPE-1209

## Before submitting

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Explore the growing collection of ready-to-use configurations for state-of-the-a
 
 | Model | Example Configurations |
 |-------|------------------------|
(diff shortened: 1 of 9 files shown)


nikg4:
Is there a dependency on Oumi PyPi package release? 
If yes, we may want to hold documentation updates until it's released

wizeng23:
The dependency is transformers 4.51, which we already have! The Qwen team was smart and released the Qwen3 model architecture on HF weeks before their model release so it was available immediately.

PR #1672, opened 2025-05-06
Update the Oumi launcher and e2e tests to support runpod.

taenin:
# Description

This PR comes in two related parts:
1) Add support for `image_id` in the launcher config so we can use runpod containers with proper CUDA drivers
2) Update our e2e test script to run on RunPod

Note that runpod only supports A100-80GB per skypilot:
```
sky show-gpus --cloud runpod
COMMON_GPU  AVAILABLE_QUANTITIES  
A100-80GB   1, 2, 4, 8            
H100        1, 2, 4, 8            
L4          1, 2, 4, 8   
```

I've also removed H100s as they're pricier and harder to obtain.

## Related issues

N/A

## Before submitting

## Reviewers

### DIFF
--- a/tests/scripts/launch_tests.sh
+++ b/tests/scripts/launch_tests.sh
@@ -2,7 +2,7 @@
 set -e
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-E2E_TEST_CONFIG="${SCRIPT_DIR}/lambda_e2e_tests_job.yaml"
+E2E_TEST_CONFIG="${SCRIPT_DIR}/runpod_e2e_tests_job.yaml"
 echo "Using test config: ${E2E_TEST_CONFIG}"
 
 export E2E_CLUSTER_PREFIX="oumi-${USER}-e2e-tests"
@@ -11,7 +11,7 @@ export E2E_CLUSTER="" # Cloud provider to use (e.g., "lambda", "aws", etc.)
 
 # An alternative to H100 is A100-80GB, if they are available.
 # However, A100-80GB:4 isn't available in Lambda.
-declare -a accelerators_arr=("A100:1" "A100:4" "H100:4")
+declare -a accelerators_arr=("A100-80GB:1" "A100-80GB:4")
 
 # Reset the variable to make sure that CLI `--resources.use_spot` arg is not ignored.
 OUMI_USE_SPOT_VM=""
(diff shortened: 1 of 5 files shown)


tests/scripts/launch_tests.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-E2E_TEST_CONFIG="${SCRIPT_DIR}/lambda_e2e_tests_job.yaml"
+E2E_TEST_CONFIG="${SCRIPT_DIR}/runpod_e2e_tests_job.yaml"
```
wizeng23, line 5: We have the E2E_CLUSTER flag which controls which cloud the tests are run on. Since it seems there's not a big difference between runpod_e2e_tests_job.yaml and lambda_e2e_tests_job.yaml, could we just keep one? Could name it e2e_tests_job.yaml to be generic.

wizeng23 (approved):

tests/scripts/launch_tests.sh
```diff
@@ -11,7 +11,7 @@ export E2E_CLUSTER="" # Cloud provider to use (e.g., "lambda", "aws", etc.)
 
 # An alternative to H100 is A100-80GB, if they are available.
 # However, A100-80GB:4 isn't available in Lambda.
-declare -a accelerators_arr=("A100:1" "A100:4" "H100:4")
```
wizeng23, line 14: A problem with this is that we are losing test coverage by increasing the memory that our tests run on. I've noticed bugs where a config that used to not OOM on 40GB now does OOM, which is a regression; this is when upgrading package versions. IMO we should run on the exact same GPUs that the config was meant to run on. For example, `configs/examples/letter_counting/grpo/gcp_job.yaml` is run on A100:4, but instead would be run on A100-80GB:4 here.

Seems like the only option is GCP, but IMO the cost shouldn't be too high given this is infrequent.

tests/scripts/launch_tests.sh
taenin: RunPod actually requires a specific docker container to run, specified in `runpod_e2e_tests_job.yaml`. As such, the pip install instructions are also slightly modified.

tests/scripts/launch_tests.sh
taenin: I agree that GCP is the way to go. Let's reconvene after our discussion this afternoon.

wizeng23 (approved):

PR #1676, opened 2025-05-06
Mark an e2e test as is_lora

taenin:
# Description

Currently our e2e test `test_train_multimodal_lora_1gpu_40gb` is broken. Upon inspection, the output looks fine. However, the test is failing because it's looking for a `config.json` in what is actually a LoRA training job. The root cause is that the test is mislabeled. Adding `is_lora` should resolve the issue.

(edit): Fixed another test with the same issue.

## Related issues

N/A

## Before submitting

## Reviewers


wizeng23 (approved):
Thanks!

Issue #1628, opened 2025-04-11
[Feature] Gracefully handle saving OmegaConf configs containing non-primitives


wizeng23:
### Feature request

Our OmegaConfigs have fields for kwargs that can contain arbitrary values. When we try to save these as yaml files for logging purposes here: https://github.com/oumi-ai/oumi/blob/a59c9d7887c8a5de8c4e9c3263b549dea8fe4986/src/oumi/train.py#L232

They will trigger an error:
```
UnsupportedValueType: Value 'function' is not a supported primitive type
    full_key: trainer_kwargs.preprocess_logits_for_metrics
    reference_type=Dict[str, Any]
    object_type=dict
```

We should gracefully handle this. For example, we can parse through the omegaconf and find non-primitive fields, warn when trying to log these fields, and remove them. However, there is a greater question on how to ensure reproducibility of our jobs if the saved YAML configs of jobs don't contain all the required information to reproduce them.

### Motivation / references

See #1623 for context

### Your contribution

Happy to advise

abhiramvad:
Raised a pull request with suggested changes to this issue https://github.com/oumi-ai/oumi/pull/1684

PR #1684, opened 2025-05-14
feat: Updated BaseConfig class for non primitive fields

abhiramvad:
# Description
This is a PR for the issue [1628](https://github.com/oumi-ai/oumi/issues/1628). The `to_yaml` does not warn when trying to save non primitive data types (such as functions). This leads to errors while loading from saved yaml files.

I have added a check to gracefully check for non primitive types recursively. If present, they are removed from the config. The removed types and their paths are specified to the user via logs.

## Related issues

Fixes # (issue)
https://github.com/oumi-ai/oumi/issues/1628

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/base_config.py
+++ b/src/oumi/core/configs/base_config.py
@@ -18,7 +18,7 @@ import re
 from collections.abc import Iterator
 from io import StringIO
 from pathlib import Path
-from typing import Any, Optional, TypeVar, Union, cast
+from typing import Any, Optional, TypeVar, Union, cast, Set
 
 from omegaconf import OmegaConf
 
@@ -27,6 +27,60 @@ from oumi.core.configs.params.base_params import BaseParams
 T = TypeVar("T", bound="BaseConfig")
 
 _CLI_IGNORED_PREFIXES = ["--local-rank"]
+
+# Set of primitive types that OmegaConf can handle directly
+_PRIMITIVE_TYPES = {str, int, float, bool, type(None)}
+
+def _is_primitive_type(value: Any) -> bool:
+    """Check if a value is a primitive type that OmegaConf can handle.
+    
+    Args:
+        value: The value to check
+        
+    Returns:
+        bool: True if the value is a primitive type, False otherwise
+    """
+    if type(value) in _PRIMITIVE_TYPES:
+        return True
+    if isinstance(value, (list, dict)):
+        return True
+    return False
+
+def _handle_non_primitives(config: Any, path: str = "", removed_paths: Optional[Set[str]] = None) -> Any:
+    """Recursively process config object to handle non-primitive values.
+    
+    Args:
+        config: The config object to process
+        path: The current path in the config (for logging)
+        removed_paths: Set to track paths of removed non-primitive values
+        
+    Returns:
+        The processed config with non-primitive values removed
+    """
+    if removed_paths is None:
+        removed_paths = set()
+        
+    if _is_primitive_type(config):
+        return config
+        
+    if isinstance(config, list):
+        return [_handle_non_primitives(item, f"{path}[{i}]", removed_paths) 
+                for i, item in enumerate(config)]
+                
+    if isinstance(config, dict):
+        result = {}
+        for key, value in config.items():
+            current_path = f"{path}.{key}" if path else key
+            if _is_primitive_type(value):
+                result[key] = value
+            else:
+                removed_paths.add(current_path)
+                result[key] = None
+        return result
+        
+    # For any other type, remove it and track the path
+    removed_paths.add(path)
+    return None
 
 
 def _filter_ignored_args(arg_list: list[str]) -> list[str]:
@@ -57,8 +111,26 @@ def _read_config_without_interpolation(config_path: str) -> str:
 @dataclasses.dataclass
 class BaseConfig:
     def to_yaml(self, config_path: Union[str, Path, StringIO]) -> None:
-        """Saves the configuration to a YAML file."""
-        OmegaConf.save(config=self, f=config_path)
+        """Saves the configuration to a YAML file.
+        
+        Non-primitive values are removed and warnings are logged.
+        
+        Args:
+            config_path: Path to save the config to
+        """
+        config_dict = OmegaConf.to_container(self, resolve=True)
+        removed_paths = set()
+        processed_config = _handle_non_primitives(config_dict, removed_paths=removed_paths)
+        
+        # Log warnings for removed values
+        if removed_paths:
+            logging.warning(
+                "The following non-primitive values were removed from the config "
+                "as they cannot be saved to YAML:\n" + 
+                "\n".join(f"- {path}" for path in sorted(removed_paths))
+            )
+        
+        OmegaConf.save(config=processed_config, f=config_path)
 
     @classmethod
     def from_yaml(


src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +28,60 @@
...
+    Returns:
+        bool: True if the value is a primitive type, False otherwise
+    """
+    if type(value) in _PRIMITIVE_TYPES:
+        return True
+    if isinstance(value, (list, dict)):
+        return True
+    return False
+
+def _handle_non_primitives(config: Any, path: str = "", removed_paths: Optional[Set[str]] = None) -> Any:
```
wizeng23, line 49: Could removed_paths just be type `Set[str]`? I don't see a need for it to be None. You also don't need the first two lines of this fn

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +28,60 @@
...
+        
+    if isinstance(config, list):
+        return [_handle_non_primitives(item, f"{path}[{i}]", removed_paths) 
+                for i, item in enumerate(config)]
+                
+    if isinstance(config, dict):
+        result = {}
+        for key, value in config.items():
+            current_path = f"{path}.{key}" if path else key
+            if _is_primitive_type(value):
```
wizeng23, line 74: we have nested dicts in Oumi, but this function only checks one level in. Could this logic be recursive as well, just like for lists?

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +28,60 @@
...
+                result[key] = value
+            else:
+                removed_paths.add(current_path)
+                result[key] = None
+        return result
+
```
wizeng23, line 80: As an extension, could you try converting functions to strings, ex. using `inspect.getsource(func)`? You can try doing so inside a try except block.

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +28,60 @@
 
 _CLI_IGNORED_PREFIXES = ["--local-rank"]
 
+# Set of primitive types that OmegaConf can handle directly
+_PRIMITIVE_TYPES = {str, int, float, bool, type(None)}
```
wizeng23, line 32: Could you add all the types mentioned here as well: https://omegaconf.readthedocs.io/en/2.2_branch/structured_config.html#simple-types

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +28,60 @@
...
+        path: The current path in the config (for logging)
+        removed_paths: Set to track paths of removed non-primitive values
+        
+    Returns:
+        The processed config with non-primitive values removed
+    """
+    if removed_paths is None:
+        removed_paths = set()
+        
+    if _is_primitive_type(config):
```
wizeng23, line 63: Could you move this below the list and dict checks? Then, you don't even need the is_primitive_type function, as you can just replace it with `type(value) in _PRIMITIVE_TYPES`

wizeng23 (commented):
Thank you for creating this PR! I've left a couple comments. Could you also confirm that you've tested this PR, and that it works for non-primitives, such as functions?

oelachqar (changes requested):
Thank you @abhiramvad for the contribution! Given this changes a very sensitive class, would you be able to add unit tests to validate the logic ?

abhiramvad:
Sure, I need some time to add them. Will do that soon.

src/oumi/core/configs/base_config.py
abhiramvad: Made the change

src/oumi/core/configs/base_config.py
abhiramvad: Done

src/oumi/core/configs/base_config.py
abhiramvad: Done

src/oumi/core/configs/base_config.py
abhiramvad: Done

src/oumi/core/configs/base_config.py
abhiramvad: Done

abhiramvad:
> Thank you @abhiramvad for the contribution! Given this changes a very sensitive class, would you be able to add unit tests to validate the logic ?

I made the changes requested previously in comments. I have added few baseline tests, checking the test_config.py . Hope they align with what is needed. Kindly provide me comments on those. I'll update them soon.

wizeng23:
Hi @abhiramvad , could you please fix the precommit errors? You can verify it passes locally by running `pre-commit run --all-files --show-diff-on-failure`.

oelachqar:
Thank you @abhiramvad for the updates! I think this is very close to the finish line. It looks like some of the new tests are failing?

abhiramvad:
can we run the workflow again please?

wizeng23:
@abhiramvad could you fix the test error? You can run the checks locally with `pre-commit run --all-files --show-diff-on-failure`. Please re-request review when all tests pass locally.

abhiramvad:
@wizeng23 i fixed these errors. Here is a precommit check screenshot from my machine 
image


The tests work on my machine, and i also ran the CPU tests workflow on my fork. Lets hope it doesn't break here... https://github.com/abhiramvad/oumi/actions/runs/18120697923

abhiramvad:
@wizeng23 can you review please?

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +30,94 @@
...
 _CLI_IGNORED_PREFIXES = ["--local-rank"]
 
+# Set of primitive types that OmegaConf can handle directly
+_PRIMITIVE_TYPES = {str, int, float, bool, type(None), bytes, Path, Enum}
+
+
+def _is_primitive_type(value: Any) -> bool:
+    """Check if a value is of a primitive type that OmegaConf can handle."""
+    return (
+        isinstance(value, (str, int, float, bool, bytes))
```
wizeng23: nit: could you re-use _PRIMITIVE_TYPES here to reduce duplication?  I believe you still need the separate checks for Path and Enum though.

```
    return (
        type(value) in _PRIMITIVE_TYPES
        or isinstance(value, Path)
        or isinstance(value, Enum)
    )
```

src/oumi/core/configs/params/peft_params.py
```diff
@@ -325,7 +325,6 @@ def to_lora(self) -> LoraConfig:
             lora_alpha=self.lora_alpha,
             lora_dropout=self.lora_dropout,
             target_modules=target_modules,
-            target_parameters=self.lora_target_parameters,
```
wizeng23, line 328: Can you add this line back?

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +30,94 @@
...
+        or value is None
+        or isinstance(value, Path)
+        or isinstance(value, Enum)
+    )
+
+
+def _handle_non_primitives(config: Any, removed_paths, path: str = "") -> Any:
```
wizeng23: Can you add type to removed_paths?

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +30,94 @@
...
+
+    if hasattr(config, "__dataclass_fields__"):
+        result = {}
+        for field_name in config.__dataclass_fields__:
+            field_value = getattr(config, field_name)
+            current_path = f"{path}.{field_name}" if path else field_name
+            processed_value = _handle_non_primitives(
+                field_value, removed_paths, current_path
+            )
+            if processed_value is not None:
```
wizeng23, line 92: IMO if we redact a value, we should keep the key and have the value be None. You did something similar up in line 79.

src/oumi/core/configs/base_config.py
```diff
@@ -28,6 +30,94 @@
...
+                result[field_name] = processed_value
+            elif (
+                field_value is not None
+            ):  # Only track removal if original value was not None
+                removed_paths.add(current_path)
+        return result if result else None
```
wizeng23: Could you always return result here and not None? I'd rather have empty dict than a None value to properly reflect the object.

wizeng23 (changes requested):
Could you run an E2E test of this PR? Main things to test for:
1. non-primitive fields are properly redacted to None, or to the function source code if possible
2. All other fields are unaffected.

wizeng23:
Thank you for making this change @abhiramvad !

PR #1689, opened 2025-05-15
Support VLM-s with VERL_GRPO trainer

nikg4:
# Description

-- Add plumbing to pass `processor` object to `verl` trainer.
-- New VERL data path: If input data is single-turn SFT dataset, then auto-convert SFT Conversation to verl data format, and save to Parquet files.  
-- Add sample config for `Qwen/Qwen2.5-VL-3B-Instruct`.
-- Minor changes in Processor API e.g.,  allow `images` param to be used as a positional argument.
-- Sample run: https://wandb.ai/lema-train-test/oumi-train/runs/rbp7xqgj

## Related issues

Towards OPE-1222

## Before submitting

## Reviewers

### DIFF
--- a/configs/examples/grpo_verl_geometry3k/train.yaml
+++ b/configs/examples/grpo_verl_geometry3k/train.yaml
@@ -1,0 +1,84 @@
+# VERL GRPO training config for Countdown.
+#
+# Requirements:
+#   - Log into WandB (`wandb login`) or disable `enable_wandb`
+#
+# Usage:
+#   oumi train -c configs/examples/grpo_verl_countdown/train.yaml
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/train/train.html
+#   - Config class: oumi.core.configs.TrainingConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/training_config.py
+#   - Other training configs: configs/**/pretraining/, configs/**/sft/, configs/**/dpo/
+
+model:
+  model_name: "d1shs0ap/cognitive-behaviors-Llama-3.2-3B"
+
+data:
+  train:
+    datasets:
+      - dataset_name: "d1shs0ap/countdown-final"
+        split: "train"
+  validation:
+    datasets:
+      - dataset_name: "d1shs0ap/countdown-final"
+        split: "test"
+
+training:
+  trainer_type: "VERL_GRPO"
+  num_train_epochs: 5
+  save_steps: 150
+  eval_strategy: "steps"
+  eval_steps: 12
+
+  learning_rate: 1.0e-6
+  enable_gradient_checkpointing: True
+
+  reward_functions: ["countdown"]
+
+  grpo:
+    max_completion_length: 1024
+    use_vllm: True
+    temperature: 0.6
+    vllm_gpu_memory_utilization: 0.7
+
+  verl_config_overrides:
+    data:
+      train_files: "/home/cmu/countdown-curriculum/data/countdown/train-3-3.parquet"
+      val_files: "/home/cmu/countdown-curriculum/data/countdown/test-3-10.parquet"
+      train_batch_size: 128
+      val_batch_size: 256
+      max_prompt_length: 256
+      max_extrapolation_length: 2048
+      shuffle: False
+    actor_rollout_ref:
+      model:
+        use_remove_padding: True
+      actor:
+        ppo_mini_batch_size: 32
+        use_dynamic_bsz: True
+        gradients: "normal"
+        use_kl_loss: True
+        kl_loss_coef: 0.001
+        kl_loss_type: "low_var_kl"
+        entropy_coeff: 0
+      rollout:
+        tensor_model_parallel_size: 2
+        n: 8
+        enforce_eager: False
+        free_cache_engine: False
+        val_kwargs:
+          temperature: 0.6
+          n: 8
+          do_sample: True
+        max_num_batched_tokens: 16384
+      ref:
+        fsdp_config:
+          param_offload: True
+    trainer:
+      n_gpus_per_node: 8
+      nnodes: 1
+
+  output_dir: "output/grpo_verl_countdown"
+  enable_wandb: True
--- a/configs/examples/grpo_verl_geometry3k/train.yaml
+++ b/configs/examples/grpo_verl_geometry3k/train.yaml
@@ -51,7 +51,7 @@ training:
   learning_rate: 1.0e-6
   enable_gradient_checkpointing: True
 
-  reward_functions: ["countdown"]
+  reward_functions: [] # Use built-in reward func for data_source=="hiyouga/geometry3k".
 
   grpo:
     max_completion_length: 1024
--- a/configs/examples/grpo_verl_geometry3k/train.yaml
+++ b/configs/examples/grpo_verl_geometry3k/train.yaml
@@ -13,7 +13,8 @@
 #   - Other training configs: configs/**/pretraining/, configs/**/sft/, configs/**/dpo/
 
 model:
-  model_name: "Qwen/Qwen2.5-VL-3B-Instruct" #  "Qwen/Qwen2.5-VL-7B-Instruct"
+  # For 7B model "Qwen/Qwen2.5-VL-7B-Instruct", a node with 8 GPUs is recommmeneded.
+  model_name: "Qwen/Qwen2.5-VL-3B-Instruct"
 
 data:
   train:
@@ -26,7 +27,6 @@ data:
         dataset_kwargs:
           return_conversations: True
           processor_name: "Qwen/Qwen2.5-VL-3B-Instruct"
-          # limit: 4096 # Uncomment to limit dataset size!
           return_tensors: True
   validation:
     datasets:
@@ -38,7 +38,6 @@ data:
         dataset_kwargs:
           return_conversations: True
           processor_name: "Qwen/Qwen2.5-VL-3B-Instruct"
-          # limit: 4096 # Uncomment to limit dataset size!
           return_tensors: True
 
 training:
--- a/configs/examples/grpo_verl_geometry3k/train.yaml
+++ b/configs/examples/grpo_verl_geometry3k/train.yaml
@@ -64,8 +64,8 @@ training:
     data:
       # train_files: "/home/cmu/countdown-curriculum/data/countdown/train-3-3.parquet"
       # val_files: "/home/cmu/countdown-curriculum/data/countdown/test-3-10.parquet"
-      train_batch_size: 256 # 512
-      val_batch_size: 256 # 512
+      train_batch_size: 512
+      val_batch_size: 512
       max_prompt_length: 1024
       # max_extrapolation_length: 2048
       # shuffle: False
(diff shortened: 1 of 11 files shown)


src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+        )
         self._train_dataset = train_dataset
         self._eval_dataset = eval_dataset
-        # Sets self._train_filepath and self._val_filepath.
-        self._create_dataset_files()
-
+        self._processor = processor
+        # Detect what dataset post-processing function to use (if any).
+        process_fn = self._detect_dataset_process_fn()
+        # Generate files and  set self._train_filepath and self._val_filepath.
```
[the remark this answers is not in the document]
nikg4: updated

configs/examples/grpo_verl_geometry3k/train.yaml
```diff
@@ -0,0 +1,104 @@
...
+    data:
+      train_batch_size: 512
+      val_batch_size: 512
+      max_prompt_length: 1024
+      max_response_length: 2048
+      filter_overlong_prompts: True
+      truncation: "error"
+      image_key: "images"
+    algorithm:
+      use_kl_in_reward: False
```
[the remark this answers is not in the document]
nikg4: Added a comment with pointers to verl documentation. I think it's  good enough. I don't want to replicate  verl documentation here (may go out-of-date, and hard to maintain)

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -194,7 +365,7 @@ def _setup_verl_trainer(self):
                 "Please install it with 'pip install `oumi[gpu]`'."
             )
         self._verl_config = self._create_config()
-        logger.info(f"verl config: {self._verl_config}")
+        logger.info(f"verl config: {pformat(self._verl_config)}")
```
wizeng23, line 368: Does this work for you? I tried it and it didn't

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -233,6 +404,11 @@ def _setup_verl_trainer(self):
         self._verl_trainer = RayPPOTrainer(
             config=self._verl_config,
             tokenizer=tokenizer,
+            processor=(
+                # verl trainer uses private methods and properties of `transformers`
+                # processor, so we need to pass the raw processor here.
+                self._processor.raw_processor if self._processor is not None else None
```
wizeng23, line 410: nit: Consider moving this logic into the init fn, and have the self._processor you save be the raw processor. This keeps this initialization lighter, and makes it clear early on which processor you'll be using throughout this file.

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+    ) -> None:
         """Creates dataset files for verl in Parquet format.
 
         The Parquet files are saved to the Oumi cache directory.
         """
         train_file = self._cache_dir / "train.parquet"
-        self._train_dataset.to_parquet(train_file)
+        train_dataset = self._train_dataset
+        # Limit the max number of sub-processes
+        num_proc = min(8, os.cpu_count() or 1)
```
wizeng23, line 246: Why 8? Also, consider making this an optional arg

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+        }
+        return data
+
+    def _create_dataset_files(
+        self, process_fn: Optional[Callable[[dict, int, str, str], dict]] = None
+    ) -> None:
         """Creates dataset files for verl in Parquet format.
 
         The Parquet files are saved to the Oumi cache directory.
         """
```
wizeng23, line 242: nit: extend to document the process_fn

configs/examples/grpo_verl_geometry3k/train.yaml
```diff
@@ -0,0 +1,111 @@
...
+#   - Other training configs: configs/**/pretraining/, configs/**/sft/, configs/**/dpo/
+
+model:
+  # For 7B model "Qwen/Qwen2.5-VL-7B-Instruct", a node with 8 GPUs is recommmeneded.
+  model_name: "Qwen/Qwen2.5-VL-3B-Instruct"
+
+data:
+  train:
+    datasets:
+      - dataset_name: "hiyouga/geometry3k"
```
wizeng23, line 22: A quicker alternative to having the dataset processing logic in the verl trainer is registering this dataset, and having it output a dict in verl-compatible format. Ex: https://github.com/oumi-ai/oumi/blob/main/src/oumi/datasets/grpo/countdown.py.

My approach is only temporary though, as ideally we have something similar to what you did in this PR where we convert the data into a verl-compatible format somewhere outside the dataset code

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+        # Detect what dataset post-processing function to use (if any).
+        process_fn = self._detect_dataset_process_fn()
+        # Generate files and set self._train_filepath and self._val_filepath.
+        self._create_dataset_files(process_fn)
         self._setup_verl_trainer()
 
-    def _create_dataset_files(self) -> None:
+    def _detect_dataset_process_fn(
+        self,
+    ) -> Optional[Callable[[dict, int, str, str], dict]]:
```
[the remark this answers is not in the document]
wizeng23: Ditto throughout this file

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+
+        if len(images) > 0:
+            # TODO: Generalize. This only works for QwenVL 2.5, which is the only
+            # VLM supported by verl as of 2025-05-15.
+            if not prompt.startswith("<image>"):
+                prompt = "<image>" + prompt
+        return (prompt, images, answer)
+
+    @staticmethod
+    def _create_verl_data_entry_from_single_turn_conversation(
```
wizeng23, line 208: In general, I'm worried that the dataset preprocessing logic being added may be too hardcoded, and won't generalize well to the different types of datasets people may want to use here. This logic mainly seems to be for VLM examples which include a single prompt, answer, and image, right? Could we instead have this logic live in the registered geometry dataset, similar to what I did for Countdown? If VLM GRPO isn't supported by TRL yet, I think there isn't an issue with the processing logic being hardcoded for verl.

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -94,45 +101,206 @@ def __init__(
...
+        model_name = model_params.model_name
+
         # 1. Read verl default dict config from YAML.
         yaml_path = Path(__file__).parent / "verl_trainer_config.yaml"
         config = OmegaConf.load(yaml_path)
         config = cast(DictConfig, config)
 
         # 2. Set config values, ex. from Oumi config values
         config.algorithm.adv_estimator = "grpo"
+        config.data.tokenizer = model_name
```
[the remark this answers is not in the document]
nikg4: The fields is defined in [src/oumi/core/trainers/verl_trainer_config.yaml](https://github.com/oumi-ai/oumi/blob/2dcad2dab4a5a05cf6ee836a752d684897daf6c2/src/oumi/core/trainers/verl_trainer_config.yaml#L11

I'm just initializing it.

src/oumi/core/trainers/verl_grpo_trainer.py
nikg4: Added a  comment and TODO. 

Similar example: https://github.com/oumi-ai/oumi/blob/2dcad2dab4a5a05cf6ee836a752d684897daf6c2/src/oumi/core/datasets/base_map_dataset.py#L261

configs/examples/grpo_verl_geometry3k/train.yaml
nikg4: Ack

src/oumi/core/trainers/verl_grpo_trainer.py
nikg4: The added post-processing function is actually quite general:  it extracts question, answer,  and , optionally, images from any single-turn Conversation, and it can also be activated pretty deterministically  by checking the presence  of `conversation_json` key in input example.   The function will raise an error if a conversation doesn't match expectations (e.g., not single turn). 

So I think this solution  is robust and general enough  for now. We can always add new registry type/config params later when it becomes needed (note that adding more params and settings has its own drawbacks: the more params the more difficult it is to configure everything correctly)

src/oumi/core/trainers/verl_grpo_trainer.py
nikg4: updated

src/oumi/core/trainers/verl_grpo_trainer.py
nikg4: moved

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -166,6 +334,9 @@ def _create_config(self) -> DictConfig:
         config.trainer.experiment_name = training_params.run_name
         config.trainer.default_local_dir = training_params.output_dir
 
+        config.critic.model.path = model_name
```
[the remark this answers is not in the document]
nikg4: The reason I set these is because I wasn't comfortable seeing semi-random model names set in src/oumi/core/trainers/verl_trainer_config.yaml

Removed to remove confusion

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -166,6 +334,9 @@ def _create_config(self) -> DictConfig:
...
+        config.reward_model.model.path = model_name
```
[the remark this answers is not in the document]
nikg4: updated

src/oumi/core/trainers/verl_grpo_trainer.py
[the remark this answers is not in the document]
wizeng23: What does setting it do though? The experiment I was reproducing wasn't setting it.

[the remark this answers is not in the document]
nikg4: Removed this setting. it doesn't seem necessary

wizeng23 (approved):

nikg4:
> > Could you confirm that non-VLM examples still work, ex. countdown? You don't have to run the examples on 8 H100s as the config says. You can change the hparams to get it to run on a smaller instance
> 
> What hparams you recommend changing to make it run on smaller instance? If you tried it before, could you share sample script/param values?

Countdown jobs runs OK so far on a node with 4 GPU-s https://wandb.ai/lema-train-test/oumi-train/runs/bozgrcjw?nw=nwusernikolaig

(w/ some manual changes as discussed offline)

src/oumi/core/trainers/verl_grpo_trainer.py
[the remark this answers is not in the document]
nikg4: updated

PR #1698, opened 2025-05-23
 Make SFT datasets usable with GRPO_TRL trainer

nikg4:
# Description

-- Make SFT datasets compatible with `TRL_GRPO` trainer: Read single-turn SFT Conversation, and treat USER  text as `prompt` and ASSISTANT text as `completion` (requires `dataset_kwargs.return_conversations: True`). Raise an exception if Conversation isn't single-turn. I think it's better to rely on SF Conversation format when possible (vs `trl`-specific GRPO dataset format)
-- Refactor the helper function that is doing the above operation into `grpo_utils.py` (previously was a private method of VERL trainer)
-- Update `train.py` to invoke the data mapping op for `TRL_GRPO` trainer.
-- `BaseSftDataset`:  Add `return_conversations` flag and `@override`  annotation
-- TODO: Add unit tests.

Sample dataset config:

```
data:
  train:
    datasets:
      - dataset_name: "argilla/databricks-dolly-15k-curated-en"
        split: "train"
        dataset_kwargs:
          return_conversations: True
```

## Related issues

Towards OPE-1260, OPE-1259

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -47,6 +47,7 @@ class BaseSftDataset(BaseMapDataset, ABC):
         assistant_only: bool = False,
         response_template: Optional[str] = None,
         instruction_template: Optional[str] = None,
+        return_conversations: bool = False,
         **kwargs,
     ) -> None:
         """Initializes a new instance of the BaseSftDataset class."""
@@ -65,6 +66,7 @@ class BaseSftDataset(BaseMapDataset, ABC):
         self._assistant_only = assistant_only
         self._response_template = response_template
         self._instruction_template = instruction_template
+        self._return_conversations = return_conversations
 
         if self._assistant_only:
             self._verify_assistant_only_compatibility()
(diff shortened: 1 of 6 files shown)


src/oumi/core/datasets/base_sft_dataset.py
```diff
@@ -46,6 +47,7 @@ def __init__(
         assistant_only: bool = False,
         response_template: Optional[str] = None,
         instruction_template: Optional[str] = None,
+        return_conversations: bool = False,
```
wizeng23, line 50: How do you plan for this parameter to be used, especially from config-only?

src/oumi/utils/grpo_utils.py
```diff
@@ -0,0 +1,121 @@
...
+
+import os
+from typing import Union
+
+import datasets as hf_datasets
+
+from oumi.core.types.conversation import Conversation, Role
+
+
+def extract_question_images_answer_from_single_turn_conversation(
```
wizeng23, line 23: Thoughts on naming this prompt_images_completion? question and answer I'm assuming are specific to VLM datasets, while prompt and response is more generic. In the implementation of this function, I don't see any indication that the user message is necessarily a question.

wizeng23 (approved):
Could you confirm if you've tested that this works E2E?

nikg4:
> dataset_kwargs:

yes I tested it with the dataset listed in PR description and it  works!

src/oumi/core/datasets/base_sft_dataset.py
nikg4: Example is provided in PR description . 

This is already used for some VLM datasets: https://github.com/oumi-ai/oumi/blob/a1c07c409177c558c43911ad70a34ad6bd644532/configs/examples/grpo_verl_geometry3k/train.yaml#L28

src/oumi/utils/grpo_utils.py
nikg4: this makes sense. Renamed

PR #1704, opened 2025-05-30
Add support for additional exception types for remote inference engine, as well as fast failing for non-retryable status codes.

jgreer013:
# Description

Different inference APIs have different methods of producing errors, so in an effort to make our inference more robust, this PR attempts to increase coverage of possible error types.

In addition, I add support for fast failing on non-retryable error codes.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -22,7 +22,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
-from typing import Any, Optional, Union
+from typing import Any, Optional
 
 import aiofiles
 import aiohttp
@@ -48,19 +48,15 @@ from oumi.utils.conversation_utils import (
     convert_message_to_json_content_list,
     create_list_of_message_json_dicts,
 )
+from oumi.utils.http import (
+    get_failure_reason_from_non_retriable_error,
+    get_non_200_retriable_failure_reason,
+    is_non_retryable_status_code,
+)
 
 _AUTHORIZATION_KEY: str = "Authorization"
 _BATCH_PURPOSE = "batch"
 _BATCH_ENDPOINT = "/v1/chat/completions"
-
-# HTTP status codes that should not be retried
-_NON_RETRYABLE_STATUS_CODES = {
-    400,  # Bad Request
-    401,  # Unauthorized
-    403,  # Forbidden
-    404,  # Not Found
-    422,  # Unprocessable Entity
-}
 
 
 class BatchStatus(Enum):
@@ -211,9 +207,6 @@ class RemoteInferenceEngine(BaseInferenceEngine):
 
     _remote_params: RemoteParams
     """Parameters for running inference against a remote API."""
-
-    _unsuccessful_request_count: int = 0
-    """Number of failed requests."""
 
     def __init__(
         self,
@@ -419,45 +412,6 @@ class RemoteInferenceEngine(BaseInferenceEngine):
         if not remote_params.api_key:
             remote_params.api_key = self._remote_params.api_key
 
-    async def _get_failure_reason_from_non_retryable_error(
-        self,
-        response: aiohttp.ClientResponse,
-    ) -> str:
-        """Handle non-retryable errors."""
-        try:
-            response_json = await response.json()
-            if isinstance(response_json, list):
-                response_json = response_json[0]
-            error_msg = (
-                response_json.get("error", {}).get("message")
-                if response_json
-                else f"HTTP {response.status}"
-            )
-
-        except Exception:
-            error_msg = f"HTTP {response.status}"
-
-        failure_reason = error_msg
-        return f"Non-retryable error: {failure_reason}"
-
-    def _get_non_200_retryable_failure_reason(
-        self,
-        response: aiohttp.ClientResponse,
-        response_json: Union[dict[str, Any], list[dict[str, Any]]],
-    ) -> str:
-        """Handle non-200 retryable response."""
-        # Handle error response
-        if isinstance(response_json, list):
-            # If the response is a list, it is likely an error message
-            response_json = response_json[0]
-
-        error_msg = (
-            response_json.get("error", {}).get("message") if response_json else None
-        )
-        failure_reason = error_msg if error_msg else f"HTTP {response.status}"
-
-        return failure_reason
-
     async def _query_api(
         self,
         conversation: Conversation,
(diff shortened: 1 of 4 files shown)


src/oumi/inference/remote_inference_engine.py
```diff
@@ -407,6 +419,45 @@ def _set_required_fields_for_inference(self, remote_params: RemoteParams):
         if not remote_params.api_key:
             remote_params.api_key = self._remote_params.api_key
 
+    async def _get_failure_reason_from_non_retryable_error(
```
nikg4, line 422: consider moving these functions into some new utility library for reusability

They don't seem to depend on `self` i.e., static methods.

src/oumi/inference/remote_inference_engine.py
```diff
@@ -53,6 +53,15 @@
 _BATCH_PURPOSE = "batch"
 _BATCH_ENDPOINT = "/v1/chat/completions"
 
+# HTTP status codes that should not be retried
+_NON_RETRYABLE_STATUS_CODES = {
+    400,  # Bad Request
+    401,  # Unauthorized
+    403,  # Forbidden
+    404,  # Not Found
```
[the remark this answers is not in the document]
jgreer013: Not exhaustive but covers the most common ones:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

src/oumi/inference/remote_inference_engine.py
```diff
@@ -203,6 +212,9 @@ class RemoteInferenceEngine(BaseInferenceEngine):
     _remote_params: RemoteParams
     """Parameters for running inference against a remote API."""
 
+    _unsuccessful_request_count: int = 0
+    """Number of failed requests."""
```
taenin, line 216: It looks like this value is only incremented but never used. I'd recommend removing it.

tests/unit/inference/test_remote_inference_engine.py
```diff
@@ -595,21 +600,35 @@ def test_infer_online_fails(mock_asyncio_sleep):
...
+            payload={"error": {"message": "Internal server error"}},
+        )
+        m.post(
+            _TARGET_SERVER,
+            status=500,
+            payload={"error": {"message": "Internal server error"}},
+        )
+        m.post(
+            _TARGET_SERVER,
+            status=500,
```
taenin, line 631: It's better to test multiple error codes here rather than just the same one repeatedly.

tests/unit/inference/test_remote_inference_engine.py
```diff
@@ -571,15 +572,19 @@ def test_infer_online_empty():
 
 
 def test_infer_online_fails(mock_asyncio_sleep):
+    """Test that non-retryable errors (like 401) fail immediately without retries."""
```
taenin, line 575: It's better to update the name of the test to reflect what it's doing than to add a comment.

tests/unit/inference/test_remote_inference_engine.py
```diff
@@ -2219,3 +2260,235 @@ def callback(url, **kwargs):
             assert backoff_sleeps[1] == pytest.approx(
                 0.4
             )  # Second retry: base delay * 2
+
+
+@pytest.mark.asyncio
```
taenin, line 2265: Here and below: do these need to be `asyncio` tests? Can you call the public methods of RemoteInferenceEngine instead?

taenin (changes requested):

src/oumi/inference/remote_inference_engine.py
jgreer013: I'll remove it.

For now 0_0

src/oumi/inference/remote_inference_engine.py
```diff
@@ -499,32 +575,51 @@ async def _query_api(
...
                             failure_reason = (
-                                response_json.get("error", {}).get("message")
-                                if response_json
-                                else f"HTTP {response.status}"
+                                f"Failed to process successful response: {str(e)}"
                             )
-                            if attempt < remote_params.max_retries:
-                                continue
+                            self._unsuccessful_request_count += 1
+                            if attempt == remote_params.max_retries:
```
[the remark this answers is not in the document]
jgreer013: Done

src/oumi/inference/remote_inference_engine.py
jgreer013: Done

src/oumi/inference/remote_inference_engine.py
```diff
@@ -471,24 +522,49 @@ async def _query_api(
                         headers=headers,
                         timeout=remote_params.connection_timeout,
                     ) as response:
+                        # Check for non-retryable status codes first to fail fast.
+                        if response.status in _NON_RETRYABLE_STATUS_CODES:
```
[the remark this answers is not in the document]
jgreer013: Done

tests/unit/inference/test_remote_inference_engine.py
jgreer013: Done

tests/unit/inference/test_remote_inference_engine.py
jgreer013: Done!

src/oumi/utils/http.py
```diff
@@ -0,0 +1,67 @@
...
+    422,  # Unprocessable Entity
+}
+
+
+def is_non_retryable_status_code(status_code: int) -> bool:
+    """Check if a status code is non-retryable."""
+    return status_code in _NON_RETRYABLE_STATUS_CODES
+
+
+async def get_failure_reason_from_non_retriable_error(
```
taenin: We should be consistent in our spelling here. Either `non_retryable` or `non_retriable`, but not both.

src/oumi/utils/http.py
```diff
@@ -0,0 +1,67 @@
...
+    response: aiohttp.ClientResponse,
+) -> str:
+    """Handle non-retryable errors."""
```
taenin: Can we update this doc string to be more descriptive? Perhaps `Return a string describing the error for the provided response`.

It may be good to also document how this would work for responses that aren't errors (like 200s)

src/oumi/utils/http.py
```diff
@@ -0,0 +1,67 @@
...
+    """Handle non-200 retryable response."""
+    # Handle error response
+    if isinstance(response_json, list):
+        # If the response is a list, it is likely an error message
+        response_json = response_json[0]
+
+    error_msg = response_json.get("error", {}).get("message") if response_json else None
+    failure_reason = error_msg if error_msg else f"HTTP {response.status}"
+
+    return failure_reason
```
taenin: Is this method needed? Can you instead just call `get_failure_reason_from_non_retriable_error`?

src/oumi/utils/http.py
jgreer013: Done.

oelachqar (approved):

taenin (approved):

PR #1705, opened 2025-06-02
Feat: add falcon-e integration

younesbelkada:
# Description

This PR adds the fine-tuning support for [Falcon-E (Falcon-Edge)](https://huggingface.co/collections/tiiuae/falcon-edge-series-6804fd13344d6d8a8fa71130) models. Falcon-E series are a set of fine-tunable bitnet models. To correctly perform fine-tuning, one has to use `onebitllms`, a lightweight toolkit library to patch the models linear layers with `BitNetLinear`, as well as exposing a saving callback to correctly save the model in BitNet format after fine-tuning. 

Tested FFT and evaluation on a small 1B model: https://huggingface.co/tiiuae/Falcon-E-1B-Base

@wizeng23 @oelachqar

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,7 @@ repos:
         types: [python]
         pass_filenames: false
         args: [--warnings]
+        exclude: tests/
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.19.1
     hooks:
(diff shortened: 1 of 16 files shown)


configs/recipes/falcon_e/dpo/falcon_e_1b_instruct/dpo.yaml
```diff
@@ -0,0 +1,34 @@
+# Phi3 DPO train config.
+#
+# Usage:
+#   oumi train -c configs/recipes/phi3/dpo/train.yaml
```
oelachqar: nit: the config path seems incorrect

younesbelkada: yes good catch, let me modify it

younesbelkada:
```suggestion
#   oumi train -c configs/recipes/falcon_e/falcon_e_1b_instruct/dpo/dpo.yaml
```

PR #1706, opened 2025-06-02
[vision] Add option to process images individually

oelachqar:
# Description

- Add option to process images individually:
  - Some processors (e.g. Molmo) do not support processing a batch of images
  - This PR adds an option to process each image individually instead
- Also add more docstrings / explanations of each option

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/collators/vision_language_sft_collator.py
+++ b/src/oumi/core/collators/vision_language_sft_collator.py
@@ -12,6 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Vision-Language SFT collator for conversation-based multimodal training.
+
+This module provides a collator specifically designed for supervised fine-tuning (SFT)
+of vision-language models using conversation data.
+
+Unlike VisionLanguageCollatorWithPadding which expects pre-processed features,
+this collator works with raw conversation objects and handles the complete feature
+generation pipeline.
+
+Example:
+    >>> from oumi.builders import build_tokenizer
+    >>> from oumi.core.configs import ModelParams
+    >>> tokenizer = build_tokenizer(ModelParams(model_name="llava-hf/llava-1.5-7b-hf"))
+    >>> collator = VisionLanguageSftCollator(
+    ...     tokenizer=tokenizer,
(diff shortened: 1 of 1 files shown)


wizeng23 (approved):

src/oumi/core/collators/vision_language_sft_collator.py
```diff
@@ -101,9 +212,66 @@ def __call__(self, batch) -> dict[str, Any]:
...
+
+        Raises:
+            ValueError: If results have inconsistent keys or non-tensor values
+        """
+        if not results:
+            return {}
+
+        # Get keys from first result and verify consistency
+        expected_keys = set(results[0].keys())
+        for i, result in enumerate(results[1:], 1):
```
taenin: Minor nit: I had to verify that `results[1:]` wouldn't throw an index error when len(results) <= 1.
It is indeed safe, but feels brittle to me if this functionality changes in the future for python.

taenin (approved):
Would love to see tests added, but LGTM as they don't exist yet.

src/oumi/core/collators/vision_language_sft_collator.py
oelachqar: I'll update to make it clearer. Slice operations in python do not throw errors (unlike index)

src/oumi/core/collators/vision_language_sft_collator.py
```diff
@@ -101,9 +212,66 @@ def __call__(self, batch) -> dict[str, Any]:
...
+        collated = {}
+        for key in expected_keys:
+            values = [result[key] for result in results]
+
+            # Determine max variable dimensions based on feature type
+            # For multi-image models, we may need 2 variable dims (num_images, seq_len)
+            max_var_dims = 2 if self._allow_multi_image_inputs else 1
+
+            # Pad and stack tensors
+            collated[key] = pad_to_max_dim_and_stack(
```
nikg7, line 273: note that this simple padding logic isn't compatible with all  VLM models. Consider adding a note about it

(delegating everything to `processor` was the original motivation for adding this collator)

PR #1711, opened 2025-06-03
[tiny] fix pre-commits checks on a fresh install

oelachqar:
# Description

- `alpaca_eval` is an optional dependency, and if not installed users will get a type check error
- This PR adds the proper type ignore to mitigate the issue (not the most elegant fix, open to suggestions)
 

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/evaluation/backends/alpaca_eval.py
+++ b/src/oumi/core/evaluation/backends/alpaca_eval.py
@@ -19,6 +19,7 @@ from typing import Any
 
 try:
     import alpaca_eval  # pyright: ignore[reportMissingImports]
+    import alpaca_eval.evaluate  # pyright: ignore[reportMissingImports]
 except ImportError:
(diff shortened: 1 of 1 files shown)


wizeng23 (commented):
Any idea why it isn't an issue for vllm here: https://github.com/oumi-ai/oumi/blob/main/src/oumi/inference/vllm_inference_engine.py

Is it because Alpaca's code has type checking?

kaisopos (approved):

oelachqar:
> Any idea why it isn't an issue for vllm here: https://github.com/oumi-ai/oumi/blob/main/src/oumi/inference/vllm_inference_engine.py
> 
> Is it because Alpaca's code has type checking?

Updated with a similar approach -- much better!

nikg4 (approved):

PR #1714, opened 2025-06-03
Update Verl trainer to export models in HF format

nikg4:
# Description

-- Update Verl trainer to export model in HF format after successful training
-- Copy `model_merger.py` script from verl repo, and use it as a library
-- Tested using `configs/examples/grpo_verl_geometry3k/gcp_job.yaml`

Context: https://verl.readthedocs.io/en/latest/advance/checkpoint.html#convert-fsdp-and-megatron-checkpoints-to-huggingface-format-model

## Related issues

Fixes OPE-1313

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/trainers/verl_grpo_trainer.py
+++ b/src/oumi/core/trainers/verl_grpo_trainer.py
@@ -53,7 +53,7 @@ from oumi.core.processors.base_processor import BaseProcessor
 from oumi.core.tokenizers import BaseTokenizer
 from oumi.core.trainers.base_trainer import BaseTrainer
 from oumi.utils.logging import logger
-from oumi.utils.verl_model_merger import ModelMergerConfig
+from oumi.utils.verl_model_merger import FSDPModelMerger, ModelMergerConfig
 
 # Dataset processing function type. This function takes the following arguments:
 # 1. a dataset sample.
@@ -460,7 +460,7 @@ class VerlGrpoTrainer(BaseTrainer):
         temp_dir = Path(self._temp_output_dir)
         checkpoint_dir = temp_dir
 
-        ModelMergerConfig(
+        config = ModelMergerConfig(
             operation="merge",
             backend="fsdp",
             tie_word_embedding=False,
@@ -468,3 +468,5 @@ class VerlGrpoTrainer(BaseTrainer):
             hf_model_config_path=str(checkpoint_dir / "huggingface"),
             target_dir=str(final_dir),
         )
+        merger = FSDPModelMerger(config)
+        merger.merge_and_save()
--- a/src/oumi/core/trainers/verl_grpo_trainer.py
+++ b/src/oumi/core/trainers/verl_grpo_trainer.py
@@ -357,7 +357,7 @@ class VerlGrpoTrainer(BaseTrainer):
             config.trainer.logger.append("wandb")
         config.trainer.project_name = os.environ.get("WANDB_PROJECT", "oumi_verl")
         config.trainer.experiment_name = training_params.run_name
-        config.trainer.default_local_dir = str(self._temp_output_dir)
+        config.trainer.default_local_dir = str(self._temp_output_dir or "")
 
         # 3. Apply user overrides
         overrides_config = OmegaConf.create(training_params.verl_config_overrides)
--- a/src/oumi/core/trainers/verl_grpo_trainer.py
+++ b/src/oumi/core/trainers/verl_grpo_trainer.py
@@ -107,18 +107,19 @@ class VerlGrpoTrainer(BaseTrainer):
         )
         self._processing_class = processing_class
         self._oumi_config = copy.deepcopy(config)
-        self._final_output_dir: Optional[Path] = None
-        self._temp_output_dir: Optional[Path] = (
-            Path(self._oumi_config.training.output_dir)
+        self._final_output_dir: Optional[Path] = (
+            Path(self._oumi_config.training.output_dir).absolute().resolve()
             if self._oumi_config.training.output_dir
             else None
         )
-        if config.training.save_final_model:
-            self._final_output_dir = Path(self._oumi_config.training.output_dir)
-            if not self._final_output_dir:
-                raise ValueError("Output directory must be specified")
-            self._final_output_dir = self._final_output_dir.absolute().resolve()
-            self._temp_output_dir = self._final_output_dir / "verl_output"
+        self._temp_output_dir: Optional[Path] = (
+            self._final_output_dir / "verl_output" if self._final_output_dir else None
+        )
+
+        if not self._final_output_dir and config.training.save_final_model:
+            raise ValueError(
+                "Output directory must be specified when saving final model is enabled."
+            )
 
         # TODO: OPE-1192 - Support multiple reward functions.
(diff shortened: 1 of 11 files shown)


src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -106,6 +107,11 @@ def __init__(
         )
         self._processing_class = processing_class
         self._oumi_config = copy.deepcopy(config)
+        self._final_output_dir: Path = Path(self._oumi_config.training.output_dir)
```
oelachqar: Is this required even if saving models is disabled? e.g. `config.training.save_final_model=False` ?

nikg7: Updated

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -343,7 +357,7 @@ def _create_config(self) -> DictConfig:
             config.trainer.logger.append("wandb")
         config.trainer.project_name = os.environ.get("WANDB_PROJECT", "oumi_verl")
         config.trainer.experiment_name = training_params.run_name
-        config.trainer.default_local_dir = training_params.output_dir
+        config.trainer.default_local_dir = str(self._temp_output_dir or "")
```
wizeng23, line 361: What happens if this is set to empty string?

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -106,6 +107,19 @@ def __init__(
         )
         self._processing_class = processing_class
         self._oumi_config = copy.deepcopy(config)
+        self._final_output_dir: Optional[Path] = None
```
wizeng23: IMO it's a bit confusing if the directory structure varies depending on whether save_final_model is set. IMO the verl output should be in a consistent directory. What about this logic:
1. Verify training.output_dir is set, and error if not. 
2. Verl output goes to f"{training.output_dir}/verl_output"
3. if save_final_model is true, it's saved to training.output_dir

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -448,3 +464,53 @@ def save_model(self, config: TrainingConfig, final: bool = True) -> None:
             final: Whether this is the final model being saved during training.
         """
         pass
+
+    def _export_hf_model(self) -> bool:
+        """Exports the tuned model to HF format."""
```
wizeng23: Can you document what the return value is?

src/oumi/utils/verl_model_merger.py
```diff
@@ -0,0 +1,809 @@
+# Copyright 2025 - Oumi
```
wizeng23: IMO we shouldn't add the copyright header here if we didn't write this file. Consider deleting it and adding this file path  as an exception in `tests/unit/test_apache_license_header.py`

src/oumi/utils/verl_model_merger.py
```diff
@@ -0,0 +1,809 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Copied from the Verl script:
+# volcengine/verl/blob/main/scripts/model_merger.py
+# with minor modifications.
```
wizeng23: Consider mentioning the modifications here

src/oumi/utils/verl_model_merger.py
```diff
@@ -0,0 +1,809 @@
...
+
+"""This script is used to merge huggingface model and test verl checkpoints from FSDP and Megatron backends.
+
```
wizeng23, line 22: What about cases where we're neither using FSDP nor Megatron?

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -448,3 +464,53 @@ def save_model(self, config: TrainingConfig, final: bool = True) -> None:
...
+
+        logger.info(
+            f"Merging and exporting model from '{latest_checkpoint_dir}' "
+            f"to '{final_dir}' ..."
+        )
+
+        config = ModelMergerConfig(
+            operation="merge",
+            backend="fsdp",
+            tie_word_embedding=False,
```
wizeng23, line 515: A lot of 1B models, and possibly 3B models, tie word embeddings. Ideally we should determine if the model uses tied word embeddings to set this value here. This may be non-trivial, so consider leaving a TODO here to update this

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -448,3 +464,53 @@ def save_model(self, config: TrainingConfig, final: bool = True) -> None:
             final: Whether this is the final model being saved during training.
```
wizeng23, line 463: Considering calling self._export_hf_model() in this function. This function is called here: https://github.com/oumi-ai/oumi/blob/0fe0c7d8435aaa7c09f183e8454ad30a6321d801/src/oumi/train.py#L525. It'll only be called if save_final_model is True, which would simplify some of your logic around _final_output_dir and _temp_output_dir

src/oumi/core/trainers/verl_grpo_trainer.py
nikg7: Same as before: There is no change compared to the previous state of this logic in the file.

src/oumi/core/trainers/verl_grpo_trainer.py
nikg7: updated

src/oumi/core/trainers/verl_grpo_trainer.py
nikg7: updated

src/oumi/utils/verl_model_merger.py
nikg7: These are the only two supported options. Megatron isn't applicable, but I suspect FSDP option should work for 
 non-distributed training.

src/oumi/core/trainers/verl_grpo_trainer.py
nikg7: added TODO

src/oumi/utils/verl_model_merger.py
nikg7: added

src/oumi/utils/verl_model_merger.py
nikg7: Reverted to the original header. It's the same Apache 2.0 notice but mentions Bytedance instead of Oumi

src/oumi/core/trainers/verl_grpo_trainer.py
nikg7: moved

nikg7:
@wizeng23 PTAL  (prev feedback addressed)

taenin (approved):

nikg7:
Submitting this in the interest of time. Will (@wizeng23 ) please ping me offline for any follow-ups that may be remaining.

PR #1723, opened 2025-06-04
Fix Falcon H1 dependency setup

wizeng23:
# Description

- Switch from GCP to Lambda cluster for training jobs, as GCP's CUDA version of 12.2 results in errors with certain CUDA ops needed for SSMs.
- Update some package installs to new recommendations from Falcon team. I had to install a specific dill version to prevent a version dependency issue.
- Switch 0.5B training job to not use oumi distributed to avoid a bug

I've tested the 0.5B training and evaluation jobs to confirm they work.

## Related issues

Fixes OPE-1314

## Before submitting


taenin (approved):
I'd consider updating the eval jobs to also use Lambda for consistency as well. That way a user can run both training and eval without setting up a new cloud provider.

dhiaEddineRhaiem (approved):

oelachqar (approved):

PR #1724, opened 2025-06-04
Update inference to always write intermediate results to file.

jgreer013:
# Description

Some inference jobs may fail midway either due to GPU memory issues, rate limits, etc.

This PR updates inference to *always* write intermediate results to some scratch file, ensuring that no money is wasted during inference runs.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -16,7 +16,7 @@ import copy
 import time
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
 import jsonlines
 from hdrh.histogram import HdrHistogram
@@ -161,11 +161,14 @@ class BaseInferenceEngine(ABC):
                     conversations.append(conversation)
         return conversations
 
-    def _get_scratch_filepath(self, output_filepath: str) -> str:
+    def _get_scratch_filepath(self, output_filepath: Union[str, None]) -> str:
         """Returns a scratch filepath for the given output filepath.
 
         For example, if the output filepath is "/foo/bar/output.json", the scratch
         filepath will be "/foo/bar/scratch/output.json"
+
+        If no output filepath is provided, a temporary file is used and placed in the
+        current working directory under the name "tmp/temp_inference_output.jsonl".
 
         Args:
             output_filepath: The output filepath.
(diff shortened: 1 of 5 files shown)


src/oumi/core/inference/base_inference_engine.py
```diff
@@ -161,23 +161,29 @@ def _read_conversations(self, input_filepath: str) -> list[Conversation]:
...
         Returns:
             str: The scratch filepath.
         """
-        original_filepath = Path(output_filepath)
-        return str(original_filepath.parent / "scratch" / original_filepath.name)
+        if output_filepath is not None:
+            original_filepath = Path(output_filepath)
+            return str(original_filepath.parent / "scratch" / original_filepath.name)
+
+        return "tmp/temp_inference_output.jsonl"
```
oelachqar: What if `tmp` does not exist?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -161,23 +161,29 @@ def _read_conversations(self, input_filepath: str) -> list[Conversation]:
                     conversations.append(conversation)
         return conversations
 
-    def _get_scratch_filepath(self, output_filepath: str) -> str:
+    def _get_scratch_filepath(self, output_filepath: Union[str, None]) -> str:
```
taenin: Nit, here and elsewhere: I'd recommend using `Optional[str]` over `Union[str, None]`. Optional is syntactic sugar for the latter, and it's a bit more concise.

src/oumi/core/inference/base_inference_engine.py
taenin: Do we need to prepend the current working directory to this string?

taenin (approved):

src/oumi/core/inference/base_inference_engine.py
jgreer013: Good point, done

src/oumi/core/inference/base_inference_engine.py
jgreer013: Knew I was forgetting something!

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -161,23 +161,29 @@ def _read_conversations(self, input_filepath: str) -> list[Conversation]:
...
         Returns:
             str: The scratch filepath.
         """
-        original_filepath = Path(output_filepath)
-        return str(original_filepath.parent / "scratch" / original_filepath.name)
+        if output_filepath is not None:
+            original_filepath = Path(output_filepath)
+            return str(original_filepath.parent / "scratch" / original_filepath.name)
+
+        return str(Path.cwd() / "tmp" / "temp_inference_output.jsonl")
```
oelachqar, line 183: A convention I've seen is `filename.tmp` instead of `scratch/filename`

src/oumi/core/inference/base_inference_engine.py
jgreer013: Added current working directory, but the `_save_conversation_to_scratch` method ensures that the directory is made before writing the file.

src/oumi/core/inference/base_inference_engine.py
jgreer013: Would that be "filename.jsonl.tmp" or "filename.tmp" or "filename.tmp.jsonl"?

I'd rather maintain the jsonl suffix as this file is a relevant recoverable artifact for users

PR #1734, opened 2025-06-05
Update inference to resume from temporary result file when possible

jgreer013:
# Description

Inference jobs write results to a temporary file throughout, but we should also be resuming from said files when restarting the job.

This PR loads the scratch file before inference, skips any inputs that have already completed, runs inference, then reads all results from the scratch file, cleans up the scratch file, then returns.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -14,6 +14,7 @@
 
 import copy
 import time
+import uuid
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Optional
@@ -100,22 +101,66 @@ class BaseInferenceEngine(ABC):
                     "the generation parameters that the engine was initialized with."
                 )
 
-        result: list[Conversation] = []
-        start_time = time.perf_counter()
-        histogram: Optional[HdrHistogram] = None
+        output_path = inference_config.output_path if inference_config else None
+        completed_conversations = self._load_from_scratch(output_path)
+
+        # Load input conversations either from file or use provided input
+        conversations_to_process: list[Conversation] = []
         if input is not None:
-            histogram = self._latency_histogram_online
-            result = self.infer_online(input, inference_config)
+            conversations_to_process = input
         elif inference_config and inference_config.input_path is not None:
-            histogram = self._latency_histogram_from_file
-            result = self.infer_from_file(inference_config.input_path, inference_config)
+            conversations_to_process = self._read_conversations(
+                inference_config.input_path,
+            )
         else:
             raise ValueError(
                 "One of input or inference_config.input_path must be provided."
             )
-        histogram.record_value((time.perf_counter() - start_time) * 1e3)
-        self._maybe_log_latency_histogram(histogram)
-        return result
+
+        for i, conversation in enumerate(conversations_to_process):
+            if conversation.conversation_id is not None:
+                continue
+
+            id_name = str(i)
+            # Generate a deterministic conversation ID based on index if none exists
+            last_message = conversation.last_message()
+            if last_message and isinstance(last_message.content, str):
+                id_name = str(i) + "_" + last_message.content
+
+            conversation.conversation_id = str(
+                uuid.uuid5(
+                    uuid.NAMESPACE_DNS,
+                    id_name,
+                )
+            )
+
+        # Filter out already completed conversations
+        remaining_conversations = self._filter_incomplete_conversations(
+            conversations_to_process, completed_conversations
+        )
+
+        if not remaining_conversations:
+            logger.info("All conversations have already been processed")
+            return completed_conversations
+
+        if len(remaining_conversations) < len(conversations_to_process):
+            logger.info(
+                f"Found {len(completed_conversations)} completed conversations. "
+                f"Processing remaining {len(remaining_conversations)} conversations."
+            )
+
+        # Run inference only on remaining conversations
+        if remaining_conversations:
+            start_time = time.perf_counter()
+            histogram = self._latency_histogram_online
+            _ = self.infer_online(remaining_conversations, inference_config)
+            histogram.record_value((time.perf_counter() - start_time) * 1e3)
+            self._maybe_log_latency_histogram(histogram)
+
+        # Load all results from scratch to get both previously completed and new results
+        final_results = self._load_from_scratch(output_path)
+        self._cleanup_scratch_file(output_path)
+        return final_results
 
     def _maybe_log_latency_histogram(self, histogram: Optional[HdrHistogram]) -> None:
         """Logs the histogram if it is not None.
@@ -160,6 +205,54 @@ class BaseInferenceEngine(ABC):
                     conversation = Conversation.from_json(line)
                     conversations.append(conversation)
         return conversations
+
+    def _load_from_scratch(
+        self,
+        output_filepath: Optional[str],
+    ) -> list[Conversation]:
+        """Loads conversations from a scratch file.
+
+        Args:
+            output_filepath: The path to the output file. This is used to determine the
+                location of the scratch file.
+
+        Returns:
+            list[Conversation]: A list of conversations loaded from the scratch file.
+        """
+        scratch_filepath = self._get_scratch_filepath(output_filepath)
+        if not Path(scratch_filepath).exists():
+            return []
+
+        conversations = []
+        with jsonlines.open(scratch_filepath, mode="r") as reader:
+            for line in reader:
+                conversations.append(Conversation.from_dict(line))
+        return conversations
+
+    def _filter_incomplete_conversations(
+        self,
+        input_conversations: list[Conversation],
+        completed_conversations: list[Conversation],
+    ) -> list[Conversation]:
+        """Filters out conversations that have already been completed.
+
+        Args:
+            input_conversations: List of conversations to run inference on
+            completed_conversations: List of conversations already completed
+
+        Returns:
+            list[Conversation]: List of conversations that still need inference results
+        """
+        completed_ids = {
+            conv.conversation_id
+            for conv in completed_conversations
+            if conv.conversation_id is not None
+        }
+        return [
+            conv
+            for conv in input_conversations
+            if conv.conversation_id not in completed_ids
+        ]
 
     def _get_scratch_filepath(self, output_filepath: Optional[str]) -> str:
         """Returns a scratch filepath for the given output filepath.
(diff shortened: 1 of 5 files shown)


src/oumi/core/inference/base_inference_engine.py
```diff
@@ -180,7 +299,15 @@ def _get_scratch_filepath(self, output_filepath: Optional[str]) -> str:
...
-        return str(Path.cwd() / "tmp" / "temp_inference_output.jsonl")
+        model_params = self._model_params
+        model_params_str = json.dumps(dataclasses.asdict(model_params))
+        generation_params = self._generation_params
+        generation_params_str = json.dumps(dataclasses.asdict(generation_params))
+        inference_hash = hashlib.sha256(
+            f"{model_params_str}_{generation_params_str}_{self._dataset_hash}".encode()
+        ).hexdigest()
+
+        return str(Path.cwd() / "tmp" / f"temp_inference_output_{inference_hash}.jsonl")
```
wizeng23: I know you didn't originally add this line, but given this code could be invoked from anywhere (including a repo outside Oumi), this doesn't seem like the best path for me. What about `~/.cache/oumi`, like here: https://github.com/oumi-ai/oumi/blob/5046f71bbb6bd45b00ce604fde9f490cf2bf225e/src/oumi/core/trainers/verl_grpo_trainer.py#L132

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,91 @@ def infer(
...
+        inference_results = self.infer_online(remaining_conversations, inference_config)
         histogram.record_value((time.perf_counter() - start_time) * 1e3)
         self._maybe_log_latency_histogram(histogram)
-        return result
+
+        if len(inference_results) == len(conversations_to_process):
+            final_results = inference_results
+        else:
+            # Incomplete inference results were saved to scratch file.
+            # Load all results from scratch to get all results.
```
wizeng23, line 188: I don't see the scratch file logic in this PR, but just to confirm, are results appended to the scratch file? i.e. if there were some previous results in the scratch file, they'll stack with the new results that were created?

src/oumi/core/inference/base_inference_engine.py
taenin: +1, this sounds reasonable to me as well

src/oumi/core/inference/base_inference_engine.py
jgreer013: Correct! Previous PR introduced that logic, this is primarily meant to read from the scratch file if it exists

src/oumi/core/inference/base_inference_engine.py
jgreer013: Done.

jgreer013:
> Could you confirm that this was tested? Ideally both a local and remote inference engine are tested.

Can confirm with local from manual exception throwing:
```
[base_inference_engine.py:158] Found 2 completed conversations. Processing remaining 0 conversations.
```

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,93 @@ def infer(
...
                 "One of input or inference_config.input_path must be provided."
             )
+
+        for i, conversation in enumerate(conversations_to_process):
+            if conversation.conversation_id is not None:
+                continue
+
+            id_name = str(i)
+            # Generate a deterministic conversation ID based on index if none exists
+            last_message = conversation.last_message()
```
taenin: Is it sufficient to hash only the last message? What if the system prompt is changing, but the user query isn't?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,93 @@ def infer(
...
+            if last_message and isinstance(last_message.content, str):
+                content_hash = hashlib.sha256(last_message.content.encode()).hexdigest()
+                id_name = str(i) + "_" + content_hash
+
+            conversation.conversation_id = str(
+                uuid.uuid5(
+                    uuid.NAMESPACE_DNS,
+                    id_name,
+                )
+            )
```
taenin, line 150: Why are we hashing again?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,93 @@ def infer(
...
+
+        dataset_hash = ""
+        if len(conversations_to_process) > 0:
+            row_hashes = [
+                hashlib.sha256(c.to_json().encode()).hexdigest()
+                for c in conversations_to_process
+            ]
+            dataset_hash = hashlib.sha256(",".join(row_hashes).encode()).hexdigest()
+        self._dataset_hash = dataset_hash
```
taenin, line 162: I'd strongly suggest adding a comment here explaining what you're calculating and why. It wasn't immediately clear to me why we need to keep a hash of the dataset as it's not used in this function.

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -161,6 +237,51 @@ def _read_conversations(self, input_filepath: str) -> list[Conversation]:
...
 
+    def _load_from_scratch(self, output_filepath: Optional[str]) -> list[Conversation]:
+        """Loads conversations from a scratch file.
+
+        Args:
+            output_filepath: The path to the output file. This is used to determine the
+                location of the scratch file.
+
+        Returns:
+            list[Conversation]: A list of conversations loaded from the scratch file.
```
taenin: It'd be good to specify that it returns the empty array when the scratch file doesn't exist.

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,93 @@ def infer(
...
+                inference_config.input_path,
+            )
         else:
             raise ValueError(
                 "One of input or inference_config.input_path must be provided."
             )
+
+        for i, conversation in enumerate(conversations_to_process):
+            if conversation.conversation_id is not None:
+                continue
```
taenin: Will we have problems if the user has defined their own conversation ID, and multiple rows have the same ID? If so, we should likely catch it early.

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -180,7 +301,16 @@ def _get_scratch_filepath(self, output_filepath: Optional[str]) -> str:
...
+        model_params = self._model_params
+        model_params_str = json.dumps(dataclasses.asdict(model_params))
+        generation_params = self._generation_params
+        generation_params_str = json.dumps(dataclasses.asdict(generation_params))
+        inference_hash = hashlib.sha256(
+            f"{model_params_str}_{generation_params_str}_{self._dataset_hash}".encode()
+        ).hexdigest()
+
+        path_prefix = Path.home() / ".cache" / "oumi" / "tmp"
+        return str(path_prefix / f"temp_inference_output_{inference_hash}.jsonl")
```
taenin, line 328: How long will this file name be? Some file systems have restrictions on maximum path length, so good to make sure this isn't crazy long.

tests/unit/inference/test_base_inference_engine.py
```diff
@@ -0,0 +1,331 @@
...
+                    input=conversations,
+                    inference_config=inference_config,
+                )
+
+        # Verify scratch file was cleaned up despite the error
+        scratch_path = Path(temp_dir) / "scratch" / "output.jsonl"
+        assert not scratch_path.exists()
+
+
+def test_empty_scratch_file(mock_engine):
```
taenin, line 307: What about the case where the scratch file contains all of the requests? Do we handle that properly?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -100,22 +105,93 @@ def infer(
...
+
+        # Load input conversations either from file or use provided input
+        conversations_to_process: list[Conversation] = []
         if input is not None:
-            histogram = self._latency_histogram_online
-            result = self.infer_online(input, inference_config)
+            conversations_to_process = input
         elif inference_config and inference_config.input_path is not None:
-            histogram = self._latency_histogram_from_file
-            result = self.infer_from_file(inference_config.input_path, inference_config)
```
taenin, line 111: Does this change make `infer_from_file` obsolete for all classes?

src/oumi/core/inference/base_inference_engine.py
jgreer013: The goal is mainly this:
1. If no conversation id exists, we need to create one, so that we know not to retry inference on a completed conversation
2. Since the intention is to be able to restart after a failed job, we need to be able to consistently reproduce the id between runs.
3. While we could do just a basic integer of the index of the conversation, this could cause problems when you want to identify unique samples or merge datasets (think 2 inference runs)
4. By hashing the content of the conversation with the index, we guarantee that the identifier is strongly unique which will make identifying samples easier.

src/oumi/core/inference/base_inference_engine.py
jgreer013: Done.

src/oumi/core/inference/base_inference_engine.py
jgreer013: Good point, catching

src/oumi/core/inference/base_inference_engine.py
jgreer013: Fair point, let me hash the concatenation of all messages in the conv instead

src/oumi/core/inference/base_inference_engine.py
jgreer013: Done.

src/oumi/core/inference/base_inference_engine.py
jgreer013: The hash is 64 characters long (256 / 4 bits for hex = 64 hex characters), and "~/.cache/oumi/tmp/temp_inference_output_" is 40 characters long, so 104 characters in total. Most path limits are 200+, so I think we're fine so long as the home directory isn't 100+

tests/unit/inference/test_base_inference_engine.py
jgreer013: Yes, but added a test to validate.

src/oumi/core/inference/base_inference_engine.py
jgreer013: Technically yes, I have a follow-up PR which marks it as deprecated. Future classes should only have to implement `_infer_online`

PR #1739, opened 2025-06-06
deps: Update vLLM to 0.8.3

wizeng23:
# Description

vLLM 0.8 makes their v1 inference engine the default, which should perform better in many scenarios: https://github.com/vllm-project/vllm/releases/tag/v0.8.0. It also has improvements for multimodal inference.

This requires updating torch to 2.6.0 as well. It's possible to update to 0.9 as well, but that results in long build times for flash-attn on GCP since there isn't a prebuilt wheel for CUDA 12.2, which GCP uses.

Note that upgrading to torch 2.6 updates the return type of `nn.Module.__getattr__()`  to `Union[Tensor, "Module"]` instead of `Any`. This essentially disabled type checking anywhere we access properties of `nn.Module`, which is pretty common, given we use this type instead of `transformers.PreTrainedModel` throughout. 

There is an issue with Llama 3.2 11B FFT (`configs/recipes/vision/llama3_2_vision/sft/11b_full/gcp_job.yaml`), which is not present for 90B FFT or 11B LoRA. Since this is a small impact of scope, I'm going to install the old version of torch in the job config and proceed with the upgrade.

Tested integration tests.

## Related issues

Fixes OPE-1263
Towards OPE-1328

## Before submitting


kaisopos (approved):

tests/unit/builders/test_callbacks.py
```diff
@@ -38,7 +37,7 @@ def test_build_training_callbacks_mfu_callback():
 def test_build_training_callbacks_no_cuda(mock_logger_warning):
     config = TrainingConfig()
     config.training.include_performance_metrics = True
-    model = torch.nn.Module()
+    model = torch.nn.Sequential(torch.nn.Linear(10, 10)).to(dtype=torch.bfloat16)
```
optas: do we care about bfloat not supported on most vanilla cpu's (falling back to float32)?

wizeng23: Good point! I was replicating what was originally in these tests, but setting the type explicitly isn't needed

PR #1741, opened 2025-06-09
Add ability to customize HF model config via model.model_kwargs

oelachqar:
# Description

- Add ability to customize HF model config via model.model_kwargs

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


src/oumi/core/configs/internal/supported_models.py
```diff
@@ -80,16 +80,22 @@ def _create_my_model_config() -> InternalModelConfig:
 from oumi.utils.logging import logger
 
 
+@functools.cache
```
wizeng23, line 83: Is the duplicate line intended?

oelachqar: Good catch, fixed!

PR #1745, opened 2025-06-09
Set infer_online and infer_from_file to private

jgreer013:
# Description

Inference engines had .infer, .infer_online, and .infer_from_file as separate public methods, which makes input preparation challenging due to 3 separate entry points for largely the same code. This sets infer_online and infer_from_file to private to consolidate access to only infer and infer_batch

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers


oelachqar:
@jgreer013 given this is a breaking change -- could you leave the public methods + log a deprecation warning? We'll remove in a future version

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -423,26 +425,6 @@ def infer_online(
         """
         raise NotImplementedError
```
wizeng23, line 426: Why is infer_from_file deleted instead of being made private? How will we support inference from a file?

jgreer013: infer supports inference from a file inherently - it adapts based on whether or not data was passed in to the function call directly. infer ultimately has switch to using the _infer_online method entirely as a result (it handles loading data from file beforehand, which is the only difference _infer_from_file had)

jgreer013: Deleted this from the base class to reiterate that it doesn't need to be overridden by subclasses (i.e. they don't need to implement it)

taenin: Per Oussama's comment above, wasn't the plan here to mark them as deprecated instead of deleting them?

PR #1759, opened 2025-06-13
Judge API V2 | Implement CLI

kaisopos:
# Description

Judge API V2: Implement the CLI

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/cli/judge_v2.py
+++ b/src/oumi/cli/judge_v2.py
@@ -1,0 +1,132 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


wizeng23 (commented):
I might be missing context, but to me it feels weird to add an oumi verb called judge_v2.

- Is there a reason we can't just replace the old judge verb logic with this one?
- If we still want some of the functionality of the old judge verb, could we either add this new logic under the same verb, controlled by a flag, or give this verb a more descriptive name than just v2?

oelachqar (approved):

kaisopos:
> I might be missing context, but to me it feels weird to add an oumi verb called judge_v2.
> 
> * Is there a reason we can't just replace the old judge verb logic with this one?
> * If we still want some of the functionality of the old judge verb, could we either add this new logic under the same verb, controlled by a flag, or give this verb a more descriptive name than just v2?

Restricted Judge V2 under the env flag `OUMI_EXPERIMENTAL_JUDGE_V2`, as discussed offline.
The plan is to add a deprecation notice for V1 and fully replace with V2 in 1-2 versions.

PR #1761, opened 2025-06-13
Automatically tail SkyPilot logs

wizeng23:
# Description

If the cloud is a SkyPilot cloud and we don't run in detach mode, replace the default "Running job" polling message with tailing the SkyPilot logs. This is in line with SkyPilot's default behavior. IMO this doesn't need to be configurable as I've never run a job and not wanted to see the logs yet.

Manually tested.

## Related issues

Fixes OPE-322

## Before submitting


taenin (approved):
Looks good for now, but eventually we should add support for this in all clouds :)

PR #1766, opened 2025-06-17
[verl] Support resuming from checkpoint

wizeng23:
# Description

- For verl trainer, don't run our logic to find latest checkpoint. Verl will handle this.
- Remove resume_from_checkpoint fn for train() function of verl trainer. We instead set it in the class initialization.

## Related issues

Towards OPE-1260

## Before submitting

### DIFF
--- a/src/oumi/core/configs/params/training_params.py
+++ b/src/oumi/core/configs/params/training_params.py
@@ -777,10 +777,6 @@ class TrainingParams(BaseParams):
             save_strategy=save_strategy,
             logging_first_step=self.logging_first_step,
             torch_empty_cache_steps=self.empty_device_cache_steps,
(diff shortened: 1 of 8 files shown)


src/oumi/core/configs/params/training_params.py
```diff
@@ -777,6 +777,9 @@ def to_hf(self):
             save_strategy=save_strategy,
             logging_first_step=self.logging_first_step,
             torch_empty_cache_steps=self.empty_device_cache_steps,
+            # Note that this isn't used by the HF Trainer. To specify the checkpoint
+            # to resume from, we specify the resume_from_checkpoint arg of the `train`
+            # function.
             resume_from_checkpoint=self.resume_from_checkpoint,
```
oelachqar: Should we remove it then?

oelachqar (approved):

PR #1773, opened 2025-06-17
Add permutable attributes and combination sampling for data synthesis

jgreer013:
# Description

Add logic for permuting attributes for dataset sampling

## Related issues

Fixes OPE-1292
Towards OPE-1291

## Before submitting

## Reviewers


src/oumi/core/synthesis/planner.py
```diff
@@ -0,0 +1,170 @@
...
+)
+
+
+class DatasetPlanner:
+    def plan(
+        self,
+        synthesis_params: GeneralSynthesisParams,
+        sample_count: int,
+    ) -> list[dict]:
+        """Plan the dataset.
```
ryan-arman: nit: it would be great to a have more detailed description.

src/oumi/core/synthesis/planner.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        permutable_attributes = self._plan_permutable_attributes(
+            synthesis_params.permutable_attributes,
+            synthesis_params.combination_sampling,
+            sample_count,
+        )
+
+        dataset_plan = permutable_attributes
+
+        return dataset_plan
```
taenin, line 59:
```suggestion
        return self._plan_permutable_attributes(
            synthesis_params.permutable_attributes,
            synthesis_params.combination_sampling,
            sample_count,
        )
```

taenin: Is there a reason for this method to be different from `_plan_permutable_attributes` ? It looks like this method is just a direct pass through.

src/oumi/core/synthesis/planner.py
```diff
@@ -0,0 +1,170 @@
...
+
+                    value_distribution = attribute_distributions[perm_attr.id]
+                    sample_combination[perm_attr.id] = random.choices(
+                        list(value_distribution.keys()),
+                        list(value_distribution.values()),
+                        k=1,
+                    )[0]
+
+            for key, value in sample_combination.items():
+                samples[i][key] = value
```
taenin: Why are you using `enumerate` on line 98, but accessing the sample as `samples[i]` here?

src/oumi/core/synthesis/planner.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # If cumulative probability is greater than 1, raise an error
+        if cumulative_override_probability > 1.0:
+            raise ValueError(
+                "Cumulative probability of combination sampling must "
+                "be less than or equal to 1.0."
+            )
+
+        # Generate sample count permutations of the permutable attributes
+        samples = [{} for _ in range(sample_count)]
```
taenin: I would recommend against preallocating the samples and instead append them as you create them. It's far less error prone.

tests/unit/core/synthesis/test_planner.py
```diff
@@ -0,0 +1,271 @@
...
+from oumi.core.configs.params.synthesis_params import (
+    AttributeCombination,
+    GeneralSynthesisParams,
+    PermutableAttribute,
+    PermutableAttributeValue,
+)
+from oumi.core.synthesis.planner import DatasetPlanner
+
+
+class TestDatasetPlanner:
```
taenin: I'd recommend following the test structure we use in our other files. Each test should be a top-level function instead of wrapped in a class. If you're using Cursor for this, tell it to use other test files as a reference.

src/oumi/core/synthesis/planner.py
jgreer013: This method will be expanded over time as more aspects of planning get implemented. Leaving this portion alone as permutable attributes will be only one aspect of this.

tests/unit/core/synthesis/test_planner.py
jgreer013: Done.

src/oumi/core/synthesis/planner.py
jgreer013: Done.

src/oumi/core/synthesis/planner.py
jgreer013: Good catch, swapping from enumeration to instantiating samples one at a time.

src/oumi/core/synthesis/planner.py
jgreer013: Done.

tests/unit/core/synthesis/test_planner.py
```diff
@@ -25,247 +25,236 @@
...
-        )
-        result = planner.plan(params, sample_count=0)
-        assert isinstance(result, list)
-        assert len(result) == 0
+@pytest.fixture(autouse=True)
+def setup_random_seed():
+    """Set up a fixed random seed for deterministic tests."""
+    random.seed(42)
+    yield
+    random.seed()  # Reset the seed after the test
```
taenin, line 33: A better approach is patch `random.random()` to return a certain value

taenin (approved):

tests/unit/core/synthesis/test_planner.py
jgreer013: Given that random.random() is called numerous times in the underlying code I'm actually inclined to disagree here, it would either result in irregular behavior (same number returned every time) or make the tests themselves too verbose (in needing to specify N-many unique returns per-test)

PR #1776, opened 2025-06-19
Judge API V2 | Merge Judge and Inference configs

kaisopos:
# Description

Judge API V2: Merge Judge and Inference configs

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/judge_config_v2.py
+++ b/src/oumi/core/configs/judge_config_v2.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass, field
-from enum import Enum
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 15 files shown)


src/oumi/core/configs/judge_config_v2.py
```diff
@@ -174,17 +73,25 @@ def _resolve_path(unresolved_path: str) -> Optional[str]:
...
-            f"Could not resolve JudgeConfig from path: {path}. "
-            "Please provide a valid local or GitHub repo path."
-        )
+        if resolved_path:
+            try:
+                return cls.from_yaml_and_arg_list(resolved_path, extra_args)
+            except Exception as e:
+                raise ValueError(
+                    f"Failed to parse {resolved_path} as JudgeConfig. "
+                    f"Please ensure the YAML file contains both 'prompt_config' and "
```
wizeng23: Is it supposed to be judge_params instead of prompt_config?

src/oumi/judges_v2/simple_judge.py
```diff
@@ -76,37 +77,37 @@ class SimpleJudge(BaseJudge):
     def __init__(
         self,
         judge_config: Union[JudgeConfig, str],
```
wizeng23, line 79: Is there a reason why you use this Union type everywhere? This isn't a pattern we have for any other CLI verb. Could you not just instantiate a judge_config from the path in the very beginning of the code, and pass around a JudgeConfig within the code?

wizeng23 (approved):

wizeng23:
Overall LGTM for readability. Please wait for another approval from someone with context.

src/oumi/core/configs/judge_config_v2.py
kaisopos: Yes, that might be a better name - see [here](https://oumi-ai.slack.com/archives/C081RNZUBCZ/p1750261702959839?thread_ts=1750243519.247059&cid=C081RNZUBCZ)

src/oumi/judges_v2/simple_judge.py
kaisopos: IMO, CLI should not be the only entry point to the judge code. For some cases, instantiating a SimpleJudge might be preferable and more flexible. Our 2 V1 notebooks as well showcase this: [1](https://github.com/oumi-ai/oumi/blob/main/notebooks/Oumi%20-%20Custom%20Judge.ipynb), [2](https://github.com/oumi-ai/oumi/blob/main/notebooks/Oumi%20-%20Oumi%20Judge.ipynb). 

So, when instantiating a SimpleJudge, I still want to be able to point to a default ("built-in") judge, rather than always having to construct a config object

src/oumi/core/configs/judge_config_v2.py
kaisopos: Yes - good catch! fixed

oelachqar (approved):

PR #1780, opened 2025-06-23
Add adaptive semaphore to enable future adaptive throughput scenarios

jgreer013:
# Description

Bounded Semaphores are a lock mechanism allowing a finite amount of "permits" to allow so many "waiters" to access a given resource.

Traditionally, bounded semaphores are a fixed size - once initialized, they keep that capacity.

This PR introduced an Adaptive Semaphore - one which can change its size allocation. For size decreases, nothing special really happens - the existing permits stay valid, but no new permits are granted.

For size increases, the semaphore will immediately grant <delta> waiters access to the resource.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/adaptive_semaphore.py
+++ b/src/oumi/inference/adaptive_semaphore.py
@@ -15,8 +15,10 @@
 import asyncio
 from collections import deque
 
+from oumi.core.async_utils import safe_asyncio_run
 
-class AdaptiveSemaphore:
+
+class AdaptiveSemaphore(asyncio.Semaphore):
     """A semaphore that can dynamically adjust capacity while preserving waiters."""
 
     def __init__(self, initial_capacity: int):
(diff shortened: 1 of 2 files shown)


src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self, initial_capacity: int):
+        """Initialize the adaptive semaphore.
+
+        Args:
+            initial_capacity: The initial capacity of the semaphore.
+        """
+        self._capacity = initial_capacity
+        self._current_count = initial_capacity
+        self._waiters: deque = deque()
+        self._condition = asyncio.Condition()
```
taenin: Just curious: why are you using a condition instead of a Lock here? I'm assuming you're using this condition to make all access to `_capacity` atomic.

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from collections import deque
+
+
+class AdaptiveSemaphore:
```
taenin, line 21: Should we add a context manager mixin similar to what asyncio does for the base Semaphore class? https://github.com/python/cpython/blob/main/Lib/asyncio/locks.py#L352

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+                waiter = asyncio.get_event_loop().create_future()
+                self._waiters.append(waiter)
+            else:
+                self._current_count -= 1
+
+        if waiter is None:
+            return
+
+        try:
+            await waiter
```
taenin: Should we also call `self._waiters.remove` in a `finally` statement here?  This is what's done in the base semaphore implementation

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+            async with self._condition:
+                self._current_count += 1
+
+            # Wake up the next waiter if any
+            while self._waiters and self._current_count > 0:
+                waiter = self._waiters.popleft()
+                if not waiter.cancelled():
+                    self._current_count -= 1
+                    waiter.set_result(None)
+                    break
```
taenin: Is there a reason we're locking during the initial `self._current_count+=1` but not during the subsequent `self._current_count -=1` in the `while` block?

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # Schedule the release to run in the event loop
+        asyncio.create_task(_release())
```
taenin: I'm curious: why do we need to do this in the event loop? Is this because we need to use `async with` on our condition?

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,97 @@
...
+
+    async def adjust_capacity(self, new_capacity: int):
```
taenin, line 107: Is there a bug here if I call `adjust_capacity(-10)` followed by `adjust_capacity(-5)`? I think the code as written will cause 5 waiters to wake up even though the capacity is still below 0.

src/oumi/inference/adaptive_semaphore.py
jgreer013: Looks like the better call is simply to subclass the Semaphore class itself since those mixins aren't publically exposed.

src/oumi/inference/adaptive_semaphore.py
jgreer013: Good point, adding input validation

src/oumi/inference/adaptive_semaphore.py
jgreer013: I think this was an artifact of a previous implementation which used notify unnecessarily. Updating to use a lock instead.

src/oumi/inference/adaptive_semaphore.py
jgreer013: Good point, added

src/oumi/inference/adaptive_semaphore.py
taenin: You could probably write your own mixin for the context manager without subclassing the base sempahore class. That mixin just allows people to use it like `with AdaptiveSemaphore: ...`

src/oumi/inference/adaptive_semaphore.py
jgreer013: Artifact of previous work which was resolved elsewhere but never changed back

src/oumi/inference/adaptive_semaphore.py
jgreer013: Done.

src/oumi/inference/adaptive_semaphore.py
jgreer013: Switched to using our `safe_asyncio_run` method, but yes release needs to be async because it uses asyncio's lock.

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -0,0 +1,134 @@
...
+            try:
+                self._waiters.remove(waiter)
+            except ValueError:
+                pass
+
+    async def _release_async(self):
+        """Release a permit."""
+        async with self._lock:
+            self._current_capacity += 1
+            self._current_capacity = min(self._current_capacity, self._max_capacity)
```
taenin: Nit:
```suggestion
            self._current_capacity = min(self._current_capacity+1, self._max_capacity)
```

taenin (approved):

PR #1784, opened 2025-06-26
Add adaptive throughput controller in preparation for adaptive inference

jgreer013:
# Description

Now that we have an AdaptiveSemaphore, we need a way of keeping track of successful and failed requests, and use this information to influence our inference concurrency.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/remote_params.py
+++ b/src/oumi/core/configs/params/remote_params.py
@@ -81,3 +81,54 @@ class RemoteParams(BaseParams):
             raise ValueError(
                 "Retry backoff max must be greater than or equal to retry backoff base."
             )
+
+
+@dataclass
+class AdaptiveThroughputParams(BaseParams):
+    """Configuration for adaptive throughput control."""
+
+    initial_concurrency: int = 5
+    """Initial number of concurrent requests to start with."""
+
+    max_concurrency: int = 100
+    """Maximum number of concurrent requests allowed."""
+
+    concurrency_step: int = 5
+    """How much to increase concurrency during warmup."""
+
+    update_interval: float = 10.0
+    """Seconds between attempted updates."""
+
+    error_threshold: float = 0.01
+    """Error rate threshold (0.01 = 1%) to trigger backoff."""
+
+    backoff_factor: float = 0.8
+    """Factor to multiply concurrency by during backoff.
+    0.8 = 80% of current concurrency)."""
+
+    recovery_threshold: float = 0.00
+    """Error rate threshold (0.00 = 0%) to allow recovery."""
+
+    window_size: int = 50
+    """Number of recent requests to consider for error rate calculation."""
+
+    def __post_init__(self):
+        """Validate the adaptive throughput parameters."""
+        if self.initial_concurrency < 1:
+            raise ValueError("Initial concurrency must be greater than or equal to 1.")
+        if self.max_concurrency < self.initial_concurrency:
+            raise ValueError(
+                "Max concurrency must be greater than or equal to initial concurrency."
+            )
+        if self.concurrency_step < 1:
+            raise ValueError("Concurrency step must be greater than or equal to 1.")
+        if self.update_interval <= 0:
+            raise ValueError("Update interval must be greater than 0.")
+        if self.error_threshold < 0 or self.error_threshold > 1:
+            raise ValueError("Error threshold must be between 0 and 1.")
+        if self.backoff_factor <= 0:
+            raise ValueError("Backoff factor must be greater than 0.")
+        if self.recovery_threshold < 0 or self.recovery_threshold > 1:
+            raise ValueError("Recovery threshold must be between 0 and 1.")
+        if self.window_size < 1:
+            raise ValueError("Window size must be greater than or equal to 1.")
--- a/src/oumi/core/configs/params/remote_params.py
+++ b/src/oumi/core/configs/params/remote_params.py
@@ -87,38 +87,55 @@ class RemoteParams(BaseParams):
 class AdaptiveThroughputParams(BaseParams):
     """Configuration for adaptive throughput control."""
 
-    initial_concurrency: int = 5
-    """Initial number of concurrent requests to start with."""
+    min_concurrency: int = 5
+    """Minimum number of concurrent requests to allow."""
 
     max_concurrency: int = 100
     """Maximum number of concurrent requests allowed."""
 
     concurrency_step: int = 5
-    """How much to increase concurrency during warmup."""
+    """How much to increase concurrency during warmup.
 
-    update_interval: float = 10.0
+    During warmup, concurrency will be increased by this amount. (i.e. if concurrency is
+    100, and the concurrency step is 5, the concurrency will be increased to 105).
+    """
+
+    update_interval: float = 60.0
     """Seconds between attempted updates."""
 
     error_threshold: float = 0.01
-    """Error rate threshold (0.01 = 1%) to trigger backoff."""
+    """Error rate threshold (0.01 = 1%) to trigger backoff.
+
+    If the error rate is greater than this threshold, the concurrency will be reduced.
+    """
 
     backoff_factor: float = 0.8
     """Factor to multiply concurrency by during backoff.
-    0.8 = 80% of current concurrency)."""
+
+    During backoff, the concurrency will be reduced by this factor. (i.e. if concurrency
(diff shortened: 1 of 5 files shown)


src/oumi/core/configs/params/remote_params.py
```diff
@@ -81,3 +81,86 @@ def __post_init__(self):
...
+
+
+@dataclass
+class AdaptiveConcurrencyParams(BaseParams):
+    """Configuration for adaptive concurrency control."""
+
+    min_concurrency: int = 5
+    """Minimum number of concurrent requests to allow.
+
+    The concurrency will never be allowed to go below this value.
```
taenin: Nit: it may be worth mentioning why concurrency would be reduced (throttling / backoff)

For example:

```
Backoff throttling will never reduce the concurrency below this value.
```

src/oumi/core/configs/params/remote_params.py
```diff
@@ -81,3 +81,86 @@ def __post_init__(self):
...
+    """Maximum number of concurrent requests allowed.
+
+    The concurrency will never be allowed to go above this value.
+    """
+
+    concurrency_step: int = 5
+    """How much to increase concurrency during warmup.
+
+    During warmup, concurrency will be increased by this amount. (i.e. if concurrency is
+    50, and the concurrency step is 5, the concurrency will be increased to 55).
```
taenin: Should we mention that this happens every `min_update_time` seconds?

src/oumi/core/configs/params/remote_params.py
```diff
@@ -81,3 +81,86 @@ def __post_init__(self):
...
+    """
+
+    error_threshold: float = 0.01
+    """Error rate threshold (0.01 = 1%) to trigger backoff.
+
+    If the error rate is greater than this threshold, backoff will be triggered.
+
+    Consider keeping this value low, as once we hit a particular error rate, there are
+    already other requests in-flight which will likely fail, so the sooner we can
+    reduce concurrency, the better chance we have of recovering.
```
taenin: Super nit: I'd avoid using first person in public doc strings since these become documentation

```suggestion
    Consider keeping this value low. Once hit a particular error rate is reached, there are
    already other requests in-flight which will likely fail, so the sooner
    concurrency is reduced, the better chance the system has have of recovering.
```

src/oumi/core/configs/params/remote_params.py
```diff
@@ -81,3 +81,86 @@ def __post_init__(self):
...
+    recovery_threshold: float = 0.00
+    """Error rate threshold (0.00 = 0%) to allow recovery.
+
+    If the error rate is less than this threshold, recovery will be triggered.
+    """
+
+    min_window_size: int = 10
+    """Minimum number of recent requests to consider for error rate calculation.
+
+    If the number of recent requests is less than this threshold, the concurrency will
```
taenin: Are "recent requests" considered to be requests in the last `min_update_time` seconds?

src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -0,0 +1,218 @@
...
+
+class AdaptiveConcurrencyController:
+    """Controls concurrency based on error rates.
+
+    The controller functions as a semaphore, but can dynamically adjust the number of
+    concurrent requests based on error rates by recording the outcomes of requests.
+
+    The controller will not adjust the concurrency unless the number of recent requests
+    is greater than or equal to the min window size, and the time since the last
+    adjustment is greater than or equal to the min update time.
```
taenin: I'd recommend using variable names in the description where possible (so "min_window_size" instead of "min window size", etc) so it's easier to search the our docs for where they're used.

src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def release(self):
+        """Try to adjust concurrency before releasing a permit."""
+        safe_asyncio_run(self._try_adjust_concurrency())
+        self._semaphore.release()
+
+    async def __aenter__(self):
+        """Enter the context manager."""
+        await self.acquire()
+        return None
```
taenin: I think you want to return `self` here, not None

src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -0,0 +1,218 @@
...
+            self._outcomes.append(False)
+
+    async def acquire(self):
+        """Acquire a permit."""
+        await self._semaphore.acquire()
+
+    def release(self):
+        """Try to adjust concurrency before releasing a permit."""
+        safe_asyncio_run(self._try_adjust_concurrency())
+        self._semaphore.release()
```
taenin: Any reason this isn't an async method? Using `safe_asyncio_run` will run the code in a new thread. If you make this method async instead you can let the caller choose if they need to do that.

src/oumi/inference/adaptive_concurrency_controller.py
taenin: This would let you write something like:
```
async with AdaptiveConcurrencyController(config) as controller:
   ...
```

src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Args:
+            config: Configuration for adaptive concurrency control.
+        """
+        self._config = config
+        self._current_concurrency = config.min_concurrency
+        self._semaphore = AdaptiveSemaphore(self._current_concurrency)
+
+        # Track request outcomes in a sliding window
+        self._outcomes = deque(maxlen=_MAX_OUTCOMES_WINDOW_SIZE)
```
taenin: Could you add a type annotation here? The term "outcomes" doesn't tell me much about the values held in it.

tests/unit/inference/test_adaptive_concurrency_controller.py
```diff
@@ -0,0 +1,941 @@
...
+from oumi.inference.adaptive_concurrency_controller import AdaptiveConcurrencyController
+
+
+def create_config(**kwargs):
+    """Create a test configuration with default values."""
+    defaults = {
+        "min_concurrency": 5,
+        "max_concurrency": 20,
+        "concurrency_step": 2,
+        "min_update_time": 1.0,  # Short interval for faster tests
```
taenin, line 39: I strongly recommend mocking out `time` so your tests can run super quickly and you don't have to worry about tests failing randomly: https://github.com/oumi-ai/oumi/blob/d072c6721e60f0950a5547e41060777a7fd4ee1a/tests/unit/launcher/clients/test_local_client.py#L19

src/oumi/core/configs/params/remote_params.py
jgreer013: Done.

src/oumi/core/configs/params/remote_params.py
jgreer013: Done.

src/oumi/core/configs/params/remote_params.py
jgreer013: Done.

src/oumi/core/configs/params/remote_params.py
jgreer013: Done.

src/oumi/inference/adaptive_concurrency_controller.py
jgreer013: Done.

src/oumi/inference/adaptive_concurrency_controller.py
jgreer013: Done.

src/oumi/inference/adaptive_concurrency_controller.py
jgreer013: Agreement: release being synchronous makes sense (pattern for semaphores already, and you shouldn't need to wait to release a lock)

Moving concurrency adjustment to acquire but after semaphore acquire.

src/oumi/inference/adaptive_concurrency_controller.py
jgreer013: Done.

tests/unit/inference/test_adaptive_concurrency_controller.py
jgreer013: Done.

PR #1790, opened 2025-07-01
Add support for ingesting datasets for synthesis

jgreer013:
# Description

Synthesis may want to utilize pre-existing datasets in order to use their attributes as seed data for creating new datasets (think few-shot examples, sample metadata, or simply modifying existing datasets).

This PR adds support for loading datasets in a variety of formats and integrating them into the dataset planning process.

## Related issues

Fixes OPE-1289

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/synthesis/dataset_ingestion.py
+++ b/src/oumi/core/synthesis/dataset_ingestion.py
@@ -1,0 +1,202 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+from pathlib import Path
+from typing import Optional
+
+import pandas as pd
+from datasets import load_dataset
+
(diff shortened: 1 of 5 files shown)


src/oumi/core/synthesis/dataset_ingestion.py
```diff
@@ -0,0 +1,202 @@
...
+from typing import Optional
+
+import pandas as pd
+from datasets import load_dataset
+
+from oumi.core.configs.params.synthesis_params import DatasetSource
+
+
+class StorageType(Enum):
+    """Storage type."""
```
taenin: Nit: we can make the docstring a bit more helpful than just the class name :P

src/oumi/core/synthesis/dataset_ingestion.py
```diff
@@ -0,0 +1,202 @@
...
+        revision: Optional[str] = None,
+    ) -> list[dict]:
+        """Read the dataset from HuggingFace."""
+        dataset = load_dataset(hf_path, split=split, revision=revision)
+        samples = []
+        for sample in dataset:
+            if not isinstance(sample, dict):
+                raise ValueError(f"Expected dict, got {type(sample)}")
+            samples.append(sample)
+        return samples
```
taenin, line 158: Any concern here with reading the entire dataset into memory?

taenin (approved):

src/oumi/core/synthesis/dataset_ingestion.py
jgreer013: Nonsense :P

src/oumi/core/synthesis/dataset_ingestion.py
jgreer013: The system's current design necessarily requires loading everything into memory, but outside of pretraining datasets I wouldn't expect most text-based datasets to be large enough to cause problems. That being said, it's definitely a feature gap.

PR #1791, opened 2025-07-01
Add support for adaptive inference

jgreer013:
# Description
image

Inference can now be made adaptive, such that the throughput scales up and down based on error rate, scaling up while below a 1% error rate, and scaling down when >= 1%.

## Related issues

Fixes OPE-1307

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/remote_params.py
+++ b/src/oumi/core/configs/params/remote_params.py
@@ -68,6 +68,18 @@ class RemoteParams(BaseParams):
     of the requests. As error rate increases above a threshold, the number of concurrent
     requests will decrease, and as error rate decreases below a threshold, the number of
     concurrent requests will increase.
+
+    When this is enabled, users should set `num_workers` to the requests per minute
+    (RPM/QPM) of the model/API, and the `politeness_policy` to 60s (as most APIs query
+    limits are dictated by the number of requests per minute).
+
+    The lowest concurrency can be is 1, and the highest concurrency is `num_workers`.
+    Updates to concurrency will happen no sooner than `politeness_policy` seconds after
+    the last update, and at least 10 requests must have been made since the last update.
+
+    In the event that even 1 concurrency causes the error rate to exceed the threshold,
+    it is recommended to increase the `politeness_policy` to allow more time between
+    requests.
     """
 
     def __post_init__(self):
@@ -106,6 +118,19 @@ class AdaptiveConcurrencyParams(BaseParams):
     """Maximum number of concurrent requests allowed.
 
     The concurrency will never be allowed to go above this value.
+    """
+
+    initial_concurrency_factor: float = 0.5
+    """Initial concurrency factor.
+
+    The initial concurrency will be set to (max_concurrency - min_concurrency) *
+    initial_concurrency_factor + min_concurrency.
+
+    Example:
+    - min_concurrency = 5
+    - max_concurrency = 100
+    - initial_concurrency_factor = 0.5
+    - initial_concurrency = (100 - 5) * 0.5 + 5 = 52.5 = 52
     """
 
     concurrency_step: int = 5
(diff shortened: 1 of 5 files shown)


src/oumi/core/configs/params/remote_params.py
```diff
@@ -61,6 +61,15 @@ class RemoteParams(BaseParams):
...
     """
 
+    use_adaptive_concurrency: bool = True
+    """Whether to use adaptive concurrency control.
+
+    If True, the number of concurrent requests will be adjusted based on the error rate
+    of the requests. As error rate increases above a threshold, the number of concurrent
+    requests will decrease, and as error rate decreases below a threshold, the number of
+    concurrent requests will increase.
+    """
```
taenin: If I set `use_adaptive_concurrency` do we disregard any of the other parameters here (politeness_policy, num_workers)? If so, let's document it.

src/oumi/inference/remote_inference_engine.py
```diff
@@ -237,10 +239,39 @@ def __init__(
         self._remote_params = remote_params
         self._remote_params.finalize_and_validate()
 
+        if self._remote_params.use_adaptive_concurrency:
+            max_concurrency = self._remote_params.num_workers
+            # Start with 50% of the max concurrency.
+            min_concurrency = max(5, max_concurrency // 2)
```
taenin: Why is 5 the max? Is there a reason the minimum_concurrency isn't just 1?

src/oumi/inference/remote_inference_engine.py
```diff
@@ -237,10 +239,39 @@ def __init__(
...
+            # Ensure min concurrency is not greater than max concurrency.
+            min_concurrency = min(min_concurrency, max_concurrency)
```
taenin: This case won't be necessary if you change minimum_concurrency above

src/oumi/inference/remote_inference_engine.py
```diff
@@ -237,10 +239,39 @@ def __init__(
...
+            # Step size is 1/4 of the range between min and max concurrency.
+            concurrency_step = max(1, (max_concurrency - min_concurrency) // 4)
+            # Min update time is 1 second less than the politeness policy.
+            min_update_time = max(1, self._remote_params.politeness_policy - 1)
+            self._adaptive_concurrency_controller = AdaptiveConcurrencyController(
+                AdaptiveConcurrencyParams(
```
taenin, line 253: Perhaps a meta comment here: are we missing a "starting concurrency" field? Should the minimum concurrency always be the starting_concurrency?

src/oumi/inference/remote_inference_engine.py
```diff
@@ -237,10 +239,39 @@ def __init__(
...
         self._remote_params.finalize_and_validate()
 
+        if self._remote_params.use_adaptive_concurrency:
+            max_concurrency = self._remote_params.num_workers
+            # Start with 50% of the max concurrency.
+            min_concurrency = max(5, max_concurrency // 2)
+            # Ensure min concurrency is not greater than max concurrency.
+            min_concurrency = min(min_concurrency, max_concurrency)
+            # Step size is 1/4 of the range between min and max concurrency.
+            concurrency_step = max(1, (max_concurrency - min_concurrency) // 4)
```
taenin: Should this be more gradual / non-linear? I know TCP typically used a slow start that had exponential growth (x2) and exponential backoff( x0.5): https://en.wikipedia.org/wiki/TCP_congestion_control#Slow_start

If we want to keep things linear, it may be a good idea to have more than just 4 buckets of possible concurrency.

src/oumi/inference/remote_inference_engine.py
jgreer013: Added

src/oumi/inference/remote_inference_engine.py
jgreer013: Switched.

src/oumi/inference/remote_inference_engine.py
jgreer013: Done.

src/oumi/inference/remote_inference_engine.py
jgreer013: After online discussion, agree that we should keep things linear for more "careful" warmup while keeping exponential backoff.

src/oumi/core/configs/params/remote_params.py
jgreer013: Added additional documentation.

taenin (approved):

PR #1792, opened 2025-07-01
Fixed various tutorial notebooks

shanghongsim:
# Description

1. vLLM inference notebook: adjust the VLLMInferenceEngine such that it can load on 4 A100-40GB GPUs
2. Evaluation with Oumi notebook: Load models in torch_dtype_str="bfloat16", and also removes import alpaca.evaluate from alpaca_evals.py

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -46,17 +46,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
taenin: Are the changes to this file intended? I know @kaisopos prefers to keep cell execution + output consistent to show users what they may expect.

shanghongsim: Yup line 233 has a code change that is intentional. I tried to change the file as minimally as possible but unfortunately I can't get the outputs to be exactly the same

notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -234,10 +226,15 @@
     "alpaca_lcwr = []\n",
     "\n",
     "for model_name in MODEL_NAMES:\n",
+    "    print(f\"######################### Evaluating {model_name}...\")\n",
     "    # Define the evaluation configuration.\n",
     "    evaluation_config = EvaluationConfig(\n",
     "        tasks=TASKS,\n",
-    "        model=ModelParams(model_name=model_name, model_max_length=MODEL_MAX_TOKENS),\n",
+    "        model=ModelParams(\n",
```
shanghongsim, line 246: @taenin here is the code change

notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -318,6 +315,13 @@
     "plt.legend()\n",
     "plt.show()"
    ]
+  },
+  {
```
kaisopos: Remove that empty box

notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -264,15 +261,15 @@
...
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "[0.19772273936170212, 0.3041057180851064, 0.38040226063829785]\n",
-      "[0.0864114369219142, 0.1973881863510852, 0.24729342554633926]\n"
+      "[0.0, 0.3, 0.3]\n",
+      "[0.00019544531392424612, 0.000449928098891749, 0.40702139031055007]\n"
```
kaisopos: Likely the reason you are getting random results (the graph does not make sense any more) is because you did not run it with the entire dataset. 

1. Please set these as follows and re-run the notebook (re-generate the graph):
```
NUM_ALPACA_EXAMPLES = None
NUM_MMLU_EXAMPLES = None
```

2. Then, without running it again, save the notebook with the following values:
```
NUM_ALPACA_EXAMPLES = 3  # Replace with None for full dataset evaluation.
NUM_MMLU_EXAMPLES = 10  # Replace with None for full dataset evaluation
```

Why?
The results of the notebook should be with the entire dataset, so that they are meaningful and informative. But, when someone wants to run it, just to test the code, it should not charge them $5. The user should intentionally change these variables and understand the code to re-produce our numbers

src/oumi/core/evaluation/backends/alpaca_eval.py
```diff
@@ -117,7 +116,8 @@ def evaluate(
     # Run AlpacaEval evaluation, i.e. annotate the model's responses.
     logger.info("Running AlpacaEval annotation.")
     logger.info(f"\tAlpacaEval `task_params`:\n{pformat(task_params)}")
-    df_leaderboard, _ = alpaca_eval.evaluate(
+    # Editted for type safety
```
kaisopos: drop the comment "Editted for type safety"
This wont make sense for a user reading this for the first time (they wont know the history; ie that this used to be implemented differently etc)

notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -217,7 +209,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": null,
```
kaisopos: It feels weird if a cell has `null` execution_count (this was not executed?) or if the numbers are out-of-order.
I would advise to not update the `execution_count` numbers, unless there is a reason

notebooks/Oumi - Evaluation with Oumi.ipynb
oelachqar: @kaisopos this might be confusing for users (running the code as-is should show the same results as the cell)

kaisopos: I think there a few notebooks that have this behavior. The default run is only for a few examples and produces meaningless results, but the user is prompted (in a comment) to set NUM_EXAMPLES=None to get the proper results. The motivation is that we don't want to charge the user a large amount of money (if they execute the notebook without inspecting what it does). At the same time, we want to display some useful results (not the meaningless results corresponding to a few examples) and make the notebook seem informative. @wizeng23 and I discussed this before in length and settled for this approach. No strong opinion, we can change it, but we should maybe do this for the other notebooks too

kaisopos: Let's maybe set
```
NUM_ALPACA_EXAMPLES = None
NUM_MMLU_EXAMPLES = None
```
for now to repro the results?

@wizeng23 FYI

shanghongsim:
I have reran notebooks to ensure that results are consistent and small details like execution numbers are in order.

notebooks/Oumi - Evaluation with Oumi.ipynb
```diff
@@ -119,6 +119,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# # The results shown below was generated from evaluating on the full dataset.\n",
```
taenin:
```suggestion
    "# # The results shown below were generated from evaluating on the full dataset.\n",
```

tests/unit/inference/test_vllm_inference_engine.py
```diff
@@ -547,6 +547,7 @@ def test_guided_decoding_json(
     mock_vllm.LLM.return_value = mock_vllm_instance
     engine = VLLMInferenceEngine(config.model)
 
+    engine._llm.chat = MagicMock()
```
taenin: You'll want to merge main with your branch. This change is already checked in.

taenin (approved):

kaisopos (approved):

PR #1795, opened 2025-07-02
Fixed issue with final conversations not consistently being saved

jgreer013:
# Description

Conversations were not being saved in the core public `infer()` method, resulting in some scenarios where only some of the conversations were being written in the case where inference was stopped and restarted, rather than ALL conversations.

## Related issues

Towards OPE-1307

## Before submitting

## Reviewers


jgreer013:
> Do you need to add / update any tests for the non-remote engines?

The core issue was with the restart logic which is only held within the base_inference_engine, the new test changes should properly validate that final output files are written and contain all examples, not just those we ran inference on in the latest try.

PR #1796, opened 2025-07-02
Add code to handle document ingestion

jgreer013:
# Description

Synthesis needs to be able to ingest a variety of document types. This PR adds the code necessary to support this, but does not integrate it into the synthesis workflow yet.

Document chunking is an additional step that has yet to be performed.

Test file is: https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf

## Related issues

Towards OPE-1297

## Before submitting

## Reviewers


src/oumi/core/synthesis/document_ingestion.py
```diff
@@ -0,0 +1,117 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+from pathlib import Path
+
+import pymupdf4llm
```
oelachqar: Maybe we can make it an optional dependency, under `oumi[synthesis]` ?

jgreer013: Do you have an example of where I can write a warning message suggesting that the user install oumi[synthesis] when they try to run the `oumi synth` command (this command doesn't exist yet but I want to be familiar with how its implemented)

oelachqar: Yes here is an example: https://github.com/oumi-ai/oumi/blob/22f688c67218680e2c8ed5fbdc2718715897778b/src/oumi/core/evaluation/backends/alpaca_eval.py#L60

jgreer013: Thank you!

oelachqar: @jgreer013 the current implementation will break oumi for users who do not have `oumi[synthesis]` installed. Instead of importing `pymupdf4llm` we need to use a try catch block, similar to [here](https://github.com/oumi-ai/oumi/blob/66d6b8c05811385ac5f97c8c4b851177a25e81c9/src/oumi/inference/llama_cpp_inference_engine.py#L27)

jgreer013: Done!

src/oumi/core/synthesis/document_ingestion.py
```diff
@@ -0,0 +1,129 @@
...
+    """Text file"""
+
+    HTML = "html"
+    """HTML file"""
+
+    MD = "md"
+    """Markdown file"""
+
+
+class DocumentPath:
```
oelachqar: nit: these two classes (`DocumentPath` and `DocumentFormat`) feel a bit uncessary. If you're using `pathlib.Path`, you can get the same result with `path.suffix`

src/oumi/core/synthesis/document_ingestion.py
```diff
@@ -0,0 +1,129 @@
...
+                "pymupdf4llm is not installed. Please install it with "
+                "`pip install oumi[synthesis]`."
+            )
+        self._pymupdf4llm = pymupdf4llm
+
+    def read(self, document_path: DocumentPath) -> list[str]:
+        """Read the document."""
+        document_format = document_path.get_document_format()
+        path = Path(document_path.get_path_str())
+        if "*" in str(path):
```
oelachqar, line 38: Do we support `*` anywhere in the name, e.g. `folder/*/pdfs/*.pdf`, or just `folder/subfolder/*.pdf` ? Per the code it looks like only the latter.

tests/testdata/pdfs/mock.pdf
oelachqar, line 1: It's ok to remove this pdf file and the associated test

jgreer013: Personally I'd rather leave the test and file - it's the bare minimum for determining if the library update introduces a breaking change.

src/oumi/core/synthesis/document_ingestion.py
jgreer013: Done.

src/oumi/core/synthesis/document_ingestion.py
jgreer013: Updated this to properly handle anywhere in the name!

PR #1798, opened 2025-07-08
Add multi-node verl SLURM job

wizeng23:
# Description

- Add Ray runtime env config `src/oumi/core/trainers/verl_ray_runtime_env.yaml`, referencing https://github.com/volcengine/verl/blob/main/verl/trainer/runtime_env.yaml, with the addition of activating the oumi conda env
- Add `configs/examples/misc/slurm_ray_init.sh`, which sets up a Ray cluster on Slurm nodes. Reference: https://docs.ray.io/en/latest/cluster/vms/user-guides/community/slurm.html.
- Add example config `configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml` for running verl on Slurm, using an Oumi job config
- Pin verl version to 0.4.0. We can't use 0.4.1 yet since it added new config fields we don't have yet.
- Misc fixes

## Related issues

Towards OPE-1338

## Before submitting

### DIFF
--- a/configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
+++ b/configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
@@ -20,10 +20,6 @@ resources:
 
 working_dir: .
 
-file_mounts:
-  ~/.netrc: ~/.netrc # WandB credentials
-  ~/.cache/huggingface/token: ~/.cache/huggingface/token # HF credentials
-
 envs:
   RAY_ENABLE_RECORD_ACTOR_TASK_LOGGING: 1
   RAY_BACKEND_LOG_LEVEL: debug
--- a/configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
+++ b/configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
@@ -1,0 +1,68 @@
+# VERL GRPO job config for GSM8K.
+#
+# Requirements:
+#   - Run step 1 of verl quickstart: https://verl.readthedocs.io/en/latest/start/quickstart.html
+#
+# Usage:
+#   oumi launch up -c configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml --cluster $OUMI_SLURM_CONNECTIONS --user wizeng
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html
+#   - Config class: oumi.core.configs.JobConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/job_config.py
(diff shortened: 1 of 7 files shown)


src/oumi/core/trainers/verl_ray_runtime_env.yaml
```diff
@@ -0,0 +1,5 @@
+working_dir: ./
```
oelachqar: Is this file needed? How would a user change the conda environment etc ?

wizeng23: It was recommended in the verl guide. They could update the value in this file, but given the pattern we use in our job configs (copying our working directory to the slurm cluster and installing that in the oumi conda env), hardcoding it makes sense IMO

oelachqar: - How about the environment "oumi" ?
- I think this would not work with pip installed oumi right ?

configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
```diff
@@ -0,0 +1,63 @@
+# VERL GRPO job config for GSM8K.
+#
+# Requirements:
+#   - Run step 1 of verl quickstart: https://verl.readthedocs.io/en/latest/start/quickstart.html
+#
+# Usage:
+#   oumi launch up -c configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml --cluster $OUMI_SLURM_CONNECTIONS --user wizeng
```
taenin:
```suggestion
#   oumi launch up -c configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml --cluster $OUMI_SLURM_CONNECTIONS --user $USER
```

src/oumi/core/trainers/verl_ray_runtime_env.yaml
taenin: +1, how would this work for users who are pip installing oumi from pypi?

oelachqar (approved):

configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml
```diff
@@ -0,0 +1,67 @@
+# VERL GRPO job config for GSM8K.
+#
+# Requirements:
+#   - Run step 1 of verl quickstart: https://verl.readthedocs.io/en/latest/start/quickstart.html
+#
+# Usage:
+#   oumi launch up -c configs/examples/misc/grpo_verl_gsm8k_slurm_job.yaml --cluster $OUMI_SLURM_CONNECTIONS --user <slurm_user>
```
oelachqar, line 7: nit: it would be useful to add a comment for what to set `$OUMI_SLURM_CONNECTIONS`

PR #1805, opened 2025-07-10
Judge API V2 | Enable prompt variable replacement by YAML

kaisopos:
# Description

Allow placeholders / variables (format: `{variable}`) in prompts (both `prompt_template` and `system_instruction` fields) to be replaced by the `template_variables` dict that is defined in YAML.
Why we need this? See [here](https://oumi-ai.slack.com/archives/C081RNZUBCZ/p1752149810332809?thread_ts=1752148996.735389&cid=C081RNZUBCZ)

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers


src/oumi/core/configs/params/judge_params.py
```diff
@@ -127,6 +136,37 @@ class JudgeParams(BaseParams):
     def __post_init__(self):
         """Validate the parameters after initialization."""
         self._validate_params()
+        self._replace_template_variables()
+
+    def _replace_template_variables(self):
+        """Apply template variables to prompt_template and system_instruction."""
```
jgreer013: Thoughts on replacing the traditional format string {word} with something more unique like <<word>>?

Also, could we segment this logic out and add support for . characters? For example:

<<attribute_id.value.description>>

I'll ultimately need this logic for synthesis, so it'd be good if we standardize. I'm not strongly opinioned on {} vs <<>>, but I'd like . support preferably.

kaisopos: `{variable}` vs `<<variable>>`: The earlier is a more common pattern in YAML, thus I am slightly more supportive of {}, but no strong opinion here. If we decide to do this change, this will be a follow-up PR. This will impact many files, since this format is already used in dataset placeholders. 

`attribute_id.value.description`: This is definitely possible (must be a follow-up PR as well), but we need to discuss where it makes sense and where not. It does not make sense for the prompt template variables (replaced in YAML) but it does make sense for the example template variables (replaced dynamically in each example). To implement this, we need to re-define the dataset (currently a list of dicts) to a list of JSON objects. @jgreer013 - let's discuss offline, so that I can better understand your needs here.

jgreer013: Something like this for dots:

```
r"\{([a-zA-Z_][\w]*(?:\.[a-zA-Z_][\w]*)*)\}"
```

jgreer013: This would include support for numbers in variable names too:
```
r"\{((?:[a-zA-Z_](?:[\w]*[a-zA-Z][\w]*)?)(?:\.(?:[a-zA-Z_](?:[\w]*[a-zA-Z][\w]*)?))*)\}"
```

Example:
```
import re

pattern = r"\{((?:[a-zA-Z_](?:[\w]*[a-zA-Z][\w]*)?)(?:\.(?:[a-zA-Z_](?:[\w]*[a-zA-Z][\w]*)?))*)\}"

test_cases = [
    "{user1.name2}",   # ✅
    "{_name1}",        # ✅
    "{name_1}",        # ✅
    "{123}",           # ❌
    "{user.456}",      # ❌
    "{user..name}",    # ❌
    "{user.}",         # ❌
    "{.name}",         # ❌
    "{name.with_underscores_1}",  # ✅
]

for case in test_cases:
    match = re.match(pattern, case)
    print(f"{case}: {'✅ Match' if match else '❌ No match'}")
```

kaisopos: Updated to use your suggested pattern + scheduled a quick sync to make sure this will address your use case.

src/oumi/core/configs/params/judge_params.py
```diff
@@ -12,13 +12,31 @@
...
+    """Resolve placeholder {variables} in the provided text from the values_dict."""
+
+    class SafeDict(dict):
+        def __missing__(self, key):
+            if missing_values_allowed:
+                return "{" + key + "}"
+            else:
+                raise ValueError(f"Missing value for placeholder: {key}")
+
+    return string.Formatter().vformat(text, (), SafeDict(values_dict))
```
oelachqar: nit: could you use named arguments for vformat?

src/oumi/core/configs/params/judge_params.py
```diff
@@ -12,13 +12,31 @@
...
 
 
+def resolve_placeholders(
+    text: str,
+    values_dict: dict[str, str],
+    missing_values_allowed: bool = False,
+) -> str:
+    """Resolve placeholder {variables} in the provided text from the values_dict."""
+
+    class SafeDict(dict):
```
oelachqar: Can we define SafeDict outside of the scope of the function?

taenin: +1. This class + method should probably be in a separate file.

src/oumi/core/configs/params/judge_params.py
taenin: I don't think the string library is needed here. Instead do the following:
```suggestion
    return text.format_map(SafeDict(values_dict))
```

src/oumi/core/configs/params/judge_params.py
```diff
@@ -85,6 +103,14 @@ class JudgeParams(BaseParams):
     system_instruction: Optional[str] = field(default=None)
     """Optional system message to guide judge behavior."""
 
+    template_variables: dict[str, str] = field(default_factory=dict)
+    """Variables to be replaced in `prompt_template` and `system_instruction`.
+
+    This dictionary contains variable names and their corresponding values that should
+    be replaced in the `prompt_template` and `system_instruction` fields, before the
+    dataset-based placeholders are processed. These variables have the following format:
+    {variable_name}."""
```
taenin, line 95: Also note that users can delimit curly brackets by doubling them up: `{{I'm not a variable}}`

src/oumi/core/configs/params/judge_params.py
kaisopos: Done

src/oumi/core/configs/params/judge_params.py
kaisopos: Done

PR #1807, opened 2025-07-11
Add parameter logging to oumi trainer

oelachqar:
# Description

- Update oumi trainer to log paramters to supported platforms
  - mlflow: only supports flat dicts, so we flatten the full config
  - wandb: supports nested dicts, but using the same flattened dict as mlflow for consistency
  - tensorboard: does not support param logging, so logging it as text
- Tested with training/eval/infer configs on all 3 platforms
 

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/trainers/oumi_trainer.py
+++ b/src/oumi/core/trainers/oumi_trainer.py
@@ -713,7 +713,9 @@ class Trainer(BaseTrainer):
                 # Log config as a formatted text to tensorboard
                 config_text = "\n".join([f"{k}: {v}" for k, v in config_dict.items()])
                 self.tensorboard_writer.add_text(
-                    "config/training_config", config_text, 0
+                    tag="config/training_config",
+                    text_string=config_text,
+                    global_step=self.state.global_step,
                 )
             except Exception as e:
                 self.log(f"Failed to log config to TensorBoard: {e}")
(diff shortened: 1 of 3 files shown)


src/oumi/core/trainers/oumi_trainer.py
```diff
@@ -679,6 +683,41 @@ def _init_logging(
...
+            except Exception as e:
+                self.log(f"Failed to log config to wandb: {e}")
+
+        # Log to TensorBoard
+        if self.params.enable_tensorboard and self.tensorboard_writer:
+            try:
+                # Log config as a formatted text to tensorboard
+                config_text = "\n".join([f"{k}: {v}" for k, v in config_dict.items()])
+                self.tensorboard_writer.add_text(
+                    "config/training_config", config_text, 0
```
taenin: What's the `0` here?

taenin (approved):

src/oumi/core/trainers/oumi_trainer.py
oelachqar: It's logged at the beginning of training, I.e step 0. Will update to clarify

src/oumi/utils/serialization_utils.py
```diff
@@ -63,6 +63,65 @@ def convert_all_keys_to_serializable_types(dictionary: dict) -> None:
             convert_all_keys_to_serializable_types(value)
 
 
+def flatten_config(
+    config: Any, prefix: str = "", separator: str = "."
+) -> dict[str, Any]:
+    """Flattens a nested config object into a flat dictionary with dot notation keys.
```
wizeng23, line 69: nit: consider giving a small example

oelachqar: updated

PR #1808, opened 2025-07-11
[GRPO] Add multi-node Oumi-verl GSM8K Slurm example

wizeng23:
# Description

- Add GSM8K dataset class ([reference](https://github.com/volcengine/verl/blob/main/examples/data_preprocess/gsm8k.py))
- Add GSM8K reward function ([reference](https://github.com/volcengine/verl/blob/main/verl/utils/reward_score/gsm8k.py))
- Add GSM8K train config, modified to run GRPO instead of PPO ([reference](https://verl.readthedocs.io/en/latest/start/quickstart.html))
- Updated Slurm job to call Oumi instead of verl for training
- Add sleep instruction after ray head node is initialized to give it time to set up. Ray documentation said it's not needed, but this was causing an error. 

Confirmed that the underlying verl config is functionally equivalent to the one in the verl quickstart (aside from being modified to be GRPO). Also confirmed that this job runs on Slurm.

## Related issues

Fixes OPE-1360

## Before submitting


configs/examples/misc/slurm_ray_init.sh
```diff
@@ -40,6 +40,8 @@ srun --nodes=1 --ntasks=1 -w "$head_node" \
     --num-cpus "$SLURM_CPUS_PER_TASK" --num-gpus "$SLURM_GPUS_ON_NODE" \
     --block &
 
+sleep 10
```
oelachqar, line 46: nit: curious why this is needed? A comment would be helpful

wizeng23: Mentioned it in the PR description. I'll add to the code as well

PR #1814, opened 2025-07-11
Add inference and post-processing to attribute synthesizer

jgreer013:
# Description

This PR adds logic for inference and post-processing of inference results to the attribute synthesizer.

## Related issues

Towards OPE-1293

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/infer.py
+++ b/src/oumi/infer.py
@@ -27,7 +27,7 @@ from oumi.core.types.conversation import (
 from oumi.utils.logging import logger
 
 
-def _get_engine(config: InferenceConfig) -> BaseInferenceEngine:
+def get_engine(config: InferenceConfig) -> BaseInferenceEngine:
     """Returns the inference engine based on the provided config."""
     if config.engine is None:
         logger.warning(
@@ -48,7 +48,7 @@ def infer_interactive(
 ) -> None:
     """Interactively provide the model response for a user-provided input."""
     # Create engine up front to avoid reinitializing it for each input.
-    inference_engine = _get_engine(config)
+    inference_engine = get_engine(config)
     while True:
         try:
             input_text = input("Enter your input prompt: ")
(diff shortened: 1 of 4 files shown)


src/oumi/infer.py
```diff
@@ -27,7 +27,7 @@
 from oumi.utils.logging import logger
 
 
-def _get_engine(config: InferenceConfig) -> BaseInferenceEngine:
+def get_engine(config: InferenceConfig) -> BaseInferenceEngine:
```
taenin: If you're making this public it's a good idea to add a test for it as well. Alternatively, consider using `build_inference_engine`

tests/unit/core/synthesis/test_attribute_synthesizer.py
```diff
@@ -95,6 +109,35 @@ def mock_generated_attribute():
     )
 
 
+@pytest.fixture
+def mock_generated_attribute_with_postprocessing():
```
taenin: It doesn't look like you're using a mock here. Consider renaming this method.

tests/unit/core/synthesis/test_attribute_synthesizer.py
```diff
@@ -12,20 +12,34 @@
...
+@pytest.fixture
+def mock_inference_config():
+    """Create a mock inference config."""
+    return Mock(spec=InferenceConfig)
+
+
+@pytest.fixture
+def mock_inference_engine():
+    """Create a mock inference engine."""
+    return Mock()
```
taenin: You never use this mock

jgreer013: Ah, leftover from before get_engine, removed.

tests/unit/core/synthesis/test_attribute_synthesizer.py
jgreer013: Looks like this is also unused, removing.

src/oumi/infer.py
jgreer013: Feeling a bit stupid for not realizing I could've just used that method lol (and no this wasn't an AI change)

PR #1817, opened 2025-07-12
Added more Qwen3 recipes for training, evaluation, and inference

shanghongsim:
# Description

Added more Qwen3 recipes for training, evaluation, and inference

## Related issues

Fixes OPE-1346

## Before submitting

## Reviewers

### DIFF
--- a/configs/recipes/qwen3/evaluation/0.6b_gcp_job.yaml
+++ b/configs/recipes/qwen3/evaluation/0.6b_gcp_job.yaml
@@ -1,0 +1,61 @@
+# Job config to eval Qwen3 0.6B.
+#
+# Requirements:
+#   - Set up SkyPilot GCP: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html#setup
+#   - Log into WandB (`wandb login`) or disable `enable_wandb`
+#
+# Usage:
+#   oumi launch up -c oumi://configs/recipes/qwen3/evaluation/0.6b_gcp_job.yaml --cluster qwen3-0.6b-eval
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html
+#   - Config class: oumi.core.configs.JobConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/job_config.py
+#   - Other job configs: configs/**/*job.yaml
+
+name: qwen3-0.6b-eval
+
+resources:
+  cloud: gcp
+  accelerators: "A100:4"
+  use_spot: false
+  disk_size: 400 # Disk size in GBs
+
+working_dir: .
+
+file_mounts:
+  ~/.netrc: ~/.netrc  # WandB credentials
+
+envs:
+  # NOTE: For SFT, update this to point to your model checkpoint.
+  # NOTE: For LoRA, instead update this to point to your LoRA adapter.
+  #       The base model will be inferred automatically.
+  MODEL_CHECKPOINT_DIR: Qwen/Qwen3-0.6B
+  WANDB_PROJECT: oumi-eval
+  OUMI_RUN_NAME: qwen3-0.6b.eval
+
+setup: |
+  set -e
+  pip install uv && uv pip install oumi[gpu] hf_transfer
+  # Install model from HF Hub. This tool increases download speed compared to
+  # downloading the model during eval.
+  HF_HUB_ENABLE_HF_TRANSFER=1 huggingface-cli download Qwen/Qwen3-0.6B
+
+run: |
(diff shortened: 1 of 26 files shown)


configs/recipes/qwen3/evaluation/0.6b_gcp_job.yaml
```diff
@@ -0,0 +1,61 @@
...
+  MODEL_CHECKPOINT_DIR: Qwen/Qwen3-0.6B
+  WANDB_PROJECT: oumi-eval
+  OUMI_RUN_NAME: qwen3-0.6b.eval
+
+setup: |
+  set -e
+  pip install uv && uv pip install oumi[gpu] hf_transfer
+  # Install model from HF Hub. This tool increases download speed compared to
+  # downloading the model during eval.
+  HF_HUB_ENABLE_HF_TRANSFER=1 huggingface-cli download Qwen/Qwen3-0.6B
```
wizeng23, line 42: nit: FYI, for smaller models, these lines aren't necessary. I generally only include it for models 8B+ params where the download speed difference is more noticeable

configs/recipes/qwen3/evaluation/0.6b_gcp_job.yaml
```diff
@@ -0,0 +1,61 @@
...
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html
+#   - Config class: oumi.core.configs.JobConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/job_config.py
+#   - Other job configs: configs/**/*job.yaml
+
+name: qwen3-0.6b-eval
+
+resources:
+  cloud: gcp
+  accelerators: "A100:4"
```
wizeng23, line 20: For smaller models, you don't need as many GPUs for the job. Generally for eval, anything <=8B we can do on 1 GPU. With 2 bytes per param @ bfloat16, plus a bit of extra memory needed for activations and whatnot, we likely could fit the 14B model as well.

configs/recipes/qwen3/sft/0.6b_full/gcp_job.yaml
```diff
@@ -0,0 +1,50 @@
...
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/launch/launch.html
+#   - Config class: oumi.core.configs.JobConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/job_config.py
+#   - Other job configs: configs/**/*job.yaml
+
+name: qwen3-0.6b-full
+
+resources:
+  cloud: gcp
+  accelerators: "A100:4"
```
wizeng23, line 20: Ditto. For 0.6 and 1.7B models, you can do 1 GPU.

configs/recipes/qwen3/sft/8b_full/gcp_job.yaml
```diff
@@ -0,0 +1,50 @@
...
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/job_config.py
+#   - Other job configs: configs/**/*job.yaml
+
+name: qwen3-8b-full
+
+resources:
+  cloud: gcp
+  accelerators: "A100:4"
+  use_spot: false
+  disk_size: 400 # Disk size in GBs
```
wizeng23, line 22: nit: you may want to bump this up for the 8B model. Each checkpoint is 16GB, so you could run out of space with longer training runs. For example, See: configs/recipes/llama3_1/sft/8b_full/gcp_job.yaml. 800 is probably fine if it's 1 epoch.

Note that this generally isn't an issue for LoRA and QLoRA, as the adapters are much smaller.

wizeng23 (approved):
Could you confirm in your PR descriptions what kind of testing you did for the PR? In this case, I want to know if you ran any of these jobs to sanity check them (ex. do the larger model's jobs OOM?). Note that you only need to run it for a few steps to verify correctness IMO, and not to completion.

You don't have to run every config necessarily, just a small representative subset. I generally reuse the same cluster when testing:
```
oumi launch up -c oumi://configs/recipes/qwen3/sft/14b_lora/gcp_job.yaml --cluster qwen3
# After verifying it runs a few steps, I cancel
sky cancel qwen3 -y
# Run the next job, reusing the same cluster
oumi launch up -c oumi://configs/recipes/qwen3/sft/8b_full/gcp_job.yaml --cluster qwen3
# and so on
```

Note that the oumi:// prefixes won't work because these configs aren't checked in yet. You unfortunately need to remember to remove the prefixes when running the job, both for the `oumi launch up` command and in the `oumi train` command in the run section. Given the hassle, it's all the more reason to just test a select few configs

PR #1819, opened 2025-07-12
misc. cleanup

wizeng23:
# Description

- Remove batch_size from inference configs that don't use it (anything but native)

## Before submitting

### DIFF
--- a/configs/projects/coalm/8b_infer.yaml
+++ b/configs/projects/coalm/8b_infer.yaml
@@ -6,7 +6,6 @@
 
 model:
   model_name: "uiuc-convai/CoALM-8B"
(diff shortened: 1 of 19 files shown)


configs/projects/coalm/8b_infer.yaml
```diff
@@ -6,7 +6,6 @@
 
 model:
   model_name: "uiuc-convai/CoALM-8B"
-  adapter_model: null  # Update for LoRA-tuned models.
```
oelachqar: nit: why remove? Some of the CoALM models are LoRA/Q-LoRA tuned and this is helpful pointer for users

oelachqar (approved):

PR #1824, opened 2025-07-14
[1] Add initial attribute synthesizer logic

jgreer013:
# Description

Synthesis will create "GeneratedAttributes", which are the outputs of the synthesis process produced by LLMs. This PR produces the instructions/prompts needed to produce such outputs, but does not run inference or post-processing (yet).

## Related issues

Fixes OPE-1355

## Before submitting

## Reviewers


src/oumi/core/synthesis/attribute_synthesizer.py
```diff
@@ -0,0 +1,86 @@
...
+                continue
+
+            formatted_content = self._formatter.format(
+                sample,
+                turn.content,
+                missing_values_allowed=False,
+            )
+
+            # Create new Message with formatted content
+            new_message = turn.model_copy(update={"content": formatted_content})
```
ryan-arman: is it intentional not to deep_copy and so mutate the original conversation? if so, maybe we should document this with a comment

tests/unit/core/synthesis/test_attribute_synthesizer.py
```diff
@@ -0,0 +1,320 @@
...
+        PermutableAttribute(
+            id="style",
+            attribute="Writing Style",
+            description="The style of writing to use",
+            possible_values=[
+                PermutableAttributeValue(
+                    id="formal",
+                    value="Formal",
+                    description="A formal writing style",
+                    sample_rate=0.6,
```
ryan-arman: sample_rate is optional and it seems we are not testing it. Maybe better to remove it here?

src/oumi/core/synthesis/attribute_synthesizer.py
```diff
@@ -0,0 +1,86 @@
...
+        # TODO: Run inference
+
+        # TODO: Post-process inference results
+
+        # TODO: Return inference results
+        return inference_conversations
+
+    def _format_instructions(
+        self,
+        sample: dict,
```
ryan-arman, line 60: isn't it a bit confusing to have both samples (items) and sample_rate (probability based selection)?

src/oumi/core/synthesis/attribute_synthesizer.py
jgreer013: Not intentional, updated.

tests/unit/core/synthesis/test_attribute_synthesizer.py
jgreer013: Fair point, removing

src/oumi/core/synthesis/attribute_synthesizer.py
jgreer013: Sample rate/sampling is fairly separate from the rest of the generation logic but I'd be open to revisiting this later.

ryan-arman (approved):

PR #1826, opened 2025-07-14
[4] Add support for Transform Attributes

jgreer013:
# Description

To be able to output datasets in a specific format (i.e. training, eval, etc.), we need to be able to support "Transformed" Attributes, which are effectively attributes of various types (str, list[str], dict[str, str], and Conversation) which can be "filled in" using the other attributes in the sample.

This PR adds the class necessary to handle these attributes, but does not integrate it into the synthesis flow.

## Related issues

Fixes OPE-1385

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/synthesis/attribute_transformation.py
+++ b/src/oumi/core/synthesis/attribute_transformation.py
@@ -1,0 +1,154 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import uuid
+from typing import Union
+
+from oumi.core.configs.params.synthesis_params import (
+    ChatTransform,
+    DictTransform,
+    ListTransform,
+    TransformedAttribute,
+)
+from oumi.core.types.conversation import Conversation, Message
+from oumi.utils.placeholders import resolve_placeholders
+
+SampleValue = Union[str, list[str], dict[str, str], Conversation]
+
+
+class AttributeTransformer:
+    """Transforms attributes of a dataset plan to a particular format."""
+
+    def __init__(self, transformed_attributes: list[TransformedAttribute]):
+        """Initializes the attribute transformer.
+
+        Args:
+            transformed_attributes: The attributes describing transformations to apply.
+        """
+        self._transformed_attributes = transformed_attributes
+
+    def transform(
+        self,
+        samples: list[dict[str, SampleValue]],
+    ) -> list[dict[str, SampleValue]]:
+        """Transforms attributes of a dataset plan to a particular format.
+
+        Args:
+            samples: The samples to add transformed attributes to.
+
+        Returns:
+            The samples with the transformed attributes added.
+        """
+        for attribute in self._transformed_attributes:
+            transformed_attribute_id = attribute.id
+            sample_transforms = self._apply_transform_to_samples(samples, attribute)
+            for sample, transformed_attribute_value in zip(samples, sample_transforms):
+                sample[transformed_attribute_id] = transformed_attribute_value
+
+        return samples
+
+    def _apply_transform_to_samples(
+        self,
+        samples: list[dict[str, SampleValue]],
+        attribute: TransformedAttribute,
+    ) -> list[SampleValue]:
+        """Applies a transform to a list of samples."""
+        transformed_samples = []
+        for sample in samples:
+            transformed_sample = self._transform_attribute(sample, attribute)
+            transformed_samples.append(transformed_sample)
+        return transformed_samples
+
+    def _transform_attribute(
+        self,
+        sample: dict[str, SampleValue],
+        attribute: TransformedAttribute,
+    ) -> SampleValue:
+        """Transforms an attribute of a sample to a particular format."""
+        if isinstance(attribute.transformation_strategy, str):
+            return self._transform_string(sample, attribute.transformation_strategy)
+        if isinstance(attribute.transformation_strategy, ListTransform):
+            return self._transform_list(sample, attribute.transformation_strategy)
+        elif isinstance(attribute.transformation_strategy, DictTransform):
+            return self._transform_dict(sample, attribute.transformation_strategy)
+        elif isinstance(attribute.transformation_strategy, ChatTransform):
+            return self._transform_chat(
+                sample, attribute.transformation_strategy, attribute.id
+            )
+        else:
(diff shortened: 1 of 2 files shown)


src/oumi/core/synthesis/attribute_transformation.py
```diff
@@ -0,0 +1,169 @@
...
+        self,
+        samples: list[dict[str, SampleValue]],
+        attribute: TransformedAttribute,
+    ) -> list[SampleValue]:
+        """Applies a transform to a list of samples."""
+        transformed_samples = []
+        for sample in samples:
+            transformed_sample = self._transform_attribute(sample, attribute)
+            transformed_samples.append(transformed_sample)
+        return transformed_samples
```
taenin:
```suggestion
        return [self._transform_attribute(sample, attribute) for sample in samples]
```

Alternatively, delete this method if it's no longer used after my other comments

src/oumi/core/synthesis/attribute_transformation.py
```diff
@@ -0,0 +1,169 @@
...
+            each sample as the input to the transformation.
+
+        Returns:
+            The samples with the transformed attributes added.
+        """
+        for attribute in self._transformed_attributes:
+            transformed_attribute_id = attribute.id
+            sample_transforms = self._apply_transform_to_samples(samples, attribute)
+            for sample, transformed_attribute_value in zip(samples, sample_transforms):
+                sample[transformed_attribute_id] = transformed_attribute_value
```
taenin: Given what you're doing inside of `_apply_transform_to_samples`, I think you could greatly simplify this method (remove two loops). If you go this route, you no longer need to call `_apply_transform_to_samples`

```suggestion
            for sample in samples:
                sample[transformed_attribute_id] = self._transform_attribute(sample, attribute)
```

src/oumi/core/synthesis/attribute_transformation.py
```diff
@@ -0,0 +1,169 @@
...
+    def _transform_list(
+        self,
+        sample: dict[str, SampleValue],
+        transform: ListTransform,
+    ) -> list[str]:
+        """Transforms a list attribute of a sample to a particular format."""
+        result = []
+        for element_transform in transform.element_transforms:
+            result.append(self._transform_string(sample, element_transform))
+        return result
```
taenin:
```suggestion
        return [self._transform_string(sample, e) for e in transform.element_transforms]
```

src/oumi/core/synthesis/attribute_transformation.py
```diff
@@ -0,0 +1,169 @@
...
+    def _transform_dict(
+        self,
+        sample: dict[str, SampleValue],
+        transform: DictTransform,
+    ) -> dict[str, str]:
+        """Transforms a dict attribute of a sample to a particular format."""
+        result = {}
+        for key, value_transform in transform.transforms.items():
+            result[key] = self._transform_string(sample, value_transform)
+        return result
```
taenin:
```suggestion
        return {k: self._transform_string(sample, v) for k, v in transform.transforms.items()}
```

tests/unit/core/synthesis/test_attribute_transformation.py
```diff
@@ -0,0 +1,383 @@
...
+        {"name": "Alice", "age": "25", "city": "New York"},
+        {"name": "Bob", "age": "30", "city": "San Francisco"},
+        {"name": "Charlie", "age": "35", "city": "Chicago"},
+    ]
+
+
+@pytest.fixture
+def empty_params():
+    """Empty synthesis parameters for testing."""
+    return GeneralSynthesisParams()
```
taenin: This is never used.

tests/unit/core/synthesis/test_attribute_transformation.py
```diff
@@ -0,0 +1,383 @@
...
+    """Parameters with no transformed attributes."""
+    return GeneralSynthesisParams(transformed_attributes=None)
+
+
+@pytest.fixture
+def params_with_empty_transformed_attributes():
+    """Parameters with empty transformed attributes list."""
+    # Note: GeneralSynthesisParams validates that transformed_attributes cannot be empty
+    # so we use None instead
+    return GeneralSynthesisParams(transformed_attributes=None)
```
taenin: I wouldn't create a fixture for these. It's far less code to simply write `GeneralSynthesisParams(transformed_attributes=None)` when you need it.

tests/unit/core/synthesis/test_attribute_transformation.py
```diff
@@ -0,0 +1,383 @@
...
+def test_transform_with_no_transformed_attributes(
+    basic_samples, params_with_no_transformed_attributes
+):
+    """Test transform with no transformed attributes returns original samples."""
+    transformer = AttributeTransformer(params_with_no_transformed_attributes)
+    result = transformer.transform(basic_samples)
+
+    # Should return the same samples unchanged
+    assert result == basic_samples
+    assert len(result) == len(basic_samples)
```
taenin: Doesn't the equality check above also check len()?

Ditto here and below

tests/unit/core/synthesis/test_attribute_transformation.py
```diff
@@ -0,0 +1,383 @@
...
+
+    assert len(result) == 3
+    assert result[0]["greeting"] == "Hello Alice!"
+    assert result[1]["greeting"] == "Hello Bob!"
+    assert result[2]["greeting"] == "Hello Charlie!"
+
+    # Original attributes should still be present
+    assert result[0]["name"] == "Alice"
+    assert result[1]["name"] == "Bob"
+    assert result[2]["name"] == "Charlie"
```
taenin: Could we instead test equality of the entire object? How do we know that your function is not adding new attributes? This test wouldn't fail in that case.

src/oumi/core/synthesis/attribute_transformation.py
jgreer013: Good point, done

src/oumi/core/synthesis/attribute_transformation.py
jgreer013: Deleted

src/oumi/core/synthesis/attribute_transformation.py
jgreer013: Done

src/oumi/core/synthesis/attribute_transformation.py
jgreer013: Done.

tests/unit/core/synthesis/test_attribute_transformation.py
jgreer013: Removed

tests/unit/core/synthesis/test_attribute_transformation.py
jgreer013: Damn you cursor

tests/unit/core/synthesis/test_attribute_transformation.py
jgreer013: Good point, removed

tests/unit/core/synthesis/test_attribute_transformation.py
jgreer013: Good point, updated

taenin (approved):

PR #1829, opened 2025-07-16
refactor: use Params instead of config for sampleAnalyzer  adds unit tests

ryan-arman:
# Description
This PR use Params instead of config for sampleAnalyzer and adds unit tests

## Related issues

Towards OPE-1407

## Before submitting

## Reviewers


src/oumi/core/configs/analyze_config.py
```diff
@@ -68,8 +72,8 @@ def __post_init__(self):
         # Validate analyzer configurations
         analyzer_ids = set()
         for analyzer in self.analyzers:
-            if not analyzer.id:
-                raise ValueError("Each analyzer must have a unique 'id'")
+            # Validate each analyzer using BaseParams validation
+            analyzer.finalize_and_validate()
```
jgreer013: I'm a bit confused - isn't the above method private __finalize_and_validate__()?

Also, why do we need to run this in the post_init of the config? Couldn't we just validate in the analyzer's post_init, meaning every analyzer we get is already valid?

tests/unit/core/configs/test_analyze_config.py
```diff
@@ -0,0 +1,116 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from oumi.core.configs.analyze_config import AnalyzeConfig, SampleAnalyzerParam
+
+
+class TestSampleAnalyzerParam:
```
jgreer013: Our test structure is flat, no test suites.

Cursor does this by default generally unless you tell it not to...

jgreer013 (changes requested):

src/oumi/core/configs/analyze_config.py
ryan-arman: finalize_and_validate is a public method of the BaseParams class which is called here and will call BaseParams._finalize_and_validate() which then calls analyzer.__finalize_and_validate__() which is the private method overridden by SampleAnalyzerParam.

The reason to do it this way is to fail fast and catch configuration errors immediately when the config is created, not when the analyzer is instantiated.

Does that make sense?

tests/unit/core/configs/test_analyze_config.py
ryan-arman: thanks, made the tests flat

src/oumi/core/configs/analyze_config.py
jgreer013: Right, I'm just wondering why we're doing our validation in the finalize_and_valid method, instead of the post_init method, which is a method that 1. doesn't need to be called and 2. runs as soon as the config is instantiated

ryan-arman: yes I see your point now. Made the change to check it in post_init

src/oumi/core/configs/__init__.py
```diff
@@ -80,8 +80,8 @@
 """
 
 from oumi.core.configs.analyze_config import (
-    DatasetAnalyzeConfig,
-    SampleAnalyzeConfig,
+    AnalyzeConfig,
+    SampleAnalyzerParam,
```
taenin: This should be plural, "Params". See our naming conventions here: https://github.com/oumi-ai/oumi/tree/main/src/oumi/core/configs/params

src/oumi/core/configs/analyze_config.py
```diff
@@ -18,10 +18,11 @@
 from omegaconf import MISSING
 
 from oumi.core.configs.base_config import BaseConfig
+from oumi.core.configs.params.base_params import BaseParams
 
 
 @dataclass
-class SampleAnalyzeConfig:
+class SampleAnalyzerParam(BaseParams):
     """Configuration for a single sample analyzer plugin."""
```
taenin:
```suggestion
    """Params for a single sample analyzer plugin."""
```

src/oumi/core/configs/__init__.py
ryan-arman: thanks for catching that. Made the change

taenin (approved):

PR #1830, opened 2025-07-16
[6] Add support for `oumi synth` in CLI

jgreer013:
# Description

This PR allows users to call `oumi synth` with an appropriate config to generate synthetic data.

Screenshot 2025-07-28 at 9 38 26 AM

## Related issues

Fixes OPE-1386

## Before submitting

## Reviewers

### DIFF
--- a/tests/unit/builders/test_quantizers.py
+++ b/tests/unit/builders/test_quantizers.py
@@ -1,0 +1,165 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for quantizer builders."""
+
+from unittest.mock import Mock, patch
+
+import pytest  # type: ignore
+
+from oumi.builders.quantizers import (
+    build_quantizer,
+    get_available_methods,
+    get_supported_formats,
+    list_all_methods,
+)
+from oumi.quantize.base import BaseQuantization
+
+
+class TestQuantizerBuilders:
+    """Test cases for quantizer builder functions."""
+
+    def test_build_quantizer_awq_method(self):
+        """Test building AWQ quantizer for AWQ methods."""
+        # Test various AWQ methods
+        for method in ["awq_q4_0", "awq_q4_1", "awq_q8_0", "awq_f16"]:
+            quantizer = build_quantizer(method)
+
+            # Verify correct type
+            assert quantizer.__class__.__name__ == "AwqQuantization"
+            assert isinstance(quantizer, BaseQuantization)
+
+    def test_build_quantizer_bnb_method(self):
+        """Test building BnB quantizer for BnB methods."""
+        # Test BnB methods
+        for method in ["bnb_4bit", "bnb_8bit"]:
+            quantizer = build_quantizer(method)
+
+            # Verify correct type
+            assert quantizer.__class__.__name__ == "BitsAndBytesQuantization"
+            assert isinstance(quantizer, BaseQuantization)
+
+    def test_build_quantizer_unsupported_method(self):
+        """Test building quantizer with unsupported method."""
+        with pytest.raises(
+            ValueError, match="Unsupported quantization method: invalid_method"
+        ):
+            build_quantizer("invalid_method")
+
+    @patch("oumi.quantize.awq_quantizer.AwqQuantization")
+    @patch("oumi.quantize.bnb_quantizer.BitsAndBytesQuantization")
+    def test_build_quantizer_fallback(self, mock_bnb, mock_awq):
+        """Test build_quantizer fallback when method doesn't match prefix."""
+        # Setup mocks
+        mock_awq_instance = Mock()
+        mock_awq_instance.supports_method.return_value = False
+        mock_awq.return_value = mock_awq_instance
+
+        mock_bnb_instance = Mock()
+        mock_bnb_instance.supports_method.return_value = True
+        mock_bnb.return_value = mock_bnb_instance
+
+        # Test with method that doesn't match any prefix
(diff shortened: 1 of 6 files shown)


src/oumi/cli/synth.py
```diff
@@ -0,0 +1,87 @@
...
+            table.add_row(f"... and {len(results) - 5} more samples")
+        cli_utils.CONSOLE.print(table)
+        cli_utils.CONSOLE.print(
+            f"\n[green]Successfully synthesized {len(results)} samples[/green]"
+        )
+    else:
+        cli_utils.CONSOLE.print(
+            f"[green]Successfully synthesized {len(results)} samples and saved to "
+            f"{parsed_config.output_path}[/green]"
+        )
```
taenin: Can you add a screenshot for these cases in your PR description?

taenin: I think it would also be helpful if we printed instructions for how the user can register their dataset with Oumi (if needed). Will an `oumi train` command work where they set the dataset_path to their output file?

src/oumi/core/configs/params/synthesis_params.py
```diff
@@ -13,17 +13,48 @@
...
 from oumi.core.configs.params.base_params import BaseParams
-from oumi.core.types.conversation import Conversation
+from oumi.core.types.conversation import Conversation, Message, Role
 
 _SUPPORTED_DATASET_FILE_TYPES = {".jsonl", ".json", ".csv", ".parquet", ".tsv"}
 
 
+@dataclass
+class TextMessage:
+    """Text-only message to make it usable in omegaconf."""
```
taenin, line 29: What's the issue with the current conversation class RE: omegaconf?

src/oumi/core/configs/params/synthesis_params.py
```diff
@@ -13,17 +13,48 @@
...
+from typing import Any, Optional
 
 from oumi.core.configs.params.base_params import BaseParams
-from oumi.core.types.conversation import Conversation
+from oumi.core.types.conversation import Conversation, Message, Role
 
 _SUPPORTED_DATASET_FILE_TYPES = {".jsonl", ".json", ".csv", ".parquet", ".tsv"}
 
 
+@dataclass
```
taenin, line 27: OK for now, but in the future I'd recommend breaking changes like this into a separate PR

src/oumi/cli/synth.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Annotated
+
+import typer
+from rich.table import Table
+
+import oumi.cli.cli_utils as cli_utils
+from oumi.utils.logging import logger
+
+
+def synth(
```
taenin, line 30: Add tests for your CLI changes (should simply mock out the underlying synth library)

src/oumi/core/configs/params/synthesis_params.py
jgreer013: Is this a breaking change when the feature isn't released yet though? 🤔

src/oumi/core/configs/params/synthesis_params.py
jgreer013: Message contains a ContentItem class which is a union of different types, which omegaconf isn't happy about. In addition, I ran into some issues with the fact that Message and Conversation were frozen dicts, hence ultimately I went with this change.

src/oumi/cli/synth.py
jgreer013: Done.

jgreer013: Done.

src/oumi/cli/synth.py
jgreer013: Done

src/oumi/cli/synth.py
```diff
@@ -0,0 +1,122 @@
...
+        show_lines=True,
+    )
+    columns = results[0].keys()
+    for column in columns:
+        table.add_column(column, style="green")
+    for i, result in enumerate(results[:5]):  # Show first 5 samples
+        representations = []
+        for column in columns:
+            representation = repr(result[column])
+            if len(representation) > 20:
```
oelachqar: it would be nice to avoid truncating, or truncate to something larger than 20 characters, which is too low to meaningfully inspect the sample

src/oumi/cli/synth.py
```diff
@@ -0,0 +1,122 @@
...
+
+    config = str(cli_utils.resolve_and_fetch_config(config))
+
+    with cli_utils.CONSOLE.status(
+        "[green]Loading configuration...[/green]", spinner="dots"
+    ):
+        # Delayed imports
+        from oumi import synthesize as oumi_synthesize
+        from oumi.core.configs.synthesis_config import SynthesisConfig
+        # End imports
```
oelachqar, line 58: nit: not sure that the extra comment (e.g. `# End imports`) are very useful

taenin: That's a convention I started when I added delayed imports everywhere to logically split it from the rest of the code

src/oumi/core/configs/params/synthesis_params.py
taenin: I meant "I'd recommending breaking up your changes into a separate PR" :P

src/oumi/cli/main.py
```diff
@@ -75,6 +76,10 @@ def get_app() -> typer.Typer:
         context_settings=CONTEXT_ALLOW_EXTRA_ARGS,
         help="Run inference on a model.",
     )(infer)
+    app.command(
+        context_settings=CONTEXT_ALLOW_EXTRA_ARGS,
```
taenin, line 80: Similar to how we allow users to use both `eval` and `evaluate` should we let users use `synth` and `synthesize`?

src/oumi/cli/synth.py
jgreer013: The table is largely unreadable if we don't truncate, curious to hear your thoughts on what would be the preferable experience. Please see screenshot above for example of current truncation behavior.

jgreer013: Updated to 200 characters, 1 row, and max 6 columns

src/oumi/cli/synth.py
jgreer013: Resolving as-is to maintain convention

src/oumi/cli/main.py
jgreer013: Done

taenin (approved):
Let's also add a test to tests/unit/cli/test_cli_main.py to ensure that your entrypoint is exposed properly.

Otherwise LGTM

PR #1833, opened 2025-07-17
Skip pymupdf4llm tests if dependency is missing

jgreer013:
# Description

`make setup` doesn't install synthesis currently, resulting in failed tests

## Related issues

Fixes OPE-1408

## Before submitting

## Reviewers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,6 @@ setup:
 				conda create -n $(CONDA_ENV) python=3.11 -y; \
 				$(CONDA_RUN) pip install uv; \
 				$(CONDA_RUN) uv pip install -e ".[dev]"; \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -61,6 +61,7 @@ setup:
 				conda create -n $(CONDA_ENV) python=3.11 -y; \
 				$(CONDA_RUN) pip install uv; \
 				$(CONDA_RUN) uv pip install -e ".[dev]"; \
+				$(CONDA_RUN) uv pip install -e ".[synthesis]"; \
```
oelachqar: @jgreer013 the recommendation here would be to skip the unit tests if the dependency is not present (similar to llamacpp etc)

jgreer013: Done.

oelachqar (approved):

PR #1841, opened 2025-07-21
[bug] Fix a bug where inference engines would wait for one politeness policy cycle after all requests were finished.

taenin:
# Description

This PR updates our concurrency logic to use a "Polite" semaphore. From a producer-consumer standpoint, the semaphore is producing "start times", which are consumed on subscription. The semaphore will then force the holder to wait if necessary given the politeness policy. When the subscriber releases the semaphore, their release time is appended to the queue.

## Before submitting

## Reviewers


src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -54,15 +54,18 @@ class AdaptiveConcurrencyController:
     ```
     """
 
-    def __init__(self, config: AdaptiveConcurrencyParams):
+    def __init__(self, config: AdaptiveConcurrencyParams, politeness_policy: float):
```
ryan-arman, line 57: how about adding politeness_policy  to AdaptiveConcurrencyParams to simplify the input config?

tests/unit/inference/test_adaptive_semaphore.py
```diff
@@ -13,10 +13,229 @@
...
+
+
+@pytest.mark.asyncio
+async def test_polite_adaptive_semaphore(mock_time, mock_asyncio_sleep):
+    semaphore = PoliteAdaptiveSemaphore(capacity=1, politeness_policy=10)
+    mock_time.time.return_value = 1
+    await semaphore.acquire()
+    assert len(semaphore._queue) == 0
+    semaphore.release()
+    assert semaphore._queue == deque([11])
```
ryan-arman, line 44: 11 is time at release = 10 + politeness policy = 1.0, right? maybe useful to add a comment

tests/unit/inference/test_adaptive_semaphore.py
```diff
@@ -13,10 +13,229 @@
...
+        38.0,
+    ]
+
+    async def acquire_and_release(semaphore: PoliteAdaptiveSemaphore):
+        await semaphore.acquire()
+        semaphore.release()
+
+    tasks = [acquire_and_release(semaphore) for _ in range(4)]
+
+    # Tasks 1 finishes at 10.0.
```
ryan-arman: we can add more comments to better clarify the logic. how about this?

Task 1: Acquires at first available slot and finishes at 10.0. Sets slot to next allowed start at 10.0 + 20 = 30.0.
Task 2: Acquires the other slot, finishes at 11.0. Sets slot to next allowed start at 11.0 + 20 = 31.0
Task 3: Starts at 12.0. Both queue slots are now [30.0, 31.0]. Needs to wait for min(queue) - current time = 30.0 - 12.0 = 18.0 seconds.
Task 4: Starts at 37.0. The queue slots are now [32.0, 33.0] (after Task 3’s release). 37.0 is after both of these, so no sleep needed.

src/oumi/inference/adaptive_semaphore.py
```diff
@@ -131,3 +131,64 @@ async def adjust_capacity(self, new_capacity: int):
...
+        if wait_time > 0:
+            await asyncio.sleep(wait_time)
+
+    async def _release_async(self):
+        """Releases the semaphore.
+
+        Adds the current time to the queue. So the next task will wait for the
+        politeness policy to be respected.
+        """
+        self._queue.append(time.time() + self._politeness_policy)
```
jgreer013, line 193: Q: Should we modify politeness policy to instead be based on time delta of when the semaphore was acquired?

For example:
Acquire semaphore & set timer
Send request
Receive response 5 seconds later
Release semaphore
Next request waits 60 - 5 seconds before being sent, rather than 60

jgreer013 (changes requested):

tests/unit/inference/test_adaptive_semaphore.py
```diff
@@ -13,10 +13,229 @@
...
+    mock_time, mock_asyncio_sleep
+):
+    """Test that multiple releases create a queue that subsequent acquires respect."""
+    semaphore = PoliteAdaptiveSemaphore(capacity=2, politeness_policy=20)
+
+    mock_time.time.side_effect = [
+        10.0,
+        11.0,
+        12.0,
+        36.0,
```
ryan-arman:
```suggestion
        36.0, # release time for Task 3
```

src/oumi/inference/adaptive_semaphore.py
taenin: I think this is a reasonable request, but for the sake of time I'm going to keep this as-is. Eventually I'd rather deprecate politeness policy + num workers and just set QPS--that's likely a better end-user experience for users.

src/oumi/inference/adaptive_concurrency_controller.py
taenin: Sure, since AdaptiveConcurrencyParams are internal this is reasonable. Updated

tests/unit/inference/test_adaptive_semaphore.py
taenin: Added

ryan-arman (approved):

tests/unit/inference/test_adaptive_semaphore.py
taenin: Done.

tests/unit/inference/test_adaptive_semaphore.py
taenin: Updated, though I've changed your description as it doesn't quite match the logic we follow.

src/oumi/inference/remote_inference_engine.py
```diff
@@ -592,15 +594,6 @@ async def _query_api(
                             f"unexpected error: {str(e)}"
                         ) from e
                     continue
-                finally:
-                    # If the request was successful or non-retriable, and we haven't
-                    # reached the max number of retries, sleep for politeness policy.
```
shanghongsim, line 597: Could you explain why this behavior was initially set? If its successful and we have not hit max, why do we need to sleep? If it is successful, would it not mean that the message has been processed and the API (e.g. OpenAI) can take in another request? Same question for non-retriable errors except that if its an error, wouldn't it mean that "message" is not retained in queue and therefore we are free to send another one without over loading the API? To me, it seems to make more sense to apply politeness policy after the system is overloaded and not on first request.

tests/unit/inference/test_remote_inference_engine.py
```diff
@@ -1051,10 +1060,11 @@ def response_callback(url: str, **kwargs: Any) -> CallbackResult:
             inference_config,
         )
         total_time = time.time() - start
-        assert 1.0 < total_time < 1.5
+        assert 0.5 < total_time < 1.0
```
shanghongsim, line 1063: 1. Does this mean that between the inference call for conversation 1 and conversation 2, there is a sleep of 0.5s? So if each inference call completes in 0s/0.25s, it will be 0s/0.5s (inference) + 0.5s (politeness policy) for the entire query? 
2. Can I also confirm that the expected behavior is for no politeness policy to be applied on first request (due to empty queue and behaviour expected in `test_polite_adaptive_semaphore_empty_queue_no_wait`). When we sat down to debug together previously, we noticed that politeness policy kicked in even upon first request. I think that was by design due to the `finally` block below.

src/oumi/inference/remote_inference_engine.py
shanghongsim: Overall, is this PR about changing the part that controls the politeness policy such that it is now controlled by the semaphore?

taenin: > Overall, is this PR about changing the part that controls the politeness policy such that it is now controlled by the semaphore?

That's right. The user can set a politeness policy which we interpret as "this worker will wait X seconds before sending another request". Previously the logic was enforce in such a way that the worker would make an API call then sleep for X seconds, which resulted in a bug where 1 request would call the API then sleep for the politeness_policy instead of returning immediately.

Now the semaphore handles the politeness policy. When a worker acquires the semaphore, they will wait before sending a request if necessary.

For the initial logic, it looks like the backoff logic for retries is handled before the request, hence no waiting on retryable errors.

tests/unit/inference/test_remote_inference_engine.py
taenin: 1. Correct
2. Also correct that the politeness policy should not be enforced for the first request! A single request should not have any politeness involved.

src/oumi/core/configs/params/remote_params.py
```diff
@@ -108,6 +108,9 @@ def __post_init__(self):
 class AdaptiveConcurrencyParams(BaseParams):
     """Configuration for adaptive concurrency control."""
 
+    politeness_policy: float = 60.0
```
oelachqar: nit: does `AdaptiveConcurrencyParams.politeness_policy` override `RemoteParams.politeness_policy` when adaptive concurrency is enabled? it's a little bit confusing to have both

shanghongsim (approved):

src/oumi/core/configs/params/remote_params.py
jgreer013: Users of the platform can't actually modify the concurrency params, those are only controlled by us at the moment.

That being said, I agree with Oussama, I'm not sure I would've chosen to add a politeness policy to concurrency params.

src/oumi/inference/adaptive_concurrency_controller.py
```diff
@@ -59,10 +59,13 @@ def __init__(self, config: AdaptiveConcurrencyParams):
 
         Args:
             config: Configuration for adaptive concurrency control.
+            politeness_policy: The politeness policy to use for the semaphore.
```
jgreer013: nit: "The amount of time workers will wait until sending their next request, in seconds"

tests/unit/inference/test_adaptive_semaphore.py
```diff
@@ -13,10 +13,236 @@
...
+    await semaphore.acquire()
+    assert len(semaphore._queue) == 0
+    semaphore.release()
+    # 11 = 1.0 + politeness_policy  (10.0)
+    assert semaphore._queue == deque([11])
+
+
+@pytest.mark.asyncio
+async def test_polite_adaptive_semaphore_adjust_capacity(mock_time, mock_asyncio_sleep):
+    semaphore = PoliteAdaptiveSemaphore(capacity=1, politeness_policy=10)
```
jgreer013, line 49: nit: I would add comments near asserts to help clarify expected behavior for each step in this timeline

src/oumi/core/configs/params/remote_params.py
oelachqar: What happens when I set `RemoteParams(politeness_policy=10, use_adaptive_concurrency=True)`? Is the effective policy 10, or 60 (default for `AdaptiveConcurrencyParams`?

taenin: I can move it back out of AdaptiveConcurrencyParams. As of right now, the default value in AdaptiveConcurrencyParams would never be used.

tests/unit/inference/test_adaptive_semaphore.py
taenin: Updated

PR #1847, opened 2025-07-22
[vision] Add vision dpo dataset class

oelachqar:
# Description

**Changes**
- Add a base vision DPO dataset class, similar to the language DPO dataset class
  - Note that both DPO datasets (text, and now vision-language) are still experimental, and do not use the `Conversation` format. 
- Add a first vision DPO dataset, to consume `jsonl` files.  
  - Add an example data file
- Fix a type annotation issue in the DPO text dataset  

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


src/oumi/core/datasets/vision_language_dpo_dataset.py
```diff
@@ -0,0 +1,174 @@
...
+        rejected_chat = sample[_REJECTED_KEY]
+        images = sample[_IMAGES_KEY] or []
+
+        if images is not None:
+            images = [self._resize_image(self._load_image(image)) for image in images]
+
+        # Only use the last message of the chosen and rejected chats.
+        # TODO: add support for conversation format
+        chosen_chat_response = self._extract_from_chat_format(chosen_chat)
+        rejected_chat_response = self._extract_from_chat_format(rejected_chat)
```
shanghongsim, line 140: Why are we only using the last assistant turn in the chat? Is it meant to be temporary (give the TODO written above)?

shanghongsim (approved):

src/oumi/core/datasets/vision_language_dpo_dataset.py
oelachqar: Yes it's temporary! the downstream code does not support multi-turns and we need to implement that

PR #1857, opened 2025-07-23
Deprecate Judge V1

kaisopos:
# Description

Deprecating Judge V1.
Per discussion with Oussama, we'll immediately remove it and not offer backwards compatibility here. 

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers


src/oumi/__init__.py
```diff
@@ -350,8 +263,6 @@ def quantize(config: QuantizationConfig) -> QuantizationResult:
     "evaluate",
     "infer_interactive",
     "infer",
-    "judge_conversations",
```
oelachqar, line 353: Should we provide an equivalent with the new judge?

oelachqar (approved):

src/oumi/__init__.py
kaisopos: Added as a new task (will do this next)

PR #1865, opened 2025-07-24
Judge API V2 | Add `JUDGE` alias

kaisopos:
# Description

Adding `JUDGE` alias

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers


src/oumi/cli/alias.py
```diff
@@ -155,6 +156,19 @@ class AliasType(str, Enum):
     "molmo-7b-o-grpo": {
         AliasType.TRAIN: "oumi://configs/recipes/vision/molmo/grpo/train.yaml",
     },
+    # Judge aliases (for generic judges only): <model_name>-<judge_name>
+    "gpt-4o-format-compliance": {
```
oelachqar: nit: IMO the emphasis should be on the judge not the model (similar to how the filename does not include the model)

oelachqar (approved):

src/oumi/cli/alias.py
kaisopos: So, should I name instead  `format-compliance-gpt-4o` or even just `format-compliance`?
WDYT?

kaisopos: @wizeng23 and @taenin any feedback here?
Oussama and I are leaning towards `format-compliance` (i.e. not mentioning the model that backs up our built-in judge)

kaisopos: I ll stick to this (`format-compliance`) for now, we can update in the future if needed

PR #1886, opened 2025-07-31
Penfever/deepspeed

penfever:
# Description

This PR adds:

* A config for Qwen 2.5 SFT on the LIMO dataset using FSDP

* Two additional Qwen 2.5 SFT configs demonstrating the use of YARN + ROPE Scaling for fine-tuning on long context (131k) and DeepSpeed ZERO-3

 * Updates to the pyproject.toml with pinned, compatible DeepSpeed versions

* Minor changes to training and training_params to enable passing through complex nested DeepSpeed parameters

* Custom @dict_cache utility to allow caching of nested dictionaries in kwargs, such as those used in rope_scaling

 * Support for standard DeepSpeed configurations
        Stage 0: Disabled (standard data parallelism)
        Stage 1: Optimizer state sharding
        Stage 2: Optimizer + gradient sharding
        Stage 3: Full sharding (optimizer + gradients + parameters)

* Tests

## Related issues

Towards OPE-1513

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/llama_cpp_inference_engine.py
+++ b/src/oumi/inference/llama_cpp_inference_engine.py
@@ -86,6 +86,8 @@ class LlamaCppInferenceEngine(BaseInferenceEngine):
             - n_threads: 4
             - filename: "*q8_0.gguf" (applies Q8 quantization by default)
             - flash_attn: True
+            - use_mmap: True (loads model parts as needed)
+            - use_mlock: True (locks the model pages in physical RAM)
             These defaults can be overridden by specifying them in
             `model_params.model_kwargs`.
         """
@@ -122,6 +124,9 @@ class LlamaCppInferenceEngine(BaseInferenceEngine):
             # Use Q8 quantization by default.
             "filename": "*8_0.gguf",
             "flash_attn": True,
+            # Memory safety defaults
+            "use_mmap": True,
+            "use_mlock": True,
         }
 
         model_kwargs = model_params.model_kwargs.copy()
--- a/tests/unit/inference/test_llama_cpp_inference_engine.py
+++ b/tests/unit/inference/test_llama_cpp_inference_engine.py
@@ -46,6 +46,8 @@ def test_initialization(mock_llama):
         n_gpu_layers=-1,
         n_threads=4,
         flash_attn=True,
+        use_mmap=True,
+        use_mlock=True,
     )
 
 


tests/unit/core/configs/params/test_ds_params_instantiation.py
```diff
@@ -0,0 +1,64 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Test to verify DSParams instantiation works without AttributeError."""
+
+import sys
+from pathlib import Path
+
+# Add root directory to path for imports
```
oelachqar: I believe this should not be needed ?

penfever: Yeah, we could get rid of this test -- assume that's what you meant? Or you just meant adding the path manually? If the latter, yes, we can avoid that as well

scripts/deepspeed/zero_to_fp32.py
wizeng23, line 1: Is this script copied from somewhere else? If so, could you:
1. Add a link to the source in the docstring, along with any edits you made to the original
2. Remove the Oumi copyright, and add an exception to https://github.com/oumi-ai/oumi/blob/main/tests/unit/test_apache_license_header.py

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+
+    overlap_events: bool = True
+    """Whether to overlap NVMe events."""
+
+    thread_count: int = 1
+    """Number of threads for NVMe operations."""
+
+
+@dataclass
+class DSParams(BaseParams):
```
wizeng23: Could you spell it out to DeepSpeedParams for clarity?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+
+    If provided, this will override other DeepSpeed parameters.
+    Takes precedence over auto-generated configuration.
+    """
+
+    zero_stage: ZeRORuntimeStage = ZeRORuntimeStage.ZERO_3
+    """ZeRO optimization stage.
+
+    Controls the level of memory optimization:
+    - Stage 0: Disabled (standard data parallelism)
```
wizeng23: Out of curiosity, is this a no-op? What's the difference between enable_deepspeed=False and Deepspeed 0?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    """Number of threads for NVMe operations."""
+
+
+@dataclass
+class DSParams(BaseParams):
+    """Configuration options for DeepSpeed distributed training.
+
+    Based on DeepSpeed configuration schema and best practices from
+    LLaMA-Factory implementation.
+    """
```
wizeng23: Could you add a link to the DeepSpeed config that this will be converted to for reference?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    - Stage 3: Full sharding (optimizer + gradients + parameters)
+
+    Stage 3 provides maximum memory efficiency for large models.
+    """
+
+    offload_optimizer: Optional[OffloadConfig] = None
+    """Configuration for optimizer state offloading.
+
+    When enabled, optimizer states are offloaded to CPU/NVMe
+    to save GPU memory. Available for ZeRO stages 2 and 3.
```
wizeng23:
```suggestion
    to save GPU memory. Available for ZeRO stages 1, 2, and 3.
```

https://www.deepspeed.ai/docs/config-json/#optimizer-offloading

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    """Gradient clipping value."""
+
+    # Advanced options
+    zero_allow_untested_optimizer: bool = True
+    """Allow optimizers not explicitly tested with ZeRO."""
+
+    zero_force_ds_cpu_optimizer: bool = True
+    """Force DeepSpeed CPU optimizer when CPU offloading is enabled."""
+
+    # ZeRO-2 specific communication parameters
```
wizeng23: nit: Move this group above the zero-3 group

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    """Enable allgather partitions for ZeRO-2."""
+
+    reduce_scatter: bool = True
+    """Enable reduce scatter for ZeRO-2."""
+
+    round_robin_gradients: bool = False
+    """Enable round robin gradients for ZeRO-2."""
+
+    # Activation checkpointing
+    activation_checkpointing: dict[str, Any] = field(default_factory=dict)
```
wizeng23: Could you add more details on how to populate this field, or link to documentation on how to do so?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    """Maximum reuse distance for parameters in ZeRO-3."""
+
+    stage3_gather_16bit_weights_on_model_save: bool = True
+    """Whether to gather 16-bit weights during model saving in ZeRO-3."""
+
+    sub_group_size: int = int(1e9)
+    """Sub-group size for ZeRO-3 parameter sharding."""
+
+    # Training parameters (auto-configured by HuggingFace Transformers)
+    train_batch_size: str = "auto"
```
wizeng23: What other value can this take besides "auto" (given it's a string field)? If it's not meant to be configurable by the user, could we remove these fields and just add a comment in this class that these fields are auto-set?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+        if self.offload_optimizer is not None and self.zero_stage not in [
+            ZeRORuntimeStage.ZERO_2,
+            ZeRORuntimeStage.ZERO_3,
+        ]:
+            raise ValueError(
+                "Optimizer offloading requires ZeRO stage 2 or 3. "
+                f"Current stage: {self.zero_stage}"
+            )
+
+    def to_deepspeed_config(self) -> dict[str, Any]:
```
wizeng23: nit: `to_deepspeed` to be in line with the name of our other conversion methods

src/oumi/core/configs/params/ds_params.py
wizeng23: Could you confirm whether the default value for these fields are in line with DeepSpeed's, or if you modified any of them?

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+            "zero_force_ds_cpu_optimizer": self.zero_force_ds_cpu_optimizer,
+            "steps_per_print": self.steps_per_print,
+            "wall_clock_breakdown": self.wall_clock_breakdown,
+        }
+
+        # Add mixed precision configuration
+        if self.precision == DeepSpeedPrecision.FP16:
+            config["fp16"] = {
+                "enabled": "auto",
+                "loss_scale": 0,
```
wizeng23: Will users want to configure this? If so, should we add a field for it?

src/oumi/core/configs/params/training_params.py
```diff
@@ -691,8 +691,12 @@ class to define specific profiling settings.
     not satisfactory, or for new models not yet fully-integrated by Oumi.
     """
 
-    def to_hf(self):
-        """Converts Oumi config to HuggingFace's TrainingArguments."""
+    def to_hf(self, training_config=None):
```
wizeng23:
```suggestion
    def to_hf(self, training_config: TrainingConfig = None):
```

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,388 @@
...
+    """If True, enables DeepSpeed distributed training.
+
+    Allows training larger models using ZeRO optimization stages
+    for memory-efficient distributed training.
+    """
+
+    deepspeed_config_path: Optional[Union[str, Path]] = None
+    """Path to a DeepSpeed JSON configuration file.
+
+    If provided, this will override other DeepSpeed parameters.
```
wizeng23: To confirm, if this is set, other fields in this class are ignored right? If so, could the comment here mention that? Could be just me, but "override" to me sounds like the values in this class and the deepspeed config path are getting merged somehow.

src/oumi/core/distributed.py
```diff
@@ -437,6 +445,61 @@ def prepare_model_for_distributed(
...
+    Args:
+        config: The training configuration.
+
+    Returns:
+        bool: True if DeepSpeed ZeRO-3 is enabled, False otherwise.
+    """
+    return config.deepspeed.is_zero3_enabled()
+
+
+def create_deepspeed_config_file(config: TrainingConfig, output_path: str) -> str:
```
wizeng23: This doesn't seem to be used?

src/oumi/train.py
```diff
@@ -526,6 +529,24 @@ def train(
         barrier()
 
         logger.info("Saving final model...")
+
+        # Check for potential multi-node DeepSpeed ZeRO-3 saving issue
```
wizeng23: nit: Should we move this further up in the training script? Otherwise, by the time this warning prints, it may be too late for the user to address the issue. I'm assuming this issue also affects saving checkpoints too

tests/unit/core/configs/params/test_ds_params_instantiation.py
wizeng23: I'm assuming it's the latter. What happens if we don't have this line?

tests/unit/core/configs/params/test_ds_params_instantiation.py
```diff
@@ -0,0 +1,64 @@
...
+    except AttributeError as e:
+        if "'super' object has no attribute '__post_init__'" in str(e):
+            print(f"✗ FAILED: {e}")
+            print("The AttributeError is still present!")
+            return False
+        else:
+            raise
+
+
+if __name__ == "__main__":
```
wizeng23: Do we need a main fn on this test?

tests/unit/core/configs/params/test_ds_params_instantiation.py
```diff
@@ -0,0 +1,64 @@
...
+        # Test instantiation with parameters
+        DSParams(enable_deepspeed=True, zero_stage=ZeRORuntimeStage.ZERO_3)
+        print("✓ DSParams with parameters instantiation successful")
+
+        # Test instantiation with nested dataclass
+        DSParams(enable_deepspeed=True, offload_optimizer=OffloadConfig())
+        print("✓ DSParams with nested OffloadConfig instantiation successful")
+
+        # Test that validation works
+        try:
```
wizeng23: Could you use pytest.raises for this instead? Ex. tests/unit/core/configs/params/test_evaluation_params.py

tests/unit/core/configs/params/test_ds_params_instantiation.py
```diff
@@ -0,0 +1,64 @@
...
+sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent.parent.parent))
+
+from oumi.core.configs.params.ds_params import DSParams, OffloadConfig, ZeRORuntimeStage
+
+
+def test_dsparams_instantiation():
+    """Test that DSParams can be instantiated without errors."""
+    print("Testing DSParams instantiation...")
+
+    try:
```
wizeng23: Could we have the implementation of this test be in line with our other tests?
- Since it's run with multiple other tests, print lines generally aren't necessary as they'll get buried
- Use assert if you want to verify certain values
- IMO there's no need for the outer try except loop

tests/unit/core/configs/params/test_ds_params_omegaconf.py
```diff
@@ -0,0 +1,168 @@
...
+from oumi.core.configs.params.ds_params import (
+    DeepSpeedOffloadDevice,
+    DeepSpeedPrecision,
+    DSParams,
+    OffloadConfig,
+    ZeRORuntimeStage,
+)
+
+
+class TestDSParamsOmegaConf:
```
wizeng23: Does there need to be a class for this? This should be the same as having all the test fns directly in the file right?

tests/unit/core/configs/params/test_ds_params_instantiation.py
```diff
@@ -0,0 +1,64 @@
...
+import sys
+from pathlib import Path
+
+# Add root directory to path for imports
+sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent.parent.parent))
+
+from oumi.core.configs.params.ds_params import DSParams, OffloadConfig, ZeRORuntimeStage
+
+
+def test_dsparams_instantiation():
```
wizeng23: IMO if this is just one test, it doesn't need to be in its own file. Consider merging the two test files into `test_ds_params.py` to be in line with other param tests

wizeng23:
Also, I don't see this change that's mentioned in the PR description: `Updates to the pyproject.toml with pinned, compatible DeepSpeed versions`

scripts/deepspeed/zero_to_fp32.py
penfever: I'm going to remove the script from the PR to keep things simple. We will need this script or something like it to combine the sharded checkpoints Deepspeed produces during training, though. We can link to it in the documentation if you prefer. https://deepspeed.readthedocs.io/en/latest/_modules/deepspeed/utils/zero_to_fp32.html

src/oumi/core/configs/params/ds_params.py
penfever: Yes!

penfever: Some of them are modified as of this latest commit ...

reduce_bucket_size: Union[int, str] = "auto"
    - Oumi: "auto" (delegated to transformers/DeepSpeed)
    - DeepSpeed: 500,000,000 elements

stage3_prefetch_bucket_size: Union[int, str] = "auto"
    - Oumi: "auto"
    - DeepSpeed: 50,000,000 elements

stage3_param_persistence_threshold: Union[int, str] = "auto"
    - Oumi: "auto"
    - DeepSpeed: 100,000 elements

src/oumi/core/configs/params/ds_params.py
penfever: Yep!

src/oumi/core/configs/params/ds_params.py
penfever: This takes precedence over auto-generated configuration. Parameters specified in the JSON will override DeepSpeed defaults; unspecified parameters will use DeepSpeed's internal defaults (not Oumi's defaults). I added a clarifying comment.

src/oumi/core/configs/params/ds_params.py
penfever: With enable_deepspeed=True and zero_stage=ZERO_0, the DeepSpeed engine is initialized and active but ZeRO optimizations are disabled; so it will use Deepspeed engine for training and Deepspeed internal default parameter settings unless they are overridden in the config. I'll clarify this in the comments.

src/oumi/core/configs/params/ds_params.py
penfever: I updated the DType from str to Union[int, str] to properly reflect that it can accept both integers and the special "auto" string value. However, when using TRL trainers, it should remain "auto" to allow proper batch size management.

src/oumi/core/configs/params/ds_params.py
penfever: OK

src/oumi/core/configs/params/ds_params.py
penfever: When a user provides a deepspeed_config_path JSON file, it completely bypasses the to_deepspeed() method; so they can have complete control if they require it.

src/oumi/core/distributed.py
penfever: Good catch, I'll remove it

tests/unit/core/configs/params/test_ds_params_instantiation.py
penfever: I removed the line

tests/unit/core/configs/params/test_ds_params_instantiation.py
penfever: Nope! I cleaned it up

tests/unit/core/configs/params/test_ds_params_instantiation.py
penfever: Yes!

tests/unit/core/configs/params/test_ds_params_instantiation.py
penfever: This should be cleaned up now

tests/unit/core/configs/params/test_ds_params_instantiation.py
penfever: Done!

.github/workflows/gpu_tests.yaml
wizeng23, line 1: Could you explain why the changes in this file are needed? Also, is there a reason you're using the 12.4 container? Our tests were previously running on 12.8.

pyproject.toml
```diff
@@ -125,6 +125,7 @@ gpu = [
     # src/oumi/core/trainers/verl_trainer_config.yaml.
     "verl>=0.5.0,<0.6",   # Used for the VERL_GRPO trainer.
     "vllm>=0.10.0,<0.11", # For VLLMInferenceEngine, and vLLM-powered GRPO training.
+    "deepspeed>=0.17.0,<=0.17.4"
```
wizeng23: Is there a reason for not capping the version at <0.18.0? If there's a specific reason for excluding 0.17.5 (which is already released), could you add a comment in the file?

pyproject.toml
```diff
@@ -261,6 +269,7 @@ pythonVersion = "3.9"
 pythonPlatform = "All" # Include linux, mac, windows
 
 reportPrivateImportUsage = "none"
+reportMissingImports = "none"
```
wizeng23: This seems like a big change. Are you able to set it in your own personal VSCode workspace instead?

scripts/deepspeed/zero_to_fp32.py
wizeng23: Adding a doc link sounds good to me!

src/oumi/core/configs/params/ds_params.py
wizeng23: Thanks! Could you also rename the file to deepspeed_params.py?

src/oumi/core/configs/params/ds_params.py
wizeng23: Thanks! Can the same change be made to the fields below this as well, assuming the same thing is true?

src/oumi/core/configs/params/ds_params.py
wizeng23: SG! In general, one thing I try to avoid is setting default values that differ from the underlying framework's so users used to the underlying framework don't experience unexpected behavior. Ex: #1124.

I don't think it's an issue in this case if the auto value falls back to the Deepspeed default.

src/oumi/core/configs/params/ds_params.py
wizeng23: Thanks! If my understanding is correct, all other fields the user specifies in this param are ignored if this field is set right? Could you explicitly mention this? So far the comment just mentions Oumi defaults

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,430 @@
...
+
+    reduce_bucket_size: Union[int, str] = "auto"
+    """Bucket size for gradient reduction.
+
+    Can be an integer or "auto" for automatic sizing.
+    """
+
+    # ZeRO-2 specific communication parameters
+    allgather_bucket_size: int = int(5e8)
+    """Bucket size for allgather operations (ZeRO stages 0-2)."""
```
wizeng23:
```suggestion
    """Bucket size for allgather operations (ZeRO stages 1 and 2)."""
```

src/oumi/core/configs/params/ds_params.py
```diff
@@ -0,0 +1,430 @@
...
+    contiguous_gradients: bool = True
+    """Whether to ensure gradient memory is contiguous."""
+
+    reduce_bucket_size: Union[int, str] = "auto"
+    """Bucket size for gradient reduction.
+
+    Can be an integer or "auto" for automatic sizing.
+    """
+
+    # ZeRO-2 specific communication parameters
```
wizeng23:
```suggestion
    # ZeRO-1 and ZeRO-2 specific communication parameters
```

src/oumi/core/configs/params/ds_params.py
wizeng23: Could you move the other entries as well? Also just to confirm, are these params zero 2 only, or also apply to 1?

tests/scripts/predownload_for_github_gpu_tests.sh
wizeng23, line 1: Thanks for improving the GPU tests!!

configs/examples/deepspeed/llama3_1_8b_deepspeed_z2.yaml
wizeng23, line 1: 1. nit: Could you suffix the config names with '_train.yaml'?
2. Could you add a comment at the top of the configs for the minimum hardware requirement needed to run them (ex. # GPUs needed to not OOM)?

penfever: Sure!

pyproject.toml
penfever: Yep!

src/oumi/core/configs/params/ds_params.py
penfever: All other Deepspeed user-specified fields in the Oumi YAML as well as any Deepspeed default values will be overridden by settings in the config file passed in this param, if this param is passed a config, is my best understanding of the flow. I will try to clarify! :)

src/oumi/core/configs/params/ds_params.py
penfever: I'll set them to defaults

.github/workflows/gpu_tests.yaml
penfever: DeepSpeed C++/CUDA extensions are built just-in-time (JIT) using torch's JIT C++ extension loader, so nvcc is required to be in the environment. It also requires that CUDA_HOME Environment Variable be set for DeepSpeed to locate CUDA installation. In the previous testing environment, both of these requirements were failing.

By modifying it to run in a Docker container we inherit nvcc in the environment without requiring further installations, which could be slow and failure-prone.

https://www.deepspeed.ai/tutorials/advanced-install/

I don't remember the specific error that the libcurl4 install / version bump resolved any longer, it was something in the uv install. 

The echoes are just for debugging convenience in the environment.

The SETUPTOOLS_SCM_PRETEND_VERSION_FOR_OUMI=v0.0.1 environment variable was also set in gpu_install_test.yaml, it's because containers need a workaround or setuptools-scm fails to detect the Oumi version.

At the time I first submitted this PR I believe we were still on 12.4 -- I'll update to a 12.8.1 container.

src/oumi/core/configs/params/ds_params.py
penfever: OK!

src/oumi/train.py
penfever: Good idea

src/oumi/core/configs/params/ds_params.py
penfever: Sorry, not sure which other fields you're referring to here?

pyproject.toml
penfever: Not AFAIK, I'll relax it

configs/recipes/qwen2_5/sft/qwen2.5_7b_limo_fft.yaml
wizeng23, line 1: For this and the other Qwen configs, could you mention the required hardware (ex. VRAM amount)?

configs/examples/deepspeed/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `llama3_1_8b_deepspeed_z3_offload_train.yaml` - ZeRO-3 with CPU offloading for maximum memory efficiency
+
+## Checkpoint Reconstruction
+
+After training with DeepSpeed ZeRO-2 or ZeRO-3, model parameters are sharded across multiple files. To reconstruct a complete model checkpoint from these fragments, use DeepSpeed's `zero_to_fp32` utility:
+
+```bash
+# Reconstruct checkpoint from ZeRO fragments
+wget https://raw.githubusercontent.com/microsoft/DeepSpeed/master/deepspeed/utils/zero_to_fp32.py
+python zero_to_fp32.py /path/to/checkpoint/directory /path/to/output/pytorch_model.bin
```
wizeng23, line 18: In your experience, will people want to further convert it into a HF-compatible Safetensors format? Is there a way to convert to that as well?

wizeng23 (approved):

PR #1899, opened 2025-08-04
Fix quantization notebook related errors

shanghongsim:
# Description

Remove requirements for output_format="pytorch" and update the error catching for `NVML_ERROR_NOT_SUPPORTED` as `NVMLError_NotSupported` is not a valid attribute of `pynvml`.

## Related issues

Fixes OPE-1472

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/device_utils.py
+++ b/src/oumi/utils/device_utils.py
@@ -16,6 +16,8 @@ from collections.abc import Sequence
 from pprint import pformat
 from typing import NamedTuple, Optional
 
+from pynvml import NVML_ERROR_NOT_SUPPORTED, NVMLError
+
 from oumi.utils.logging import logger
 
 try:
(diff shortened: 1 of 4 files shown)


src/oumi/utils/device_utils.py
```diff
@@ -144,10 +146,13 @@ def _get_nvidia_gpu_runtime_info_impl(
 
     try:
         gpu_handle = pynvml.nvmlDeviceGetHandleByIndex(device_index)
-    except pynvml.NVMLError_NotSupported:
+    except NVMLError as e:
```
oelachqar: @shanghongsim is this needed for this PR?

src/oumi/utils/device_utils.py
```diff
@@ -16,6 +16,8 @@
 from pprint import pformat
 from typing import NamedTuple, Optional
 
+from pynvml import NVML_ERROR_NOT_SUPPORTED, NVMLError
```
oelachqar: Importing explicitely could cause problems if the user does not have `pynvml` installed (it's an optional dependency)

src/oumi/utils/device_utils.py
shanghongsim: Yup, pyright was telling me that `pynvml.NVMLError_NotSupported` wasn't allowed because that attribute doesn't exist

src/oumi/utils/device_utils.py
shanghongsim: I shifted it into the try block!

oelachqar (approved):

PR #1908, opened 2025-08-06
Update dependencies

oelachqar:
# Description

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


pyproject.toml
```diff
@@ -58,24 +58,24 @@ dependencies = [
...
+    "pillow>=11.1.0,<11.4",      # Used by image datasets
     "protobuf>=5.29.0",
-    "pydantic>=2.9.2,<2.10",
+    "pydantic>=2.9.2,<2.12",
     "responses>=0.25.0,<0.26",
     "safetensors>=0.5.0,<0.6.0",
     "skypilot>=0.7.0,<0.8",      # Used by launcher
     "tensorboard>=2.18.0,<2.19", # Optional, for monitoring training
-    "torch>=2.6.0,<2.7.0",
+    "torch>=2.7.1,<2.9.0",
```
wizeng23, line 68: Any reason to not constrain to >=2.8.0? Is it because it's too new? What ends up getting installed with this requirement?

.github/workflows/install_test.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-        torch-version: ["2.6.0", "2.7.0"]
+        torch-version: ["2.6.0", "2.7.1", "2.8.0"]
```
wizeng23, line 21: Shouldn't need 2.6.0 here right?

oelachqar: I think it's ok to keep the backward compatibility, as long as it's not causing issues

pyproject.toml
oelachqar: To give more flexibily -- our version of vllm requires v2.7.1

wizeng23 (approved):

wizeng23 (approved):

PR #1914, opened 2025-08-06
Add Conversation-Level Analysis 

ryan-arman:
# Description
This PR adds dual-level conversation analysis - enabling both message-level and conversation-level analysis of datasets.

Main changes: 
* SampleAnalyzer has a analyze_sample method (instead of analyze_message) which computes both message level and conversation level metrics and returns tuple[list[MessageAnalysisResult], ConversationAnalysisResult]
* LengthAnalyzer implements SampleAnalyzer in a way that conversation level metrics are the aggregate of message level metrics for all of the metrics (char, word, sentence) except for token. For token, it tokenizes the conversation directly using the datast.tokenize
* dataset is passed to the LengthAnalyzer so it can use it for tokenziation
* we can remove the tokenizer in a follow up PR since we are going to use the dataset directly
* All of the tests are updated accordingly
* The check for sample_count has been moved to analyze_config

## Related issues

Fixes OPE-1455

## Before submitting

## Reviewers


src/oumi/core/analyze/length_analyzer.py
```diff
@@ -36,6 +41,7 @@ def __init__(
         token_count: bool = False,
         tokenizer: Optional[Union[PreTrainedTokenizer, PreTrainedTokenizerFast]] = None,
         include_special_tokens: bool = True,
```
oelachqar: What kind of dataset do we expect?

src/oumi/core/analyze/length_analyzer.py
```diff
@@ -49,26 +55,158 @@ def __init__(
...
+                total_sentences = sum(
+                    msg.analyzer_metrics.get("sentence_count", 0)
+                    for msg in message_results
+                )
+                conversation_metrics["sentence_count"] = total_sentences
+
+        # For token count, use the tokenized conversation
+        if self.token_count and self.dataset is not None:
+            # Get tokenized result directly and extract token count
+            tokenized_result = self.dataset.tokenize(conversation, tokenize=True)
```
oelachqar: Should we use the provided tokenizer / processor?

src/oumi/core/analyze/length_analyzer.py
ryan-arman: removed dataset and used apply_chat_template directly

src/oumi/core/analyze/length_analyzer.py
ryan-arman: good point, applied apply_chat_template directly as we discussed offline

tests/unit/core/configs/test_analyze_config.py
```diff
@@ -138,3 +138,29 @@ def test_analyze_config_processor_fields_custom_values():
...
+    """Test that valid sample_count values are accepted."""
+    # Should not raise any exception
+    config = AnalyzeConfig(dataset_name="test_dataset", sample_count=1)
+    assert config.sample_count == 1
+
+    config = AnalyzeConfig(dataset_name="test_dataset", sample_count=100)
+    assert config.sample_count == 100
+
+    # None should also be valid
+    config = AnalyzeConfig(dataset_name="test_dataset", sample_count=None)
```
jgreer013, line 165: nit: I assume "None" equates to the entire dataset?

src/oumi/core/analyze/length_analyzer.py
```diff
@@ -63,12 +68,147 @@ def __init__(
...
+                )
+                conversation_metrics["sentence_count"] = total_sentences
+
+        # For token count, tokenize the full rendered conversation using the tokenizer's
+        # chat template for full control (independent of dataset internals).
+        if self.token_count:
+            tokenizer_to_use = self.tokenizer
+            if tokenizer_to_use is not None:
+                # First render the conversation to a prompt string via chat template
+                prompt_text = tokenizer_to_use.apply_chat_template(
```
jgreer013, line 189: nit: I think technically if you set tokenize=True, the encode step is no longer needed?

tests/unit/core/configs/test_analyze_config.py
ryan-arman: yes and that is the default value for sample_count

src/oumi/core/analyze/length_analyzer.py
ryan-arman: the only difference is that here we explicitly set include_special_tokens to the input whereas when calling apply_chat_template internally it calls tokenize with add_special_tokens=False. I will keep it as is to have more control

PR #1916, opened 2025-08-06
Add processor support to AnalyzeConfig for vision-language datasets

ryan-arman:
# Description
For vision datasets we need to pass a processor 

## Related issues

Fixes OPE-1479

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/analysis_utils.py
+++ b/src/oumi/utils/analysis_utils.py
@@ -46,12 +46,22 @@ def load_dataset_from_config(config: AnalyzeConfig) -> BaseMapDataset:
                 "dataset_path": None,
                 "split": split,
                 "subset": subset,
-                "trust_remote_code": False,
+                "trust_remote_code": getattr(config, "trust_remote_code", False),
             }
 
             # Add tokenizer if provided
             if tokenizer is not None:
                 dataset_kwargs["tokenizer"] = tokenizer
+
+            # Add processor parameters for vision-language datasets
+            if hasattr(config, "processor_name") and config.processor_name:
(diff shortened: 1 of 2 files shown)


src/oumi/utils/analysis_utils.py
```diff
@@ -46,13 +46,23 @@ def load_dataset_from_config(config: AnalyzeConfig) -> BaseMapDataset:
...
                 "subset": subset,
-                "trust_remote_code": False,
+                "trust_remote_code": getattr(config, "trust_remote_code", False),
             }
 
             # Add tokenizer if provided
             if tokenizer is not None:
                 dataset_kwargs["tokenizer"] = tokenizer
 
+            # Add processor parameters for vision-language datasets
```
oelachqar, line 56: Could we use the attributes directly? Why do we need hasattr and getattr?

ryan-arman: yes good point, made the change

src/oumi/utils/analysis_utils.py
```diff
@@ -46,13 +46,19 @@ def load_dataset_from_config(config: AnalyzeConfig) -> BaseMapDataset:
                 "dataset_path": None,
                 "split": split,
                 "subset": subset,
-                "trust_remote_code": False,
+                "trust_remote_code": getattr(config, "trust_remote_code", False),
```
oelachqar: same here, why the getattr?

ryan-arman: sure, done

PR #1920, opened 2025-08-08
Penfever/oss clean2

penfever:
# Description

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
+++ b/configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
@@ -19,8 +19,7 @@ model:
   torch_dtype_str: "bfloat16"
   trust_remote_code: true
   chat_template: "auto"  # Use model's built-in chat template
-  attn_implementation: "flash_attention"  # Auto-detects best available Flash Attention
-  enable_hf_kernels: true  # Enable HF kernels for optimal FA3 performance
+  attn_implementation: "flash_attention_2"  # Use Flash Attention 2
 
   # Enable MXFP4 quantization for efficient inference
   model_kwargs:
--- a/configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
+++ b/configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
@@ -1,0 +1,39 @@
+# Inference config for gpt-oss-120b (MoE model with 5.1B active parameters)
+# using vLLM for local inference with MXFP4 quantization.
+#
+# Usage:
+#   oumi infer -i -c oumi://configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
+#
+# Requirements:
+#   - Single H100 GPU (80GB) or equivalent
+#   - vLLM with GPT OSS support installed
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/infer/infer.html
+#   - Config class: oumi.core.configs.InferenceConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/inference_config.py
+#   - Other inference configs: configs/**/inference/
+
+model:
+  model_name: "openai/gpt-oss-120b"
+  torch_dtype_str: "bfloat16"
+  trust_remote_code: true
+  chat_template: "auto"  # Use model's built-in chat template
+  attn_implementation: "flash_attention"  # Auto-detects best available Flash Attention
+  enable_hf_kernels: true  # Enable HF kernels for optimal FA3 performance
+
+  # Enable MXFP4 quantization for efficient inference
+  model_kwargs:
+    quantization_config:
+      quant_method: "mxfp4"
+      bits: 4
+      group_size: 128
+
+# Use vLLM engine for local inference
+engine: VLLM
+
+# Generation parameters for reasoning model
+generation:
+  temperature: 0.7
+  max_new_tokens: 8192
+  skip_special_tokens: false  # Keep reasoning tokens for GPT OSS models
--- a/configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
+++ b/configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
@@ -1,0 +1,89 @@
+# Single-GPU LoRA fine-tuning config for GPT OSS 20B
+# Optimized for training on a single high-memory GPU (recommended: 80GB+)
+# Replicates HF GPT-OSS recipes single GPU approach with memory optimizations
+#
+# This config uses aggressive memory optimization techniques:
+#   - Small batch size with high gradient accumulation
+#   - Gradient checkpointing
+#   - Mixed precision training
+#   - Efficient LoRA parameters
+#
+# Usage:
+#   oumi train -c configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
+#
+# Hardware Requirements:
+#   - Single GPU with 80GB+ VRAM (H100, A100 80GB)
+#   - Or 40GB GPU with gradient checkpointing + smaller batch
+#
+# See Also:
+#   - HF GPT-OSS recipes for single GPU training
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/train/train.html
+
+model:
+  model_name: "openai/gpt-oss-20b"
+  model_max_length: 8192
+  torch_dtype_str: "bfloat16"
+  trust_remote_code: true
+  chat_template: "auto"  # Use model's built-in chat template
+  attn_implementation: "kernels-community/vllm-flash-attn3"  # Auto-detects best available Flash Attention
+
+  # Enable MXFP4 dequantization for training with pre-quantized models
+  model_kwargs:
+    quantization_config:
+      quant_method: "mxfp4"
+      bits: 4
+      group_size: 128
+
+data:
+  train:
+    datasets:
+      - dataset_name: "HuggingFaceH4/Multilingual-Thinking"
+    collator_name: "text_completions_only_with_padding"
+    collator_kwargs:
+      # GPT-OSS Harmony format templates
+      instruction_template: "<|start|>user<|message|>"
+      response_template: "<|start|>assistant<|channel|>final<|message|>"
+    target_col: "messages"
+
+training:
+  trainer_type: "TRL_SFT"
+  use_peft: true
+  ddp_find_unused_parameters: false
+  save_steps: 500
+  num_train_epochs: 1
+  # Small batch size for memory efficiency
+  per_device_train_batch_size: 1
+  # High gradient accumulation to maintain effective batch size
+  gradient_accumulation_steps: 4
+  enable_gradient_checkpointing: true
+  gradient_checkpointing_kwargs:
+    use_reentrant: false
+  optimizer: "adamw_torch_fused"
+  learning_rate: 2.0e-04  # HF recipe LoRA learning rate
+  warmup_ratio: 0.03
+  lr_scheduler_type: "cosine"
+  lr_scheduler_kwargs:
+    num_cycles: 0.5
+  max_grad_norm: 1.0
+  logging_steps: 5  # More frequent logging for single GPU
+  output_dir: "./outputs/gpt_oss_20b_lora_single_gpu"
+  seed: 42
+  # Reduce workers for single GPU to avoid CPU bottleneck
+  dataloader_num_workers: 4
+
+  # Performance optimizations
+  include_performance_metrics: true
+
+peft:
+  q_lora: false  # Regular LoRA for best compatibility
+  lora_r: 8        # HF recipe efficient rank
+  lora_alpha: 16   # HF recipe alpha
+  lora_dropout: 0  # No dropout for stability
+  lora_target_modules:
+    - "q_proj"
+    - "k_proj"
+    - "v_proj"
+    - "o_proj"
+    - "gate_proj"
+    - "up_proj"
+    - "down_proj"
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -125,11 +125,28 @@ def build_data_collator(
             **kwargs,
         )
     elif collator_name == "text_completions_only_with_padding":
+        # Extract instruction and response templates from kwargs if provided
+        instruction_template = kwargs.pop("instruction_template", None)
+        response_template = kwargs.pop("response_template", None)
+
+        # Default to Llama-style templates if not provided
+        instruction_prefix = (
+            instruction_template
+            if instruction_template
+            else "<|start_header_id|>user<|end_header_id|>\n\n"
+        )
+        response_prefix = (
+            response_template
+            if response_template
+            else "<|start_header_id|>assistant<|end_header_id|>\n\n"
+        )
+
         return TextCompletionsCollatorWithPadding(
             tokenizer=tokenizer,
-            instruction_prefix="<|start_header_id|>user<|end_header_id|>\n\n",
-            response_prefix="<|start_header_id|>assistant<|end_header_id|>\n\n",
+            instruction_prefix=instruction_prefix,
+            response_prefix=response_prefix,
             debug=debug,
+            **kwargs,
         )
     raise ValueError(f"Unknown data collator name: '{collator_name}'")
 
--- a/src/oumi/builders/models.py
+++ b/src/oumi/builders/models.py
@@ -20,6 +20,14 @@ import torch.nn as nn
 import transformers
 from peft import LoraConfig, PeftModel, get_peft_model, prepare_model_for_kbit_training
 
+try:
+    from transformers import Mxfp4Config
+
+    MXFP4_AVAILABLE = True
+except ImportError:
+    MXFP4_AVAILABLE = False
+    Mxfp4Config = None
+
 from oumi.core.configs import LoraWeightInitialization, ModelParams, PeftParams
 from oumi.core.configs.internal.internal_model_config import InternalModelConfig
 from oumi.core.configs.internal.supported_models import (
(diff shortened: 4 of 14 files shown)


src/oumi/builders/collators.py
```diff
@@ -125,11 +125,28 @@ def build_data_collator(
             **kwargs,
         )
     elif collator_name == "text_completions_only_with_padding":
```
wizeng23, line 127: 1. Is training on text completions only necessary for gpt? I don't see it mentioned in their training notebook. Do most SFT jobs in general want to only train on completions?
2. This seems to be a bigger issue that affects not just gpt but also other models like Qwen. We should consider spinning this logic out into a separate PR.

src/oumi/builders/models.py
```diff
@@ -20,6 +20,14 @@
 import transformers
 from peft import PeftModel, get_peft_model, prepare_model_for_kbit_training
 
+try:
```
wizeng23: Don't think this should be necessary now that we require 4.55

src/oumi/builders/models.py
```diff
@@ -184,6 +192,44 @@ def build_oumi_model(
...
+        model_params: Model parameters that may contain quantization config.
+
+    Returns:
+        Quantization config object or None if no quantization needed.
+    """
+    # Check if model_kwargs contains quantization_config
+    if not model_params.model_kwargs:
+        return None
+
+    quant_config = model_params.model_kwargs.get("quantization_config")
```
wizeng23, line 204: Having it as a model kwarg is a bit hacky as opposed to a built-in field, but we can proceed with this for now to unblock, and add an issue to make it more formal (could merge with some of the qlora fields in PeftParams perhaps)

src/oumi/builders/models.py
```diff
@@ -478,7 +525,14 @@ def build_tokenizer(
         tokenizer.model_max_length = model_params.model_max_length
 
     template_name: str = ""
-    if model_params.chat_template:
+    if model_params.chat_template == "auto":
```
wizeng23, line 514: I don't think this is necessary. As long as we don't set model_params.chat_template, don't register anything in the internal config, it'll use the model's built-in chat template if it has one.

src/oumi/core/configs/internal/internal_model_config.py
```diff
@@ -144,3 +144,15 @@ class InternalModelConfig(BaseConfig):
 
     visual_config: Optional[InternalVisualModelConfig] = None
     """Configuration specific to visual models."""
+
+    quantization_support: list[str] = field(default_factory=list)
```
wizeng23: Are these changes needed?

src/oumi/inference/vllm_inference_engine.py
```diff
@@ -33,7 +33,12 @@
 
 try:
     import vllm  # pyright: ignore[reportMissingImports]
-    from vllm.config import ModelDType  # pyright: ignore[reportMissingImports]
+
+    try:
+        from vllm.config import ModelDType  # pyright: ignore[reportMissingImports]
+    except ImportError:
+        # For compatibility with newer vLLM versions
```
wizeng23, line 40: We depend on vllm 0.10.0, which I think is the latest one. Is there issues at that version?

src/oumi/inference/vllm_inference_engine.py
```diff
@@ -129,6 +144,12 @@ def __init__(
         if quantization and quantization == "bitsandbytes":
             vllm_kwargs["load_format"] = "bitsandbytes"
             logger.info("VLLM engine loading a `bitsandbytes` quantized model.")
+        elif quantization and quantization == "mxfp4":
```
wizeng23, line 147: Is this and the below logic only needed because mxfp4 is not supported as of vllm 0.10.0? Could this just be avoided by putting a `# type: ignore` comment on the quantization arg?

src/oumi/inference/vllm_inference_engine.py
```diff
@@ -258,6 +284,7 @@ def _infer(
             stop_token_ids=generation_params.stop_token_ids,
             min_p=generation_params.min_p,
             guided_decoding=guided_decoding,
+            skip_special_tokens=generation_params.skip_special_tokens,
```
wizeng23, line 289: We should add this to native inference engine as well

src/oumi/quantize/mxfp4_quantizer.py
wizeng23, line 1: This is independent of training and inference right?

src/oumi/builders/collators.py
penfever: This change was the most straightforward way I could find to reproduce the results from https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers in Oumi while respecting the Harmony chat format.

If the collator changes need to be a stand-alone PR we can make that happen.

src/oumi/builders/models.py
penfever: Agree, will fix

src/oumi/builders/models.py
penfever: SGTM -- I can add an issue but I'm not quite sure what it would be, that quantization configs need to be added as built-in fields?

src/oumi/builders/models.py
penfever: That wasn't the behavior I was observing, for whatever reason. It was failing to locate a local jinja template for the model and then erroring out. We don't need the logging of course, but is there a downside to making it explicit?

src/oumi/core/configs/internal/internal_model_config.py
penfever: No, I think we can remove them

src/oumi/inference/vllm_inference_engine.py
penfever: I believe so. Not sure what you have been able to get working but I can't get gpt-oss to work with anything except for the very specific post-0.10.0 build they recommend.

uv pip install --pre vllm==0.10.1+gptoss \
    --extra-index-url https://wheels.vllm.ai/gpt-oss/ \
    --extra-index-url https://download.pytorch.org/whl/nightly/cu128 \
    --index-strategy unsafe-best-match

src/oumi/quantize/mxfp4_quantizer.py
penfever: It's independent of VLLM inference and training (which is bf16 anyway), correct

penfever: Removing this for now (will disable "oumi quantize" for mxfp4, we can re-enable this in a later commit)

src/oumi/builders/collators.py
wizeng23: I think we can keep the change here.

Did their cookbook train on completions only? I can't see any reference to it there, so I would've thought that to achieve parity with the training they have there, we can just not specify a collator. Please correct me if I'm wrong though

src/oumi/builders/models.py
wizeng23: Given that this seems to be an existing pattern we're using for loading quantized models for vLLM engine inference, I'm okay keeping this as a kwarg. We can promote it to a built-in field later if we see the need.

src/oumi/builders/models.py
```diff
@@ -478,11 +518,14 @@ def build_tokenizer(
...
+        # "auto" means use model's built-in template, don't override
         logger.info(
-            f"Using the chat template '{model_params.chat_template}' "
-            f"specified in model config for {tokenizer_id_str}. "
+            f"Chat template set to 'auto' - using model's built-in template "
+            f"for {tokenizer_id_str}."
         )
+        template_name = ""  # Don't set any template name to avoid override
```
wizeng23: nit: don't need this line since it's set above

src/oumi/builders/models.py
wizeng23: Ah, it might be because you added an internal model config which triggered some different logic. Without the internal config, it should fall back to the built-in chat template I think.

Agreed it doesn't hurt to allow an explicit option. If you don't mind adding some quick documentation about it to src/oumi/core/configs/params/model_params.py, that'd be great!

src/oumi/inference/vllm_inference_engine.py
wizeng23: Agreed that gpt-oss vLLM inference only works with their dev version. Do you get errors without this change when running 0.10.1+gptoss? It's weird since it still seems defined at main: https://github.com/vllm-project/vllm/blob/177e55e3bd3dbb54089d9062b763a413c8718dff/vllm/config/__init__.py#L246.

Not a big deal though, it's just a typing import, happy to keep this in if it makes things work

src/oumi/inference/vllm_inference_engine.py
wizeng23: Fine to keep this for now; I do want to revisit it later to see if this could be cleaned up after the next vllm patch version release

src/oumi/builders/models.py
```diff
@@ -184,6 +185,44 @@ def build_oumi_model(
...
+    # Handle MXFP4 quantization for training (requires dequantization)
+    if isinstance(quant_config, dict) and quant_config.get("quant_method") == "mxfp4":
+        try:
+            logger.info(
+                "Detected MXFP4 quantized model. Creating Mxfp4Config(dequantize=True) "
+                "for training."
+            )
+            return Mxfp4Config(dequantize=True)  # pyright: ignore[reportOptionalCall]
+        except Exception as e:
+            raise ImportError(
```
wizeng23: Shouldn't need this as well

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
```diff
@@ -0,0 +1,38 @@
...
+  trust_remote_code: true
+  chat_template: "auto"  # Use model's built-in chat template
+  attn_implementation: "flash_attention_2"  # Use Flash Attention 2
+
+  # Enable MXFP4 quantization for efficient inference
+  model_kwargs:
+    quantization_config:
+      quant_method: "mxfp4"
+      bits: 4
+      group_size: 128
```
wizeng23: Don't think these two fields are being ingested right? Can we delete? Ditto for other configs

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
```diff
@@ -0,0 +1,38 @@
...
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/infer/infer.html
+#   - Config class: oumi.core.configs.InferenceConfig
+#   - Config source: https://github.com/oumi-ai/oumi/blob/main/src/oumi/core/configs/inference_config.py
+#   - Other inference configs: configs/**/inference/
+
+model:
+  model_name: "openai/gpt-oss-120b"
+  torch_dtype_str: "bfloat16"
```
wizeng23, line 19: We don't want to set this right? Ditto for other configs

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
```diff
@@ -0,0 +1,38 @@
...
+  trust_remote_code: true
+  chat_template: "auto"  # Use model's built-in chat template
+  attn_implementation: "flash_attention_2"  # Use Flash Attention 2
```
wizeng23: Can delete this line, it's actually not used for vLLM inference

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
```diff
@@ -0,0 +1,89 @@
...
+# Replicates HF GPT-OSS recipes single GPU approach with memory optimizations
+#
+# This config uses aggressive memory optimization techniques:
+#   - Small batch size with high gradient accumulation
+#   - Gradient checkpointing
+#   - Mixed precision training
+#   - Efficient LoRA parameters
+#
+# Usage:
+#   oumi train -c configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
```
wizeng23: nit: Add oumi:// prefix

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
wizeng23, line 1: nit: Could you rename to 20b_lora_single_gpu_train.yaml?

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
```diff
@@ -0,0 +1,89 @@
...
+#   - HF GPT-OSS recipes for single GPU training
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/train/train.html
+
+model:
+  model_name: "openai/gpt-oss-20b"
+  model_max_length: 8192
+  torch_dtype_str: "bfloat16"
+  trust_remote_code: true
+  chat_template: "auto"  # Use model's built-in chat template
+  attn_implementation: "kernels-community/vllm-flash-attn3"  # Auto-detects best available Flash Attention
```
wizeng23: They actually set it to eager attn in the OpenAI cookbook, for "better performance": https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers#prepare-the-model. Can we change to that to align with their recommendations?

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  # Performance optimizations
+  include_performance_metrics: true
+
+peft:
+  q_lora: false  # Regular LoRA for best compatibility
+  lora_r: 8        # HF recipe efficient rank
+  lora_alpha: 16   # HF recipe alpha
+  lora_dropout: 0  # No dropout for stability
+  lora_target_modules:
```
wizeng23: FYI this may not work as expected because MoE layers sometimes are implemented with nn.Parameter instead of nn.Module, so searching for them with a module name won't work.

If you want to copy what they did in the cookbook, you can use:
```
  lora_target_modules:
    - "all-linear"
  lora_target_parameters:
    - "7.mlp.experts.gate_up_proj"
    - "7.mlp.experts.down_proj"
    - "15.mlp.experts.gate_up_proj"
    - "15.mlp.experts.down_proj"
    - "23.mlp.experts.gate_up_proj"
    - "23.mlp.experts.down_proj"
```

See: https://github.com/oumi-ai/oumi/pull/1925

src/oumi/core/configs/params/generation_params.py
```diff
@@ -120,6 +120,15 @@ class GenerationParams(BaseParams):
     guided_decoding: Optional[GuidedDecodingParams] = None
     """Parameters for guided decoding."""
 
+    skip_special_tokens: bool = True
```
wizeng23, line 123: Do you mind setting this to false in configs/recipes/gpt_oss/inference/20b_infer.yaml and configs/recipes/gpt_oss/inference/120b_infer.yaml? Thanks!

src/oumi/inference/vllm_inference_engine.py
penfever: SGTM -- I'll add a comment to that effect

src/oumi/builders/collators.py
penfever: Yeah, GPT-OSS releases were a bit confusing in how they presented the information! But per https://huggingface.co/blog/welcome-openai-gpt-oss ( Chats and Chat Templates ) --

Mask the labels on all turns except the final assistant turn, or else you will be training it on the previous turns without chains of thought, which will teach it to emit responses without CoTs. This means that you cannot train on an entire multi-turn conversation as a single sample; instead, you must break it into one sample per assistant turn with only the final assistant turn unmasked each time.

src/oumi/builders/models.py
penfever: Sure!

src/oumi/inference/vllm_inference_engine.py
penfever: I don't require that change to avoid errors as far as I know, unless the above command is automatically pulling that code. Let's keep it if you don't mind, since I'm confident this works even if it's a bit more verbose. :)

src/oumi/builders/models.py
penfever: Will remove

src/oumi/core/configs/params/generation_params.py
penfever: Sure

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
penfever: SGTM!

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
penfever: Yep

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
penfever: Yep

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
penfever: Yep!

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
penfever: OK!

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
penfever: I think it's safer to leave this explicitly set to "bfloat16", a valid dtype for the model, rather than "auto"; current versions would probably resolve to "bfloat16" in any case but after VLLM updates the behavior might change.

configs/recipes/gpt_oss/sft/20b_lora_single_gpu.yaml
penfever: Yep!

src/oumi/builders/models.py
penfever: Yes

configs/recipes/gpt_oss/inference/120b_vllm_infer.yaml
wizeng23: The type gets passed into vLLM during initialization: https://github.com/oumi-ai/oumi/blob/408a1b654c6c81b5049aa62a8687896d49011a1e/src/oumi/inference/vllm_inference_engine.py#L178

I'm not certain what the behavior is if we pass bfloat16 in but request the model gets quantized, but if it seems to work on your end, I'm happy to keep this

configs/recipes/gpt_oss/sft/20b_lora_single_gpu_train.yaml
```diff
@@ -0,0 +1,88 @@
...
+#
+# Usage:
+#   oumi train -c oumi://configs/recipes/gpt_oss/sft/20b_lora_single_gpu_train.yaml
+#
+# Hardware Requirements:
+#   - Single GPU with 80GB+ VRAM (H100, A100 80GB)
+#   - Or 40GB GPU with gradient checkpointing + smaller batch
+#
+# See Also:
+#   - HF GPT-OSS recipes for single GPU training
```
wizeng23:
```suggestion
#   - HF GPT-OSS recipes for single GPU training: https://cookbook.openai.com/articles/gpt-oss/fine-tune-transfomers
```

PR #1924, opened 2025-08-11
Improve e2e eval tests

wizeng23:
# Description

- Replace eval task for multi-modal models from `mmmu_val` to `mmmu_val_computer_science` to be in line with `mmlu_college_computer_science` for non-MM models. This also reduces runtime for E2E tests.
- Delete redundant tests. We don't need to test Llama 1B/3B/8B all for eval if they're the same family.
- Add parameter to support using vLLM inference engine for eval
- Re-enable train_mm_llama3_2_vision_11b_full
- Reduce num steps to reduce runtime

Tested that all e2e tests pass

## Related issues

Towards OPE-1483

## Before submitting

### DIFF
--- a/configs/recipes/llama3_1/evaluation/70b_eval.yaml
+++ b/configs/recipes/llama3_1/evaluation/70b_eval.yaml
@@ -24,7 +24,7 @@ model:
   shard_for_eval: True
 
 generation:
-  batch_size: 3  # 4 will OOM on a Polaris node
+  batch_size: 3 # 4 will OOM on a Polaris node
 
 tasks:
   # For all available tasks, see https://oumi.ai/docs/en/latest/user_guides/evaluate/evaluate.html
--- a/tests/e2e/test_eval_e2e.py
+++ b/tests/e2e/test_eval_e2e.py
@@ -186,14 +186,14 @@ def _test_eval_impl(
                 / "135m"
                 / "eval.yaml"
             ),
-            num_samples=3,
+            num_samples=4,
         ),
         EvalTestConfig(
             test_name="eval_text_phi3_single_gpu",
             config_path=(
                 get_configs_dir() / "recipes" / "phi3" / "evaluation" / "eval.yaml"
             ),
-            num_samples=3,
+            num_samples=4,
             use_simple_oumi_evaluate_command=True,
         ),
         EvalTestConfig(
@@ -205,7 +205,7 @@ def _test_eval_impl(
                 / "evaluation"
                 / "8b_eval.yaml"
             ),
-            num_samples=3,
+            num_samples=4,
         ),
         EvalTestConfig(
             test_name="eval_text_llama31_8b_vllm_single_gpu",
@@ -216,7 +216,7 @@ def _test_eval_impl(
                 / "evaluation"
                 / "8b_eval.yaml"
             ),
-            num_samples=3,
+            num_samples=4,
             enable_vllm=True,
         ),
     ],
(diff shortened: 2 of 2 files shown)


taenin (approved):

shanghongsim:
Questions:
1. Why do we need to re-enable train_mm_llama3_2_vision_11b_full
> Replace eval task for multi-modal models from mmmu_val to mmmu_val_computer_science to be in line with mmlu_college_computer_science for non-MM models.
2. Is this line saying that mmlu_college_computer_science is the standard for non-multi modal models (I Am guessing that is what non-MM stands for?) and so we are using mmmu_val_computer_science to be in line?

configs/recipes/vision/llama3_2_vision/evaluation/11b_eval.yaml
```diff
@@ -25,71 +25,11 @@ generation:
...
 
 tasks:
+  # For all available tasks, see https://oumi.ai/docs/en/latest/user_guides/evaluate/evaluate.html
   - evaluation_backend: lm_harness
-    task_name: "mmmu_val"
-    # num_samples: 100  # Limit the number of eval samples.
+    task_name: mmmu_val_computer_science
     eval_kwargs:
       num_fewshot: 5
-  # To enable individual sub-tasks, uncomment the entries of interest below:
```
shanghongsim, line 33: Are these options deprecated hence removed? Or just removed for brevity? If its just for brevity, I think Its good to still keep them somewhere for reference. As someone new, I find these options helpful!

shanghongsim, line 33: I think its still good to keep a note somewhere that one can activate specific sub-tasks. I searched Oumi docs and can't find this! As someone new, I find these options helpful.

configs/recipes/vision/qwen2_vl_2b/evaluation/eval.yaml
```diff
@@ -26,72 +26,9 @@ generation:
 
 tasks:
   - evaluation_backend: lm_harness
-    task_name: mmmu_val
-    # num_samples: 100  # Limit the number of eval samples.
+    task_name: mmmu_val_computer_science
     eval_kwargs:
       num_fewshot: 5
 
-  # To enable individual sub-tasks, uncomment the entries of interest below:
```
shanghongsim, line 34: Same comment on documenting specific sub-tasks as above

configs/recipes/vision/llama3_2_vision/evaluation/11b_eval.yaml
wizeng23: I think what it means is that you can add additional tasks to eval on, not necessarily activating subtasks of a main task or anything like that. This should be evident given that tasks is a list field.

Was this what you meant, or were you thinking of something else? IMO if we want this documented, it should be done in our docs, instead of trying to document it in every single eval config.

wizeng23:
1. Why do we need to re-enable train_mm_llama3_2_vision_11b_full
- We disabled the test because a version update broke it; ideally we would fix it, but the error was cryptic and it seemed like a niche issue. I also suspected it was caused by a dependency as opposed to the Oumi codebase. Since upgrading our dependency versions again fixed it, I re-enabled it.

2. Is this line saying that mmlu_college_computer_science is the standard for non-multi modal models (I Am guessing that is what non-MM stands for?) and so we are using mmmu_val_computer_science to be in line?
- Correct. We often use placeholder tasks in our train/eval datasets, and the expectation is the user will update them to whatever they want to run on. This also makes our E2E tests faster since we only need to evaluate on one subset as opposed to 30.

PR #1929, opened 2025-08-14
Ryan arman add analysis summary

ryan-arman:
# Description
Adding computation of summary statistics for message level and conversation level
Also raises an exception when analysis results are not ready instead of silently calling analyze_dataset 

## Related issues

Fixes OPE-1496
Towards OPE-1370

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/analyze/dataset_analyzer.py
+++ b/src/oumi/core/analyze/dataset_analyzer.py
@@ -123,17 +123,6 @@ class DatasetAnalyzer:
         # Decimal precision for rounding metrics
         self._decimal_precision = 2
 
-    def set_decimal_precision(self, precision: int) -> None:
-        """Set the decimal precision for rounding metrics in analysis summary.
-
-        Args:
-            precision: Number of decimal places to round to (e.g., 2 for 0.12,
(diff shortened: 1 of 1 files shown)


src/oumi/core/analyze/dataset_analyzer.py
```diff
@@ -118,6 +118,21 @@ def __init__(self, config: AnalyzeConfig):
...
+    def set_decimal_precision(self, precision: int) -> None:
+        """Set the decimal precision for rounding metrics in analysis summary.
+
+        Args:
+            precision: Number of decimal places to round to (e.g., 2 for 0.12,
+                3 for 0.123)
+        """
+        if not isinstance(precision, int) or precision < 0:
+            raise ValueError("Decimal precision must be a non-negative integer")
+        self._decimal_precision = precision
```
jgreer013: Seems like this method isn't being used or tested anywhere, also any reason to make this a setter instead of just an argument to the analyzer?

ryan-arman: that is a good point. removed it

PR #1935, opened 2025-08-18
Add support for custom datasets in Analyze

ryan-arman:
# Description
Currently Analyze only supports oumi registered datasets. This change adds support for custom text and multi-modal datasets.

## Related issues

Fixes OPE-1502

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/analysis_utils.py
+++ b/src/oumi/utils/analysis_utils.py
@@ -181,9 +181,8 @@ def _load_custom_dataset_from_path(
         dataset = VLJsonlinesDataset(**dataset_kwargs)
         logger.info(f"Loaded vision-language dataset from: {dataset_path}")
         return dataset
-
-    # If explicitly forced to text, load as text-only
-    if config.is_multimodal is False:
+    elif config.is_multimodal is False:
+        # If explicitly forced to text, load as text-only
         dataset_kwargs = {
             "dataset_path": str(path),
             "format": dataset_format,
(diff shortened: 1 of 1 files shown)


src/oumi/utils/analysis_utils.py
```diff
@@ -119,3 +137,66 @@ def load_dataset_from_config(
...
+            "processor_kwargs": config.processor_kwargs,
+            "trust_remote_code": config.trust_remote_code,
+        }
+        dataset_kwargs = {k: v for k, v in dataset_kwargs.items() if v is not None}
+        dataset = VLJsonlinesDataset(**dataset_kwargs)
+        logger.info(f"Loaded vision-language dataset from: {dataset_path}")
+        return dataset
+
+    # If explicitly forced to text, load as text-only
+    if config.is_multimodal is False:
```
jgreer013: nit: else

src/oumi/core/configs/analyze_config.py
```diff
@@ -84,10 +94,47 @@ class AnalyzeConfig(BaseConfig):
     trust_remote_code: bool = False
     """Whether to trust remote code for processor loading."""
 
+    is_multimodal: Optional[bool] = None
```
jgreer013, line 97: Why is this optional vs just "False" by default?

src/oumi/utils/analysis_utils.py
```diff
@@ -119,3 +137,66 @@ def load_dataset_from_config(
...
+            "format": dataset_format,
+        }
+        if tokenizer is not None:
+            dataset_kwargs["tokenizer"] = tokenizer
+        dataset_kwargs = {k: v for k, v in dataset_kwargs.items() if v is not None}
+        dataset = TextSftJsonLinesDataset(**dataset_kwargs)
+        logger.info(f"Loaded text dataset from: {dataset_path}")
+        return dataset
+
+    # Note: is_multimodal=None case is already caught by AnalyzeConfig validation
```
jgreer013: I suspect if you make `is_multimodal` no longer optional then you'll be able to get rid of this section

src/oumi/core/configs/analyze_config.py
ryan-arman: This forces the user of this class to always set this. Otherwise they may pass a multi-modal dataset and forget that they should set this to True

src/oumi/utils/analysis_utils.py
ryan-arman: I think it is better to keep is_multimodal as None for force the user to set it. Also, it is good to still check in the config to validate that is_multimodal is not None for any reason

src/oumi/utils/analysis_utils.py
ryan-arman: good point, added the else branch (which should not happen)

PR #1936, opened 2025-08-18
Add Pre-Populated GitHub Issue Link On Failures

rlehman221:
# Description

We want to make users more aware that when they run into any types of issues they should open an issue. This PR appends a pre-populated Github issue link after any type of failure occurs for the user to optionally open if they feel it should be tracked by the team.

Example: `oumi launch up -c configs/examples/grpo_verl_countdown/gcp_job.yaml --cluster grpo-verl-countdown`
**BEFORE CHANGES**
image

**AFTER CHANGERS**
Screenshot 2025-08-19 at 10 23 32 AM
Screenshot 2025-08-20 at 9 22 05 AM

**LONGER ERROR EXAMPLE**
Screenshot 2025-08-20 at 9 21 05 AM
Screenshot 2025-08-20 at 9 21 27 AM

NOTE: For any jobs used with the launcher, rich console logs that leverage links won't work. Instead a check is made and if its a remote job we just display the template for the user to use instead of populating any fields

Screenshot 2025-08-20 at 11 15 19 AM

## Before submitting

### DIFF
--- a/src/oumi/cli/cli_utils.py
+++ b/src/oumi/cli/cli_utils.py
@@ -252,53 +252,67 @@ def create_github_issue_url(exception: Exception, traceback_str: str) -> str:
     Returns:
         str: URL for creating a prefilled GitHub issue
     """
+    # 2000 is the max length of a URL to ensure it works with any browser
+    MAX_URL_LENGTH = 2000
+    base_url = "https://github.com/oumi-ai/oumi/issues/new?"
+
     python_version = (
         f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
     )
     system_info = f"{platform.system()} {platform.release()}"
-
-    # Truncate long error messages and tracebacks to avoid URL length limits
-    # URL Length Limit: 2000 characters to be safe in all browsers
-    error_msg = (
-        str(exception)[:200] + "..." if len(str(exception)) > 200 else str(exception)
-    )
-    short_traceback = traceback_str.split("\n")[-10:]  # Last 10 lines only
-    short_traceback = (
-        "\n".join(short_traceback)[:800] + "..."
-        if len("\n".join(short_traceback)) > 800
-        else "\n".join(short_traceback)
-    )
-
-    title = f"[CLI Error]: {type(exception).__name__}"
-
-    command_str = " ".join(sys.argv)
-    command_display = f"`{command_str[:100]}{'...' if len(command_str) > 100 else ''}`"
-
-    what_happened = error_msg
-
-    reproduction_steps = (
-        f"Steps to reproduce:\n\n"
-        f"1. Run command: {command_display}\n\n"
-        f"Error message:\n{error_msg}\n\n"
-        f"Stack trace:\n```\n{short_traceback}\n```"
-    )
-
     oumi_version = importlib.metadata.version("oumi")
     system_info_content = (
-        f"Please paste the output of `oumi env` here. "
+        f"**Please paste the output of `oumi env` here.** "
         f"If you can't run this command, here's basic system info:\n\n"
         f"- Operating system: {system_info}\n"
         f"- Python version: {python_version}\n"
         f"- Oumi version: {oumi_version}"
     )
 
-    # Use the bug-report.yaml template
-    params = {
-        "template": "bug-report.yaml",
-        "title": title,
-        "what-happened": what_happened,
-        "reproduction-steps": reproduction_steps,
-        "system-info": system_info_content,
-    }
-    query = urllib.parse.urlencode(params)
-    return f"https://github.com/oumi-ai/oumi/issues/new?{query}"
+    title_template = "[Automatic Error Report]: {error_msg}"
+    reproduction_template = (
+        "Steps to reproduce:\n\n"
+        "Command executed: `{command}`\n\n"
+        "Stack trace:\n```\n{traceback}\n```"
+    )
+
+    exception_str = str(exception)
+    command_str = " ".join(sys.argv)
+
+    # Limit the title to 50 chars (unencoded)
+    title_error = exception_str[:50]
+
+    command_part = command_str
+    error_part = exception_str
+    stack_part = traceback_str or ""
+
+    def build_url(command_part: str, error_part: str, stack_part: str) -> str:
+        params = {
+            "template": "bug-report.yaml",
+            "title": title_template.format(error_msg=title_error),
+            "what-happened": error_part,
+            "reproduction-steps": reproduction_template.format(
+                command=command_part, traceback=stack_part
+            ),
+            "system-info": system_info_content,
+        }
+        return f"{base_url}{urllib.parse.urlencode(params)}"
+
+    full_url = build_url(command_part, error_part, stack_part)
+
+    # If too long, iteratively shrink fields by encoded length
+    while len(full_url) > MAX_URL_LENGTH:
+        if stack_part:
+            # shrink stack from the front (keep tail)
+            cut = max(1, len(stack_part) // 2)
+            stack_part = stack_part[cut:]
+        elif error_part:
+            cut = max(1, len(error_part) // 2)
+            error_part = error_part[:-cut]
+        elif command_part:
+            cut = max(1, len(command_part) // 2)
+            command_part = command_part[:-cut]
+        else:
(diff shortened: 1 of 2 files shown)


src/oumi/cli/main.py
```diff
@@ -156,7 +161,26 @@ def get_app() -> typer.Typer:
...
+            f"📝 [yellow]Prefilled issue:[/yellow] "
+            f"[link={issue_url}]Click here to report[/link]"
+        )
+        CONSOLE.print(
+            "🔗 [dim]Or create your own:[/dim] "
+            "[link=https://github.com/oumi-ai/oumi/issues/new]"
+            "github.com/oumi-ai/oumi/issues/new[/link]"
+        )
+
+        sys.exit(1)
```
wizeng23, line 188: To show reviewers how it would look like and to provide proof of e2e testing, could you include in your PR description an example output before and after this PR? It would be great to also test errors that are generated by Oumi (ex. we raise a ValueError) vs those raised from a dependency (ex. a pytorch CUDA crash). For example, you can try running `oumi launch up -c configs/examples/grpo_verl_countdown/gcp_job.yaml --cluster grpo-verl-countdown`, which should be broken right now

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +238,50 @@ def resolve_and_fetch_config(
         raise
 
     return Path(local_path)
+
+
+def create_github_issue_url(exception: Exception, traceback_str: str) -> str:
```
wizeng23, line 245: Could you also provide an example of a issue that gets generated? How aligned is it with our template: https://github.com/oumi-ai/oumi/issues/new?template=bug-report.yaml

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +238,50 @@ def resolve_and_fetch_config(
...
+
+    Returns:
+        str: URL for creating a prefilled GitHub issue
+    """
+    python_version = (
+        f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
+    )
+    system_info = f"{platform.system()} {platform.release()}"
+
+    # Truncate long error messages and tracebacks to avoid URL length limits
```
wizeng23: nit: mention here what the url length limit is?

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +238,50 @@ def resolve_and_fetch_config(
...
+
+    title = f"CLI Error: {type(exception).__name__}"
+
+    command_str = " ".join(sys.argv)
+    command_display = f"`{command_str[:100]}{'...' if len(command_str) > 100 else ''}`"
+
+    body = f"""**Error**: {error_msg}
+**Python**: {python_version}
+**System**: {system_info}
+**Command**: {command_display}git
```
wizeng23: Is "git" meant to be there?

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +238,50 @@ def resolve_and_fetch_config(
...
+        exception: The exception that occurred
+        traceback_str: The full traceback string
+
+    Returns:
+        str: URL for creating a prefilled GitHub issue
+    """
+    python_version = (
+        f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
+    )
+    system_info = f"{platform.system()} {platform.release()}"
```
shanghongsim, line 262: Can we also capture package versions like those we get from from `oumi env`? Some errors are due to differences in fast attention version, pytorch version mismatch with CUDA etc.

rlehman221: Great point! I've updated to include minor details to capture but also let the user know to still run oumi env. The reason we are limited is because we only have a 2,000 character limit for url query length and the more we provide on env the less we can provide on say the stack trace

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
```
wizeng23, line 262: nit: group these lower down with system_info_content

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
+    short_traceback = (
+        "\n".join(short_traceback)[:800] + "..."
+        if len("\n".join(short_traceback)) > 800
+        else "\n".join(short_traceback)
+    )
+
+    title = f"[CLI Error]: {type(exception).__name__}"
+
+    command_str = " ".join(sys.argv)
+    command_display = f"`{command_str[:100]}{'...' if len(command_str) > 100 else ''}`"
```
wizeng23: nit: consider not limiting this at all. Knowing the exact command is IMO one of the most important pieces of information

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
+
+    what_happened = error_msg
+
+    reproduction_steps = (
+        f"Steps to reproduce:\n\n"
+        f"1. Run command: {command_display}\n\n"
+        f"Error message:\n{error_msg}\n\n"
```
wizeng23: nit: consider removing this to save on characters since this is duplicated

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
+        f"Stack trace:\n```\n{short_traceback}\n```"
+    )
+
+    oumi_version = importlib.metadata.version("oumi")
+    system_info_content = (
+        f"Please paste the output of `oumi env` here. "
```
wizeng23: nit: consider adding some sort of formatting or emphasis to this line to this to get the user to notice this. Otherwise the user may submit without doing this if they are moving fast

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
+        str: URL for creating a prefilled GitHub issue
+    """
+    python_version = (
+        f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
+    )
+    system_info = f"{platform.system()} {platform.release()}"
+
+    # Truncate long error messages and tracebacks to avoid URL length limits
+    # URL Length Limit: 2000 characters to be safe in all browsers
+    error_msg = (
```
wizeng23: nit: if we know the absolute url length limit is 2k, we can use a counter to maximize the amount of text we show in each section without going over the limit. For example:
```
num_chars = 1900 # should be less than 2k to account for chars we have by default
command_display = ...
num_chars -= len(command_display)
error_msg = ...
num_chars -= len(error_msg)
# you can then include as much of the traceback as you can fit in the remaining number of characters, not just limited to 10 lines.
```

src/oumi/cli/cli_utils.py
```diff
@@ -235,3 +240,65 @@ def resolve_and_fetch_config(
...
+        str(exception)[:200] + "..." if len(str(exception)) > 200 else str(exception)
+    )
+    short_traceback = traceback_str.split("\n")[-10:]  # Last 10 lines only
+    short_traceback = (
+        "\n".join(short_traceback)[:800] + "..."
+        if len("\n".join(short_traceback)) > 800
+        else "\n".join(short_traceback)
+    )
+
+    title = f"[CLI Error]: {type(exception).__name__}"
```
wizeng23: nit: This title isn't too informative on its own, and at a glance in the GitHub UI, we'll get a lot of bugs with duplicate titles. Consider putting the error message here instead, and maybe a placeholder message in the what happened (ex. "Automatic error report"). Other ideas welcome as well!

src/oumi/cli/cli_utils.py
wizeng23, line 1: Could you fix the error in your PR description (you need to run `wandb login`) and run again? I'd like to see an example of what it looks like with a more complex bug that has a very large traceback. If you could also copy the created github issue link to the description as well, that'd be great, thanks!

rlehman221: What do we use wandb for?

wizeng23: For experiment tracking. It's like cloud Tensorboard

PR #1938, opened 2025-08-19
Judge | Validate Input Dataset

kaisopos:
# Description

Judge: Validate the input dataset before judgments, to fast-fail (if required keys are missing) without executing inference.

## Related issues

Towards OPE-1327

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/judges/base_judge.py
+++ b/src/oumi/judges/base_judge.py
@@ -391,7 +391,7 @@ class BaseJudge:
     ) -> bool:
         """Validate that all inputs contain the required placeholder keys."""
         if self.prompt_template_placeholders is None:
-            return True  # No validation needed if placeholders aren't specified
+            return True  # No validation needed if no placeholders are specified
 
         for index, input in enumerate(inputs):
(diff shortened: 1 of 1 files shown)


src/oumi/core/configs/params/judge_params.py
```diff
@@ -83,6 +83,16 @@ class JudgeParams(BaseParams):
     prompt_template: str
     """Template for the judge prompt with placeholders, such as {question}, {answer}."""
 
+    prompt_template_placeholders: Optional[list[str]] = field(default=None)
```
min-oumi, line 86: can we just use set instead of list, similar to get_placeholders, or is order important here?

kaisopos: Nope, the order is NOT important.

Ideally this should be a set, but defining a set in yaml is super ugly and unintuitive:
```
my_set: !!set
  apple: null
  banana: null
  cherry: null
```
  
While a list is much cleaner:
```
fruits:
  - apple
  - banana
  - cherry
```  

Thus I opted to read a list from yaml and convert it to a set inside the judge.

PR #1940, opened 2025-08-20
Adding Verbose Flag

rlehman221:
# Description
By default, most configurations are displayed in the terminal during operations, which can create unnecessary noise for the average user. This PR introduces a --verbose flag, allowing users who want more detailed output to enable it, while keeping the default experience cleaner and easier to read.

# Usage

`oumi infer -i -c configs/recipes/smollm/inference/135m_infer.yaml **--verbose**`

# Tested With Flag On and Off

`oumi train -c configs/recipes/smollm/sft/135m/quickstart_train.yaml`
`oumi infer -i -c configs/recipes/smollm/inference/135m_infer.yaml`
`oumi evaluate -c configs/recipes/phi3/evaluation/eval.yaml`

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/cli/synth.py
+++ b/src/oumi/cli/synth.py
@@ -37,6 +37,7 @@ def synth(
         ),
     ],
     level: cli_utils.LOG_LEVEL_TYPE = None,
+    verbose: cli_utils.VERBOSE_TYPE = False,
 ):
     """Synthesize a dataset.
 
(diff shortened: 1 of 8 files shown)


src/oumi/cli/synth.py
```diff
@@ -37,13 +37,15 @@ def synth(
         ),
     ],
     level: cli_utils.LOG_LEVEL_TYPE = None,
+    verbose: cli_utils.VERBOSE_TYPE = False,
```
wizeng23: Could we remove it from synth if it's not used there yet? Alternatively (less preferred), we should be consistent and have it in all oumi verbs (judge, launch, etc.)

wizeng23:
Also, could you link to this issue in your PR description for tracking? https://linear.app/oumi/issue/OPE-782

Ex: https://github.com/oumi-ai/oumi/pull/1915

PR #1678, opened 2025-05-09
add debug logging capabilities to collators

aniruddh-alt:
# Description

This PR adds a comprehensive debug-logging feature to Oumi’s data collators, giving developers clear, step-by-step visibility into how raw text becomes model-ready tokens. By setting `debug=True`, you’ll get both console output and an HTML report (organized by session) that shows:

- **Raw Examples**: Original dataset records before any processing  
- **Formatted Text**: After applying prompts or templates  
- **Token Breakdown**: Each token’s ID, text, and position  
- **Collated Inputs**: Final batched tensors passed to the model  

## Key Changes

1. **Debug Flag**  
   - Added a `debug: bool` parameter to all collator builders and classes.  

2. **`debug_utils.py`**  
   - New utility module with reusable logging functions.  
   - `log_example_for_debugging()` records:
     - **Raw example** (pre-tokenization)
     - **Post-format example** (after any prompt/template)
     - **Token list** (IDs, decoded strings, special tokens)
     - **Final model inputs** (padded & batched tensors)

3. **Console & HTML Output**  
   - **Console**: Uses the existing `logger.debug()` interface.  
   - **HTML Reports**:
     - Written to `debug_logs/<session_id>/` with a timestamp.
     - A `latest.html` symlink always points to the newest report.
     - Clean, sectioned layouts with preformatted JSON for easy browser inspection.

4. **Updated Collators**  
   - `TextCollatorWithPadding`  
   - `TextCompletionsCollatorWithPadding`  
   Both now accept `debug=True` and integrate with the new utilities.

## Usage

```python
from oumi.data_collators import TextCollatorWithPadding

collator = TextCollatorWithPadding(
    tokenizer=my_tokenizer,
    max_length=512,
    debug=True  # ← enables detailed logging & HTML report
)
```

## Related issues

Fixes #1369

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [X] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [X] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

## Reviewers

At least one review from a member of `oumi-ai/oumi-staff` is required.

Issue #1369, opened 2025-02-04
[Feature] Output tokenized example for debugging during training


jgreer013:
### Feature request

When running training, it'd be useful to log a single example from the dataset that has been tokenized for debugging.

For example:
1. Take the first example or batch in the training set (peek the top of the dataloader?)
2. Print all the following:
a. Raw Example
b. Formatted Example
c. Tokenized Example (list of tuples of <token_id, decoded_token>)
d. Model Input (tokens with attention mask and labels)

Ideally this should either be a debug log or its output explicitly controlled by a flag.

Likely, all parts except the model input can be done when loading the dataset. The model input will need to be done inside the collator(s).

### Motivation / references

Concrete example using `HuggingFaceTB/SmolLM2-1.7B-Instruct` tokenizer, with model setup to be trained on **completions only**:

Raw Example
```
[
  {'role': 'system', 'content': 'This is a system instruction.'},
  {'role': 'user', 'content': 'This is a user instruction'},
  {'role': 'assistant', 'content': 'This is an assistant response.'},
]
```

Formatted Example
```
<|im_start|>system
This is a system instruction.<|im_end|>
<|im_start|>user
This is a user instruction<|im_end|>
<|im_start|>assistant
This is an assistant response.<|im_end|>
```

Tokenized Example
```
[(1, '<|im_start|>'), (9690, 'system'), (198, '\n'), (1348, 'This'), (314, ' is'), (253, ' a'), (817, ' system'), (5785, ' instruction'), (30, '.'), (2, '<|im_end|>'), (198, '\n'), (1, '<|im_start|>'), (4093, 'user'), (198, '\n'), (1348, 'This'), (314, ' is'), (253, ' a'), (2914, ' user'), (5785, ' instruction'), (2, '<|im_end|>'), (198, '\n'), (1, '<|im_start|>'), (520, 'ass'), (9531, 'istant'), (198, '\n'), (1348, 'This'), (314, ' is'), (354, ' an'), (11173, ' assistant'), (2426, ' response'), (30, '.'), (2, '<|im_end|>'), (198, '\n')]
```

Model Input Example
```
{
  'input': [1, 9690, 198, 1348, 314, 253, 817, 5785, 30, 2, 198, 1, 4093, 198, 1348, 314, 253, 2914, 5785, 2, 198, 1, 520, 9531, 198, 1348, 314, 354, 11173, 2426, 30, 2, 198],
[...]

PR #1943, opened 2025-08-21
Enable users to log data samples during training for debugging

shanghongsim:
# Description

- #1678 implements the functionality of logging of data samples for debugging 
- This PR extends functionality for `VisionLanguageCollatorWithPadding` and hooks it up to config so that users can set from yaml level
- Towards #1369 

## Related issues

Towards #1369 

## Before submitting

## Reviewers


tests/unit/test_log_example.py
```diff
@@ -0,0 +1,98 @@
+from unittest.mock import Mock, patch
```
taenin: Our convention is to name test files after the file they're testing. In this case, I would name your file `test_train.py`

src/oumi/core/configs/params/training_params.py
```diff
@@ -357,6 +357,12 @@ class TrainingParams(BaseParams):
     Possible values are "debug", "info", "warning", "error", "critical".
     """
 
+    log_examples: bool = False
+    """Whether to log an example of the data in each step for debugging purposes.
```
wizeng23: The previous PR you linked seems to only log one debug example (in the first step of training). IMO this makes more sense than doing it each step, as that may clutter the training logs, and I think just one example may be enough to understand how the input is processed in training. If that's true, could you update the description here?

wizeng23 (commented):
1. In your PR description, could you include an example for what the output of our training job looks like with this flag enabled? This helps me as a reviewer since a) it's an E2E test for your PR aside from just unit tests and b) lets me see what the end change looks like to better understand the PR.
2. Can you link to #1369 in your PR description? This helps track the PR in the future.

src/oumi/core/configs/params/training_params.py
shanghongsim: Sure, here is a snapshot of one debug sample from the logs. Currently, its is printing 256 of such lines, likely because it is spinning up 256 data collators (dataloader_num_workers=64 and I have 4 different GPUs), creating quite a bit of clutter.  I have documented this in [OPE-1511](https://linear.app/oumi/issue/OPE-1511) and we should definitely improve it in a follow up PR.
image

wizeng23 (approved):

taenin (approved):

PR #1947, opened 2025-08-22
Penfever/api and macos configs

penfever:
# Description

Adds config files for newer API and local models (GPT 5, Qwen 3, Claude Opus 4.1, Gemma 3)

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
+++ b/configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
@@ -1,11 +1,11 @@
 # LlamaCPP Inference config for DeepSeek R1 Distill Qwen 32B GGUF (macOS compatible).
 #
 # Requirements:
-#   - Run `pip install oumi`
+#   - Run `pip install oumi[llama_cpp]`
 #
 # Usage:
 #   # 1. Download the GGUF file to current working directory:
-#   huggingface-cli download unsloth/DeepSeek-R1-Distill-Qwen-32B-GGUF \
+#   hf download unsloth/DeepSeek-R1-Distill-Qwen-32B-GGUF \
 #     DeepSeek-R1-Distill-Qwen-32B-Q4_K_M.gguf --local-dir .
 #
 #   # 2. Run inference:
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@ notebooks/*tutorial*
 # Evaluation results
 results/
 
+# Pyright Config
+pyrightconfig.json
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/configs/recipes/gemma3/inference/3n_e4b_it_gguf_macos_infer.yaml
+++ b/configs/recipes/gemma3/inference/3n_e4b_it_gguf_macos_infer.yaml
@@ -1,11 +1,11 @@
 # LlamaCPP Inference config for Gemma-3n-E4B-it GGUF (macOS compatible).
 #
 # Requirements:
-#   - Run `pip install oumi`
+#   - Run `pip install oumi[llama_cpp]`
 #
 # Usage:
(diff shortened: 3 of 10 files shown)


configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
```diff
@@ -0,0 +1,34 @@
+# LlamaCPP Inference config for DeepSeek R1 Distill Qwen 32B GGUF (macOS compatible).
+#
+# Requirements:
+#   - Run `pip install oumi`
```
wizeng23: llama_cpp as well

configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
```diff
@@ -0,0 +1,34 @@
...
+#
+# Usage:
+#   # 1. Download the GGUF file to current working directory:
+#   huggingface-cli download unsloth/DeepSeek-R1-Distill-Qwen-32B-GGUF \
```
wizeng23: 1. Mind replacing `huggingface-cli download` with `hf download` throughout? The former is deprecated
2. What's the reason for downloading it to the current dir, instead of saving it to the default HF cache location?

configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_infer.yaml
```diff
@@ -0,0 +1,30 @@
+# vLLM Inference config for DeepSeek R1 Distill Qwen 32B GGUF.
+#
+# Requirements:
+#   - Run `pip install oumi[gpu]`
+#
+# Usage:
+#   oumi infer -i -c configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_infer.yaml
+#
```
wizeng23, line 8: Does this need to download the GGUF model file locally as well, similar to the mac config?

wizeng23 (commented):
Could you also add these to src/oumi/cli/alias.py?

configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_infer.yaml
penfever: Manual downloading is not required, I believe -- the vLLM engine in Oumi should automatically download GGUF files.

configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
penfever: Great point!

penfever: Let me update the pyproject.yaml to include an optional dependency here

configs/recipes/deepseek_r1/inference/distill_qwen_32b_gguf_macos_infer.yaml
penfever: 1. Sure!
2. Because the default HF cache location is sometimes not writable in HPC settings, and users may not know where it is or how to find it if they want to modify configs to point to other quants. It just saves a mental step for people, I think.

wizeng23: Fair point, though there's actually a pretty big scope for this given that almost all of our configs will be using the default HF cache location; we could add documentation about considering to update the HF cache location, but existing users of HF transformers on HPCs may already know about this.

For this PR though, my small nits are:
1. The model's being downloaded to the current dir only for MacOS configs, where there's not a worry about writability.
2. Could we download it to a working dir that's covered by our gitignore (ex. tmp/)? I'd be ok with models/ as well, but I removed it for now because the regex was causing bugs with the job launcher: #1944

PR #1950, opened 2025-08-22
Add a parsable enum state field to the launcher to convey job status

taenin:
# Description

Previously the oumi launcher only exposed a `done` enum, as well as a string for status parsing. This PR adds an enum `state` to easily check if a job is `pending`, `failed`, `succeeded`, or `cancelled`.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/launcher/base_cluster.py
+++ b/src/oumi/core/launcher/base_cluster.py
@@ -23,6 +23,7 @@ class JobState(Enum):
     """Enum to hold the state of a job."""
 
     PENDING = "pending"
+    RUNNING = "running"
     SUCCEEDED = "succeeded"
     FAILED = "failed"
     CANCELLED = "cancelled"
--- a/src/oumi/core/launcher/base_cluster.py
+++ b/src/oumi/core/launcher/base_cluster.py
@@ -14,8 +14,18 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from enum import Enum
 
 from oumi.core.configs import JobConfig
+
+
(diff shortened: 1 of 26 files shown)


src/oumi/launcher/clients/local_client.py
```diff
@@ -64,11 +64,26 @@ def __init__(self):
         self._worker = Thread(target=self._worker_loop, daemon=True)
         self._worker.start()
 
+    def _get_job_state(self, job_state: _JobState) -> JobState:
```
wizeng23, line 67: Could _JobState be renamed to _JobStatus since that's how the field seems to be directly used?

src/oumi/core/launcher/base_cluster.py
```diff
@@ -14,10 +14,20 @@
...
 from dataclasses import dataclass
+from enum import Enum
 
 from oumi.core.configs import JobConfig
 
 
+class JobState(Enum):
+    """Enum to hold the state of a job."""
+
+    PENDING = "pending"
```
wizeng23, line 25: Thoughts on splitting the pending state into queued and running?

src/oumi/launcher/clients/local_client.py
taenin: I'm going to leave this one as is for the sake of time as the local_client is more for demo purposes than actual use.

src/oumi/core/launcher/base_cluster.py
taenin: Added

PR #1951, opened 2025-08-25
Add The Ability To Trail Logs For Launcher Jobs

rlehman221:
# Description

We currently have built-in support from SkyPilot for trailing logs through their SDK after launching a job. We want to extend this functionality to our other clusters as well.

This PR adds support for Slurm clusters by tailing the log file and streaming its contents directly to the user’s terminal. For clusters not covered in this PR, the behavior will remain unchanged and continue to show the loading spinner.

Features Added:
- Tailing logs is now supported in Slurm as an IOStream to our rich console
- Tailing logs is Sky Pilot is now returning an IOStream to our rich console
- Support for passing in an IO file for the logs to be streamed to works with the parameter "--output-filepath"
- Sky Pilot upgraded and patched to support streaming to a file

**BEFORE With Slurm**
Screenshot 2025-08-25 at 1 37 31 PM

**AFTER With Slurm**
image

**AFTER With Sky Pilot**
Screenshot 2025-08-26 at 11 08 55 AM

**After With Local Cluster**
Screenshot 2025-08-26 at 11 16 46 AM

## Related issues

Towards OPE-1344

## Before submitting

## Reviewers


wizeng23:
Do you mind updating `Fixes OPE-1344` in your PR description to `Towards OPE-1344`? We should mark the issue as fixed only when all clusters support this functionality.

src/oumi/cli/launch.py
```diff
@@ -210,18 +209,11 @@ def _poll_job(
 
     assert running_cluster
 
-    # Check if this is a Skypilot job and tail logs automatically
-    if cloud in [cloud.value for cloud in sky_client.SkyClient.SupportedClouds]:
+    if callable(getattr(running_cluster, "tail_logs", None)):
```
wizeng23: Since tail_logs seems to be implemented for most clusters, won't this always pass? For example, for Frontier, the tail_logs function will still be called and then an error will be thrown.

rlehman221: Agreed! I was testing out this change locally but it should be a try / except at that point

src/oumi/core/launcher/base_cluster.py
```diff
@@ -80,3 +80,13 @@ def stop(self) -> None:
...
         """Tears down the current cluster."""
         raise NotImplementedError
+
+    @abstractmethod
+    def tail_logs(self, job_id: str, cluster_name: str) -> None:
+        """Tails the logs of the target job.
+
+        Args:
+            job_id: The ID of the job to tail the logs of.
+            cluster_name: The name of the cluster to tail the logs of.
```
wizeng23:
```suggestion
            cluster_name: The name of the cluster the job was run in.
```
Ditto for other files

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -154,10 +154,19 @@ def name(self):
             """Gets the name of the connection in the form user@hostname."""
             return f"{self.user}@{self.hostname}"
 
+    @dataclass
+    class JobInfo:
+        """Information about a submitted job."""
+
+        working_dir: Path
```
wizeng23: Can you add comments to describe the fields?

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -154,10 +154,19 @@ def name(self):
...
+        stdout_file: str
```
wizeng23: nit: stdout_filename. The convention I usually see is that "file" refers to an open file, "filename" refers just to the name, and "filepath" refers to the full path including the filename.

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -268,6 +277,12 @@ def run_job(self, job: JobConfig) -> JobStatus:
             job.num_nodes,
             job_name,
         )
+        # By default, Slurm writes to slurm-<job_id>.out.
+        self.jobs_info[job_id] = SlurmCluster.JobInfo(
```
wizeng23, line 280: Could you also make sure to add logic to remove entries from this dict?

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -288,3 +303,20 @@ def stop(self) -> None:
...
+        """Tails the logs of the target job.
+
+        Args:
+            job_id: The ID of the job to tail the logs of.
+            cluster_name: The name of the cluster to tail the logs of.
+        """
+        if job_id not in self.jobs_info:
+            raise ValueError(f"Job {job_id} not found in jobs_info")
+        job_info = self.jobs_info[job_id]
+        self._client.tail_job(
```
wizeng23: Will this function return and the logs tailing stop when the job is done, or will it hang? If not, could we add logic to implement this? If this is non-trivial to add, it can also be left for a follow-up PR.

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -268,6 +277,12 @@ def run_job(self, job: JobConfig) -> JobStatus:
             job.num_nodes,
             job_name,
         )
+        # By default, Slurm writes to slurm-<job_id>.out.
+        self.jobs_info[job_id] = SlurmCluster.JobInfo(
+            working_dir=remote_working_dir,
+            stdout_file=f"slurm-{job_id}.out",
+            stderr_file=f"slurm-{job_id}.out",
```
wizeng23: Do we need this field if it's unused and has the same value as stdout?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -494,3 +497,90 @@ def put(self, file_contents: str, destination: str) -> None:
...
+            The SSH command's exit code. This is a blocking call.
+        """
+        tail_cmd = (
+            f"ssh {_CTRL_PATH} -tt {cluster_name} "
+            f'"cd {working_dir} && tail -n +1 -F {stdout_file}"'
+        )
+        cli_utils.CONSOLE.print(
+            f"Tailing logs of job {job_id} on cluster '{cluster_name}'..."
+        )
+        cli_utils.CONSOLE.print(f"└── Following output file: {stdout_file}")
```
wizeng23: nit: consider printing working_dir / stdout_file so users know exactly where the file is

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -494,3 +497,90 @@ def put(self, file_contents: str, destination: str) -> None:
...
+        cli_utils.CONSOLE.print(
+            "Press Ctrl-C to exit log streaming; job will not be killed."
+        )
+
+        max_attempts = 3
+        base_delay = 0.5
```
wizeng23: nit: IMO since we're supposed to be polling at this point in the launch script anyway (since this logs tailing logic was just replacing generic polling), I think it's fine to poll longer. 3.5 seconds total as it is now may not be enough time for Slurm to set the job too.

How about base 5, max 20, 4 attempts (about a minute total)?

src/oumi/cli/launch.py
```diff
@@ -210,19 +209,12 @@ def _poll_job(
...
-        # Delay sky import: https://github.com/oumi-ai/oumi/issues/1605
-        import sky
-
-        sky.tail_logs(
-            cluster_name=job_status.cluster,
-            job_id=job_status.id,
-        )
-    else:
+        running_cluster.tail_logs(job_status.id, job_status.cluster)
+    except NotImplementedError:
```
wizeng23, line 266: IMO you could have a broad except here for any type of error, as we can just fall back to the default behavior in case of any issue with tailing logs. Right now, I think our launch script will crash if we encounter the other error types you added to this PR, which isn't desirable

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -288,3 +303,20 @@ def stop(self) -> None:
...
+
+    def tail_logs(self, job_id: str, cluster_name: str) -> None:
+        """Tails the logs of the target job.
+
+        Args:
+            job_id: The ID of the job to tail the logs of.
+            cluster_name: The name of the cluster to tail the logs of.
+        """
+        if job_id not in self.jobs_info:
+            raise ValueError(f"Job {job_id} not found in jobs_info")
```
wizeng23: super nit: AFAIK ValueError is when an argument has a wrong value (especially user input arguments). Consider using a catch-all generic error like RuntimeError instead

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -494,3 +497,90 @@ def put(self, file_contents: str, destination: str) -> None:
...
+        Args:
+            working_dir: Remote working directory where the job was submitted.
+            job_id: The Slurm job ID whose output file to follow.
+            cluster_name: The name of the cluster to tail the logs of.
+            stdout_file: The name of the stdout file to tail.
+
+        Returns:
+            The SSH command's exit code. This is a blocking call.
+        """
+        tail_cmd = (
```
wizeng23: super nit: consider moving this down to line 558, to group it closer to where it's used.

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -494,3 +497,90 @@ def put(self, file_contents: str, destination: str) -> None:
...
+        except KeyboardInterrupt:
+            logger.info("Stopped tailing logs for job %s", job_id)
+            if proc and proc.poll() is None:
+                os.killpg(proc.pid, signal.SIGINT)
+                proc.wait(timeout=5)
+            return 130  # Exit code for Ctrl-C
+        except Exception:
+            logger.exception("Failed while tailing logs for job %s", job_id)
+            if proc and proc.poll() is None:
+                os.killpg(proc.pid, signal.SIGTERM)
```
wizeng23: Just to confirm, is there no need for a proc.wait call here?

wizeng23 (commented):
For testing, could you try running the slurm, local, and sky clusters? Example local cluster at "notebooks/Oumi - A Tour.ipynb". For these tests, it would be good to confirm that the logs display properly (ex. for Slurm, logs that are colored text should still have color)

src/oumi/launcher/clusters/slurm_cluster.py
rlehman221: makes a lot of sense! I've seen a lot of other scenarios in the code where after you mentioned it I noticed we aren't following this standard. As I keep pushing other PR's I will fix the rest of the code as I find them

src/oumi/launcher/clusters/slurm_cluster.py
rlehman221: Yup! So inside the _client tail job there is logic to continue to poll the job and once done close the tailing ssh terminal

src/oumi/launcher/clients/slurm_client.py
rlehman221: that's actually a really good, point I had this in to ensure the zombie process if any would be cleaned up but regardless we send the SIGTERM already so the user on the CLI doesn't need to wait longer

src/oumi/launcher/clusters/slurm_cluster.py
rlehman221: I've kept it since we aren't using it for tailing but the rest of the code to leverage it in slurm still allows it to be passed. If we ever need it in the future we can just add it back in then.

src/oumi/cli/launch.py
```diff
@@ -210,19 +209,12 @@ def _poll_job(
 
     assert running_cluster
 
-    # Check if this is a Skypilot job and tail logs automatically
-    if cloud in [cloud.value for cloud in sky_client.SkyClient.SupportedClouds]:
+    try:
         cli_utils.CONSOLE.print(
             f"Tailing logs for job [yellow]{job_status.id}[/yellow]..."
```
wizeng23: nit: what about removing this line?
1. It seems you have a similar log line after this for slurm/sky
2. It might be less confusing if there's an exception, as otherwise we'd log that we're tailing and then not do it

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -518,3 +521,91 @@ def put(self, file_contents: str, destination: str) -> None:
...
+            stdout_filename: The name of the stdout file to tail.
+
+        Returns:
+            The SSH command's exit code. This is a blocking call.
+        """
+        cli_utils.CONSOLE.print(
+            f"Tailing logs of job {job_id} on cluster '{cluster_name}'..."
+        )
+        cli_utils.CONSOLE.print(
+            f"└── Following output file: {working_dir}/{stdout_filename}"
```
wizeng23: nit: consider using os.path.join or `{Path(working_dir) / stdout_filename}` in case working dir has a trailing /

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -518,3 +521,91 @@ def put(self, file_contents: str, destination: str) -> None:
...
+                if preflight.exit_code == 0:
+                    break
+
+                if attempt < max_attempts - 1:
+                    delay = min(base_delay * (2**attempt), max_delay)
+                    time.sleep(delay)
+                else:
+                    raise FileNotFoundError(
+                        f"Log file not found after {max_attempts} attempts: "
+                        f"{working_dir}/{stdout_filename}. "
```
wizeng23: ditto

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -288,3 +301,21 @@ def stop(self) -> None:
...
+        if job_id not in self.jobs_info:
+            raise RuntimeError(f"Job {job_id} not found in jobs_info")
+        job_info = self.jobs_info[job_id]
+        self._client.tail_job(
+            str(job_info.working_dir),
+            job_id,
+            cluster_name,
+            job_info.stdout_filename,
+        )
+        self.jobs_info.pop(job_id, None)
```
wizeng23: IMO it's better not to pop it here, in case users want to view the job logs afterwards as well. I've definitely had cases where I viewed the log file multiple times. Since it's a pretty small dict, IMO it's fine for now to not do any cleanup for jobs_info.

shanghongsim: Quick question (sorry if this is a silly question, I am still ramping up on working with cloud), wouldn't the logs be saved in a file which will be persistent? So clearing this in memory log should not be an issue?

shanghongsim (approved):

src/oumi/launcher/clusters/slurm_cluster.py
wizeng23: Yes that's true that the file persists, but the launcher will lose knowledge about the file location

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -518,3 +521,91 @@ def put(self, file_contents: str, destination: str) -> None:
...
+        """
+        cli_utils.CONSOLE.print(
+            f"Tailing logs of job {job_id} on cluster '{cluster_name}'..."
+        )
+        cli_utils.CONSOLE.print(
+            f"└── Following output file: {Path(working_dir) / stdout_filename}"
+        )
+        cli_utils.CONSOLE.print(
+            "Press Ctrl-C to exit log streaming; job will not be killed."
+        )
```
taenin: Drive-by comment here: why are we importing the CLI outside of the CLI? You should instead use the default logger outside of the CLI

src/oumi/core/launcher/base_cluster.py
```diff
@@ -95,3 +96,13 @@ def stop(self) -> None:
     def down(self) -> None:
         """Tears down the current cluster."""
         raise NotImplementedError
+
+    @abstractmethod
+    def get_tailed_stream(self, job_id: str, cluster_name: str) -> io.TextIOBase:
```
wizeng23: nit: get_logs_stream?

src/oumi/launcher/clients/sky_client.py
```diff
@@ -184,32 +216,34 @@ def launch(
                 f"{sky_cloud._REPR} does not support stopping clusters. "
                 "Will not set autostop."
             )
-
-        job_id, resource_handle = self._sky_lib.launch(
+        job_id = self._sky_lib.launch(
             sky_task,
             cluster_name=cluster_name,
```
wizeng23, line 217: Why can't you detach from the run anymore? What if we pass in the detach option when running the Oumi job?

src/oumi/launcher/clients/sky_client.py
```diff
@@ -58,6 +63,33 @@ def _get_sky_storage_mounts_from_job(job: JobConfig) -> dict[str, "sky.data.Stor
     return sky_mounts
 
 
+class SkyLogStream(io.TextIOBase):
+    """Wraps a log iterator into a readline()-capable stream."""
+
+    def __init__(self, iterator: Iterator[Optional[str]]):
+        """Initializes a new instance of the SkyLogStream class."""
+        self.iterator = iterator
+        self.ansi_pattern = re.compile(r"\x1b\[[0-9;]*[a-zA-Z]")
```
wizeng23, line 74: nit: Could you add a comment for what this does? I generally prefer having comments to explain any non-trivial regexes and their purpose

src/oumi/cli/launch.py
```diff
@@ -101,6 +101,45 @@ def _cancel_worker(id: str, cloud: str, cluster: str) -> bool:
     return True  # Always return true to indicate that the task is done.
 
 
+def _tail_logs(log_stream, job_id: str, output_filepath: Optional[str] = None) -> None:
```
taenin: Can we add a type annotation for `log_stream` ?

taenin (approved):

src/oumi/cli/launch.py
wizeng23: Add type for log_stream

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+            remaining = self._proc.stdout.read()
+            if remaining:
+                self._buffer += remaining
+
+        self._process_ended = True
+
+        # Return the next line if available
+        return self._read_from_buffer()
+
+    def _read_from_buffer(self) -> Optional[str]:
```
wizeng23: Should the return type be str since it always returns one? I don't see None being returned anywhere

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+
+    def __enter__(self):
+        """Context manager entry."""
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Context manager exit."""
+        self.close()
+
+    def _handle_process_end(self):
```
wizeng23: nit: IMO it would make more sense to decouple the logic of handling process end and returning the next line in the buffer, as opposed to overloading this function to do both. Thoughts on dropping the return line, and just making that call after the call to this function? i.e.
```
        if self._job_done:
            self._handle_process_end()
            return self._read_from_buffer()
```

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+    for reading job logs.
+    """
+
+    def __init__(
+        self,
+        working_dir: str,
+        job_id: str,
+        cluster_name: str,
+        stdout_filename: str,
+        client,
```
wizeng23: Add type? You could also use the type as a string I believe, ex. `client: "SlurmClient",`

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+
+        return proc
+
+    def _start_job_checking(self, proc: subprocess.Popen):
+        """Start background thread to check job status."""
+
+        def check_job_status():
+            while not self._job_done:
+                try:
+                    if self._client:
```
wizeng23: Is this check needed? Based on the input argument, it can't be none

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+                        job = self._client.get_job(self.job_id)
+                        if job is None or job.done:
+                            self._job_done = True
+                            proc.terminate()
```
wizeng23: nit: would it make sense organizationally to move the proc terminate and wait calls to _handle_process_end?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
...
+
+        # Process is still running, try to read from stdout
+        if self._proc.stdout:
+            return self._proc.stdout.readline()
+        return ""
+
+    def close(self) -> None:
+        """Close the stream and clean up resources."""
+        if self._proc and self._proc.poll() is None:
+            os.killpg(self._proc.pid, signal.SIGINT)
```
wizeng23: Just in case, should we clean up the _job_check_thread here as well?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -177,6 +181,201 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+class SlurmLogStream(io.TextIOBase):
+    """A stream that provides access to job logs.
+
+    This class inherits from io.TextIOBase to provide a file-like interface
+    for reading job logs.
+    """
```
wizeng23, line 201: There's a lot of condition checks in this class (job done, proc isn't None, process ended, buffer is empty, etc.). In the class docstring, could you provide an overview of the expected order of events?

src/oumi/launcher/clients/slurm_client.py
wizeng23, line 1: I just tried testing this, and the last line of the log isn't printed out. When I run `cat slurm-1425.out`, I get:

```
...
 Downloading skypilot
Prepared 10 packages in 2.35s
error: failed to remove directory `/home/wizeng/miniconda3/envs/oumi/lib/python3.11/site-packages/torch/__pycache__`: Directory not empty (os error 39)
```

But the end of the tailed logs is:

```
 Downloading skypilot
Prepared 10 packages in 2.35s
Job 1425 finished with status FAILED
Job metadata:
JobID                                                 JobName                           User                          State                         Reason
------------------------------ ------------------------------ ------------------------------ ------------------------------ ------------------------------
1425                                          smollm-135m-sft                         wizeng                         FAILED                           None
```

wizeng23, line 1: One thing that doesn't work well is when a line is modified in the slurm logs. For example, with progress bars, we'll get tailed logs that look like:
```
 33%|███▎      | 1056/3235 [04:33<08:45,  4.15it/s]
 33%|███▎      | 1057/3235 [04:33<08:41,  4.18it/s]
 33%|███▎      | 1058/3235 [04:33<08:40,  4.18it/s]
```

While in the slurm log text file it's just one line that repeatedly gets updated.
If this is non-trivial to fix, consider leaving a TODO and logging an issue.

wizeng23: Same thing happens for Skypilot, which is a regression since it used to handle progress bars fine when tailing. Example:

```
(smollm-135m-sft, pid=4300) (smollm-135m-sft, pid=4300)   0%|          | 0/3235 [00:00<?, ?it/s](smollm-135m-sft, pid=4300)   0%|          | 1/3235 [00:01<1:38:07,  1.82s/it](smollm-135m-sft, pid=4300)   0%|          | 2/3235 [00:02<57:22,
1.06s/it]
```

wizeng23: It doesn't seem to be related to the sky ANSI escaping logic

wizeng23:
Could you keep the PR description updated with the latest status on how the PR works, any issues with it, etc.?

src/oumi/cli/launch.py
```diff
@@ -101,6 +102,52 @@ def _cancel_worker(id: str, cloud: str, cluster: str) -> bool:
...
+    """Tails logs with pretty CLI output.
+
+    This function reads from a log stream and displays the output with rich formatting
+    for CLI users. Optionally saves the output to a file.
+
+    Args:
+        log_stream: A LogStream object that can be read from.
+        job_id: The ID of the job being tailed.
+        output_filepath: Optional path to a file to save the logs to.
+    """
```
wizeng23, line 117: nit: Could you add a console print line here mentioning that you're tailing logs?

src/oumi/cli/launch.py
```diff
@@ -101,6 +102,52 @@ def _cancel_worker(id: str, cloud: str, cluster: str) -> bool:
...
+            with cli_utils.CONSOLE.status(f"Tailing logs to {output_filepath}..."):
+                for line in iter(log_stream.readline, ""):
+                    file_handle.write(line)
+                    file_handle.flush()
+        else:
+            for line in iter(log_stream.readline, ""):
+                # Because Rich is rendering markup/styled text and
+                # escapes control characters like \r, we need to handle it specially.
+                if "\r" in line:
+                    cli_utils.CONSOLE.print(line.strip(), end="\r")
```
wizeng23: Adding a comment here for what we discussed in person. This PR is actually changing the logging metadata in our logs. For example:

```
# before
(smollm-135m-sft, pid=3140) [2025-09-02 23:12:34,734][oumi][rank0][pid:3579][MainThread][INFO]][base_map_dataset.py:376] Finished transforming dataset (AlpacaDataset)! Speed: 1418.80 examples/sec. Examples: 51760. Duration: 36.5 sec. Transform workers: 1.
# after
(smollm-135m-sft, pid=3183) [2025-09-02 21:45:19,524][MainThread][INFO]] Finished transforming dataset (AlpacaDataset)! Speed: 1422.14 examples/sec. Examples: 51760. Duration: 36.4 sec. Transform workers: 1.
```

I think this is primarily due to using Rich. This is what the log line looks like if you ran the `oumi train` job natively on Mac:

```
[09/02/25 16:17:23] INFO     [rank-0] Finished transforming dataset (AlpacaDataset)! Speed: 3830.65 examples/sec. Examples: 51760. Duration: 13.5 sec. Transform workers: 1.                                                base_map_dataset.py:376
```

Preferably the log metadata should be preserved exactly as-is from the logs output from the job, and my second preference would be for it to exactly match the formatting of our native Rich CLI output.

PR #1952, opened 2025-08-25
Run GitHub workflows in merge queue instead of on PR events

wizeng23:
# Description

Currently, our test workflows are run whenever a PR is opened, synchronize, or reopened. This is eating into our quota. This PR instead has those tests run in a merge queue when the PR is submitted (with the exception of pretest.yaml, which we keep running for PR updates). The user is still able to run tests manually before the PR is submitted with the `workflow_dispatch` hook.

Documentation: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

## Related issues

Fixes OPE-1512

## Before submitting

### DIFF
--- a/.github/workflows/pretest.yaml
+++ b/.github/workflows/pretest.yaml
@@ -4,6 +4,12 @@ on:
   schedule:
     - cron: '0 */8 * * *' # Every 8 hours
   workflow_dispatch:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'notebooks/**'
+      - 'scripts/**'
+      - 'README.md'
   merge_group:
     paths-ignore:
       - 'docs/**'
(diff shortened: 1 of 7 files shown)


.github/workflows/pretest.yaml
```diff
@@ -4,7 +4,7 @@ on:
   schedule:
     - cron: '0 */8 * * *' # Every 8 hours
   workflow_dispatch:
-  pull_request:
```
taenin: We could consider keeping `pretest` for pull_requests in addition to merge_group since it's  only a CPU job

taenin (approved):

PR #1721, opened 2025-06-04
Support OLCF (Oak Ridge Leadership Computing Facility) Frontier HPC cluster in Oumi launcher

nikg4:
# Description

-- Add support for the new `frontier` cloud, which is mapped to OLCF Frontier cluster.
-- The implementation generally follows the ALCF Polaris pattern, and relies on `SlurmClient`.
-- `SlurmClient` is extended for Frontier: new `sbatch` arguments are now supported. 
-- Add sample config: `configs/recipes/deepseek_r1/sft/distill_qwen_1_5b/full_frontier_job.yaml`
-- Minor updates to `scripts/frontier/launcher.sh` for readability.

Wiki:  https://github.com/oumi-ai/docs/wiki/OLCF-Frontier

Sample launch command:

```
oumi launch up -c configs/recipes/deepseek_r1/sft/distill_qwen_1_5b/full_frontier_job.yaml \
    --cluster batch.$OLCF_USER --user $OLCF_USER
```

## Related issues

Fixes OPE-1281

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [X] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [X] Did you link the issue(s) related to this PR in the section above?
- [X] Did you add / update tests where needed?

## Reviewers

At least one review from a member of `oumi-ai/oumi-staff` is required.

PR #1959, opened 2025-08-29
Support NERSC Perlmutter HPC cluster in Oumi launcher

wizeng23:
# Description

This is similar to #1721.

- Adds Perlmutter cluster and cloud, which are similar to their Frontier counterparts. It reuses the Slurm client.
- Add Perlmutter example job config.
- Fix bug with setting logger level in launcher CLI.
- Limit date range when listing jobs in Slurm to one month to prevent error.

Tested with `oumi launch up -c configs/recipes/qwen3/sft/8b_full/perlmutter_job.yaml --cluster regular.$NERSC_USER --user $NERSC_USER --log-level DEBUG`

## Related issues

Fixes OPE-1507

## Before submitting


scripts/perlmutter/README.md
shanghongsim, line 1: Are error logs now together with output logs?

src/oumi/launcher/clusters/perlmutter_cluster.py
```diff
@@ -0,0 +1,347 @@
...
+            node_count=job.num_nodes,
+            ntasks=job.num_nodes,
+            threads_per_core=1,
+            qos=self._queue.value,
+            stdout_file=(
+                str(stdout_file)
+                if stdout_file
+                else "$CFS/$SBATCH_ACCOUNT/users/$USER/jobs/logs/%j.out"
+            ),
+            stderr_file=str(stderr_file) if stderr_file else None,
```
shanghongsim, line 325: So for Perlmutter, stderr could be None. How are we keeping track of errors?

scripts/perlmutter/README.md
wizeng23: The default for Slurm is to combine stderr and stdout into one log file. This PR doesn't change it, and only changes the path for that log file. IMO it's better to keep the outputs merged, and it will make it easier to tail logs when I add that feature in a follow-up PR.

src/oumi/launcher/clusters/perlmutter_cluster.py
wizeng23: See above, stderr by default is streamed to the same file as stdout. I'll add a comment for clarity

PR #1964, opened 2025-09-03
Added Launcher Logs Command

rlehman221:
# Description

This PR introduces a new feature that allows logs to be retrieved directly via the `launch` command:

```bash
oumi launch logs
```

Currently, this functionality is supported only for **Slurm** and **SkyPilot**. Support for additional launchers will be added in the future.

*Usage*

```bash
oumi launch logs --cluster <cluster_name> --job-id <job_id> --output-filepath <file_path>
```

*Parameters:*
- `--cluster` (**required**): The target cluster name.
- `--job-id` (optional): If omitted, logs will be pulled from the most recent job on the specified cluster.
- `--output-filepath` (optional): If provided, logs will be written to the specified file; otherwise, they will stream directly to the console.

Notes
- When a job is still running, this command will also **tail logs in real time**.
- Only one cluster can be specified per command execution.
- Future updates will expand support beyond Slurm and SkyPilot.

Testing Done

- Ran launcher up command with slurm (both with an output file to be streamed to and the console)
- Ran launcher up command with sky (AWS & GCP) (both with an output file to be streamed to and the console)
- Ran the launcher log command for slurm, AWS, GCP
- Ran the launcher up command with unsupported clouds

## Related issues

Towards OPE-1519

## Before submitting

## Reviewers

### DIFF
--- a/tests/unit/launcher/clients/test_slurm_client.py
+++ b/tests/unit/launcher/clients/test_slurm_client.py
@@ -129,7 +129,7 @@ def test_slurm_client_submit_job_name(mock_subprocess):
                     [
                         "cd work_dir",
                         "sbatch --nodes=2 --job-name=somename"
-                        "--parsable --output=$HOME/oumi_slurm_logs/%j.out "
+                        " --parsable --output=$HOME/oumi_slurm_logs/%j.out "
                         "./job.sh",
                     ]
                 ),
@@ -199,7 +199,7 @@ def test_slurm_client_submit_job_with_all_args(mock_subprocess):
                                 "--error=$HOME/stderr.txt",
                                 "--foo=bar",
                                 "--parsable",
-                                "--output=$HOME/stdout.txt",
+                                "--output=~/stdout.txt",
                                 "./job.sh",
                             ]
                         ),


wizeng23:
Could you fix the failing test? Also, can you confirm in the PR description what testing you've done for the PR (ex. have you confirmed it works for Sky and Slurm)?

src/oumi/cli/launch.py
```diff
@@ -191,43 +190,64 @@ def _down_worker(cluster: str, cloud: Optional[str]) -> bool:
     return True  # Always return true to indicate that the task is done.
 
 
-def _stop_worker(cluster: str, cloud: Optional[str]) -> bool:
-    """Stops a cluster.
+def _find_clusters(cluster: str, cloud: Optional[str]) -> Optional["BaseCluster"]:
```
wizeng23: nit: _find_cluster, since it only returns one

src/oumi/cli/launch.py
```diff
@@ -611,3 +631,46 @@ def which(level: cli_utils.LOG_LEVEL_TYPE = None) -> None:
...
+
+def logs(
+    cluster: Annotated[str, typer.Option(help="The cluster to get the logs of.")],
+    cloud: Annotated[
+        Optional[str],
+        typer.Option(
+            help="If specified, only clusters on this cloud will be affected."
+        ),
+    ] = None,
+    job_id: Annotated[
```
wizeng23, line 642: For job_id and output_filepath, could you modify the help messages to explain what happens if they aren't specified?

src/oumi/cli/launch.py
```diff
@@ -611,3 +631,46 @@ def which(level: cli_utils.LOG_LEVEL_TYPE = None) -> None:
...
         )
     )
+
+
+def logs(
+    cluster: Annotated[str, typer.Option(help="The cluster to get the logs of.")],
+    cloud: Annotated[
+        Optional[str],
+        typer.Option(
+            help="If specified, only clusters on this cloud will be affected."
```
wizeng23: nit: affected doesn't make sense here IMO. What about "The cloud the cluster belongs to" or something like that?

src/oumi/core/launcher/base_cluster.py
```diff
@@ -98,11 +99,13 @@ def down(self) -> None:
...
-    def get_logs_stream(self, job_id: str, cluster_name: str) -> io.TextIOBase:
+    def get_logs_stream(
+        self, cluster_name: str, job_id: Optional[str] = None
+    ) -> io.TextIOBase:
         """Gets a stream that tails the logs of the target job.
 
         Args:
-            job_id: The ID of the job to tail the logs of.
             cluster_name: The name of the cluster the job was run in.
+            job_id: The ID of the job to tail the logs of.
```
wizeng23: Can you mention in the comment what happens if job_id is none

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -29,6 +29,8 @@
 
 _CTRL_PATH = "-S ~/.ssh/control-%h-%p-%r"
 
+_LOG_DIR = "/home/{user}/slurm_logs/{job_id}"
```
wizeng23: 1. Not all Slurm clusters may have the same home directory setup
2. It might be good to include oumi in the path to avoid name conflicts
3. Instead of having directories named after the job with only one file inside, how about just naming the files after the job id? Consider also suffixing with .out to match Slurm's usual log file suffix.

How about "~/oumi_slurm_logs/{job_id}.out"?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -338,16 +330,17 @@ def _start_tail_process(
                 delay = min(base_delay * (2**attempt), max_delay)
                 time.sleep(delay)
             else:
+                log_path = (
```
wizeng23: Is this duplicate line needed?

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -321,10 +312,11 @@ def _start_tail_process(
         # Wait for log file to appear
         for attempt in range(max_attempts):
             # Use direct SSH command to avoid pseudo-terminal issues
-            check_cmd = (
-                f"ssh {_CTRL_PATH} {cluster_name} "
-                f'"cd {working_dir} && test -f {stdout_filename}"'
+            log_path = (
```
wizeng23: nit: Move this and check_cmd outside the for loop

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -726,20 +725,17 @@ def put(self, file_contents: str, destination: str) -> None:
...
-            job_id,
-            cluster_name,
-            stdout_filename,
-            client=self,
-        )
+        if job_id is None:
+            lastest_job_id = self.get_latest_job()
+            if lastest_job_id is None:
+                raise RuntimeError("No jobs found on the cluster.")
+            job_id = lastest_job_id.id
```
wizeng23, line 766: nit: could you add a log line to print out what job id was found?

src/oumi/launcher/clusters/slurm_cluster.py
```diff
@@ -274,7 +274,9 @@ def run_job(self, job: JobConfig) -> JobStatus:
             str(script_path),
             str(remote_working_dir),
             job.num_nodes,
-            job_name,
+            stdout_file=f"{_LOG_DIR.format(user='%u', job_id='%j')}/stdout.log",
+            stderr_file=f"{_LOG_DIR.format(user='%u', job_id='%j')}/stderr.log",
```
wizeng23: What happens when you set these both to the same thing? If you only set stdout_file and don't set stderr_file, the stdout and stderr streams will both go to the stdout file

src/oumi/cli/launch.py
```diff
@@ -191,43 +190,64 @@ def _down_worker(cluster: str, cloud: Optional[str]) -> bool:
     return True  # Always return true to indicate that the task is done.
 
 
-def _stop_worker(cluster: str, cloud: Optional[str]) -> bool:
-    """Stops a cluster.
+def _find_clusters(cluster: str, cloud: Optional[str]) -> Optional["BaseCluster"]:
+    """Find clusters matching the given name and cloud.
```
shanghongsim: nit: same comment as above, cluster instead of clusters

src/oumi/launcher/clusters/sky_cluster.py
```diff
@@ -122,11 +122,13 @@ def down(self) -> None:
         """Tears down the current cluster."""
         self._client.down(self.name())
 
-    def get_logs_stream(self, job_id: str, cluster_name: str) -> SkyLogStream:
+    def get_logs_stream(
+        self, cluster_name: str, job_id: Optional[str] = None
```
shanghongsim, line 126: Can you also explain what happens if job_id is None here?

src/oumi/launcher/clients/slurm_client.py
rlehman221: Anybody using these Slurm clusters I thought have a user setup though?

src/oumi/launcher/clusters/slurm_cluster.py
rlehman221: Updated to have them as the defaults

src/oumi/launcher/clusters/sky_cluster.py
rlehman221: Updated in base_cluster

src/oumi/launcher/clients/slurm_client.py
wizeng23: Our Slurm cluster uses `/home/{user}` as the home directory, but that may not be true for all Slurm clusters

.pre-commit-config.yaml
```diff
@@ -40,15 +40,6 @@ repos:
       - id: ruff-format
         types_or: [python, pyi, jupyter]
         exclude: src/oumi/(models/experimental/cambrian|core/types/proto)"
-  - repo: local
```
wizeng23: If you only deleted this for local testing, could you revert the change?

src/oumi/cli/launch.py
```diff
@@ -611,3 +631,51 @@ def which(level: cli_utils.LOG_LEVEL_TYPE = None) -> None:
             border_style="blue",
         )
     )
+
+
+def logs(
+    cluster: Annotated[str, typer.Option(help="The cluster to get the logs of.")],
+    cloud: Annotated[
+        Optional[str],
+        typer.Option(help="If specified, will use the cloud the clusters belong to."),
```
wizeng23: Sorry for my previous comment wasn't clear enough, but this comment doesn't seem clear to me. What about "If specified, filter for clusters in this cloud."

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -534,8 +522,8 @@ def submit_job(
         distribution: Optional[str] = None,
         partition: Optional[str] = None,
         qos: Optional[str] = None,
-        stdout_file: Optional[str] = None,
-        stderr_file: Optional[str] = None,
+        stdout_file: str = f"{_LOG_DIR.format(user='%u', job_id='%j')}/stdout.log",
```
wizeng23: Sorry if it wasn't clear from previous comments, but IMO it's preferable to have the stdout and stderr merged in the same file. Could you set stdout_file to `"~/oumi_slurm_logs/{job_id}.out"` and stderr_file to None? This filename is also consistent with the default Slurm log filename

src/oumi/launcher/clients/slurm_client.py
rlehman221: I see, if we use ~/oumi_slurm_logs/{job_id}.out is the directory space shared? So both my jobs and your jobs will now show up in that directory? Won't we have clashing job_ids? I thought these were specific to the user?

.pre-commit-config.yaml
rlehman221: yes only deleting for local testing

rlehman221:
@wizeng23 I don't think we should make it an option to change the strderr strdout file paths for submitting a job for slurm. I think we should keep it parity with sky. If we open this up to allowing their own file paths we have to refactor how we grab the logs with sky launch logs since now we also need an extra parameter for the file path where they could of saved the logs. This starts to move away from uniformity with sky.

src/oumi/launcher/clusters/perlmutter_cluster.py
```diff
@@ -300,7 +300,7 @@ def run_job(self, job: JobConfig) -> JobStatus:
         # We pass in the stdout/stderr files via a flag in the sbatch command so that
         # we can do variable expansion. Env vars don't get expanded in #SBATCH
         # directives.
-        logging_dirs, stdout_file, stderr_file = _get_logging_dirs_and_files(job_script)
+        logging_dirs, _, _ = _get_logging_dirs_and_files(job_script)
```
rlehman221, line 303: @wizeng23 I want to make sure I got your input on this? Based on what we talked about when using slurm we will have the default path always provided

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -522,8 +519,6 @@ def submit_job(
         distribution: Optional[str] = None,
         partition: Optional[str] = None,
         qos: Optional[str] = None,
-        stdout_file: str = f"{_LOG_DIR.format(user='%u', job_id='%j')}/stdout.log",
```
wizeng23: Could you keep these flags and just give them default values? Something like:

```
stdout_file: str = _LOG_DIR.format(job_id="%j"),
stderr_file: Optional[str] = None,
```

This way, we keep the function flexible in case anyone uses it directly via code

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -597,9 +586,7 @@ def submit_job(
             cmd_parts.append(f"--{flag}={value}")
 
         cmd_parts.append("--parsable")
-        # Important that setting the output must come before the job path.
-        cmd_parts.append(f"--output={stdout_file}")
-        cmd_parts.append(f"--error={stderr_file}")
+        cmd_parts.append(f"--output={_LOG_DIR.format(job_id='%j')}")
```
wizeng23: Could you also delete line 34 in configs/recipes/qwen3/sft/8b_full/perlmutter_job.yaml and 36-37 in configs/recipes/deepseek_r1/sft/distill_qwen_1_5b/full_frontier_job.yaml given this change?

src/oumi/launcher/clients/slurm_client.py
rlehman221: This will break the launch logs though since we don't currently allow to input a custom file for the output

src/oumi/launcher/clients/slurm_client.py
```diff
@@ -534,7 +519,7 @@ def submit_job(
         distribution: Optional[str] = None,
         partition: Optional[str] = None,
         qos: Optional[str] = None,
-        stdout_file: Optional[str] = None,
+        stdout_file: str = _LOG_DIR.format(job_id="%j"),
```
wizeng23, line 522: nit: could the type still be Optional[str]? This way, users of this function can keep the default slurm log output location by setting both the log file flags to None.

PR #1977, opened 2025-09-19
Separate user docs + Sphinx and source code

stefanwebb:
This PR removes the user docs and Sphinx build commands so that the repo is independent of the documentation


stefanwebb:
WiP until new user docs released

stefanwebb:
To resume later

PR #1979, opened 2025-09-22
Update uv pip install commands to include --system

jgreer013:
# Description

E2E tests are failing now due to a message suggesting to use `uv venv` or `--system`. Updating pip installs to use `--system` which aligns with how our workflow tests run.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


wizeng23:
Why do we have to update all our uv install commands? Shouldn't we only need to update the e2e test job?

Based on the description of the flag, I don't think that's what we want, as we'd want to install into the conda env instead

PR #1983, opened 2025-09-27
Bedrock Inference Engine Integration

aniruddh-alt:
# Description

## Add AWS Bedrock Converse API Support

This PR adds support for AWS Bedrock's Converse API as a new inference engine in Oumi.

### Key Features Added:

- **New `BedrockInferenceEngine`** - Implements AWS Bedrock Converse API integration
- **Framework Integration** - Added `BEDROCK` to `InferenceEngineType` enum and registered in the engine builder
- **AWS Credentials Support** - Automatic AWS credential handling via boto3 (IAM roles, profiles, environment variables)
- **Concurrent Processing** - Async inference with proper concurrency control and rate limiting
- **Error Handling** - AWS-specific error handling with retry logic and exponential backoff
- **Parameter Support** - Supports key generation parameters: `max_new_tokens`, `temperature`, `top_p`, `stop_strings`

### Implementation Details:

- Uses **boto3** for AWS SDK integration (preferred over raw HTTP for AWS services)
- Extends `RemoteInferenceEngine` with Bedrock-specific implementations
- Overrides batch inference methods with `NotImplementedError` (Bedrock doesn't support OpenAI-style batch API)
- Implements required abstract methods: `_infer_online()`, `_infer()`, `_query_api()`
- Includes comprehensive unit tests for conversation conversion and API response parsing

### Usage Example:

```python
from oumi.builders.inference_engines import build_inference_engine
from oumi.core.configs import InferenceEngineType, ModelParams, RemoteParams

# Create Bedrock engine
engine = build_inference_engine(
    InferenceEngineType.BEDROCK,
    ModelParams(model_name="anthropic.claude-3-sonnet-20240229-v1:0"),
    RemoteParams(extra={"aws_region": "us-east-1"})
)

# Run inference
results = engine.infer(conversations)
```

### Configuration:

```yaml
model:
  model_name: "anthropic.claude-3-sonnet-20240229-v1:0"

inference_engine:
  engine_type: "BEDROCK"
  remote_params:
    num_workers: 10
    extra:
      aws_region: "us-east-1"

generation:
  max_new_tokens: 1024
  temperature: 0.7
  top_p: 0.9
```

## Related issues

Fixes # (issue number from GitHub)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/bedrock_inference_engine.py
+++ b/src/oumi/inference/bedrock_inference_engine.py
@@ -1,0 +1,347 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from typing import Any, Optional
+
+import boto3
+from botocore.exceptions import ClientError
+from tqdm.asyncio import tqdm
+from typing_extensions import override
+
+from oumi.core.async_utils import safe_asyncio_run
+from oumi.core.configs import GenerationParams, ModelParams, RemoteParams
+from oumi.core.types.conversation import Conversation, Message, Role
+from oumi.inference.adaptive_semaphore import PoliteAdaptiveSemaphore
+from oumi.inference.remote_inference_engine import RemoteInferenceEngine
+from oumi.utils.logging import logger
+
+_CONTENT_KEY: str = "content"
+_ROLE_KEY: str = "role"
+
+
+class BedrockInferenceEngine(RemoteInferenceEngine):
+    """Inference engine for running inference against the Bedrock API.
+
+    This class extends RemoteInferenceEngine to provide specific functionality
+    for interacting with Bedrock's language models via their API. It handles
+    the conversion of Oumi's Conversation objects to Bedrock's expected input
+    format, as well as parsing the API responses back into Conversation objects.
+    """
+
+    @property
+    @override
+    def base_url(self) -> Optional[str]:
+        """Return the default base URL for the Bedrock API."""
+        return None
+
+    @property
+    @override
+    def api_key_env_varname(self) -> Optional[str]:
+        """Return the default environment variable name for the Bedrock API key."""
+        return None
+
+    def _bedrock_client(self, remote_params: RemoteParams) -> Any:
+        region = getattr(remote_params, "extra", {}).get("aws_region")
+        return boto3.client("bedrock-runtime", region_name=region)
+
+    @override
+    def _convert_conversation_to_api_input(
+        self,
+        conversation: Conversation,
+        generation_params: GenerationParams,
+        model_params: ModelParams,
+    ) -> dict[str, Any]:
+        """Converts a conversation to a Bedrock API input.
+
+        Args:
+            conversation: The conversation to convert.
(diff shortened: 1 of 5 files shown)


aniruddh-alt:
I have tested the inference engine with my AWS API key and everything works fine. I have attached a script of a local script I ran for your reference:
image

src/oumi/inference/bedrock_inference_engine.py
```diff
@@ -0,0 +1,347 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from typing import Any, Optional
+
+import boto3
```
oelachqar: @aniruddh-alt since boto3 is not a core dependency for oumi, we should move the import to be within the BedrockInferenceEngine class so it's only imported when it's actually needed. If it's not available, we can raise a helpful message so that users know they need to install it. You can check https://github.com/oumi-ai/oumi/blob/0f522389dc3a9ff8a8ee88529f1d564bcd1d95b2/src/oumi/inference/llama_cpp_inference_engine.py#L33 for an example

oelachqar:
Thank you @aniruddh-alt for the contribution! Reviewing now

src/oumi/inference/bedrock_inference_engine.py
aniruddh-alt: Understood! Will make this change.

aniruddh-alt: Hello @oelachqar, I have made this change, which follows the LlamaCpp Inference Engine, to first detect if boto3 is installed and raise an error to install boto3.

aniruddh-alt:
I just realized that there are some issues with my unit tests in terms of importing boto3. I will refer to the Llama cpp engine tests and fix these issues. Apologies for the inconvenience!

aniruddh-alt:
I have fixed my unit tests and also addressed the failing CPU tests with my latest commit.

src/oumi/inference/bedrock_inference_engine.py
```diff
@@ -0,0 +1,384 @@
...
+        return None
+
+    @property
+    @override
+    def api_key_env_varname(self) -> Optional[str]:
+        """Return the default environment variable name for the Bedrock API key."""
+        return None
+
+    def _bedrock_client(self, remote_params: RemoteParams) -> Any:
+        region = getattr(remote_params, "extra", {}).get("aws_region")
```
oelachqar: @aniruddh-alt I believe remote_params does not have an `extra` attribute. This is something we can consider adding in the future, but for now what do you think about using the `AWS_REGION` environment variable?

src/oumi/inference/bedrock_inference_engine.py
```diff
@@ -0,0 +1,384 @@
...
+        """Returns the default remote parameters."""
+        return RemoteParams()
+
+    @override
+    def _set_required_fields_for_inference(self, remote_params: RemoteParams):
+        """Override to skip API key validation since Bedrock uses AWS credentials."""
+        pass
+
+    @override
+    def _infer_online(
```
oelachqar: @aniruddh-alt given the bedrom engine does not change the behavior, let's skip overriding this function

aniruddh-alt: Yes I will remove this!

src/oumi/inference/bedrock_inference_engine.py
aniruddh-alt: Hello @oelachqar, I think that makes a lot of sense! I investigated different alternatives for using the AWS_REGION variable, and I believe using the AWS_REGION env variable would be very user-friendly.  I updated the Bedrock engine to read the region from the AWS_REGION environment variable, similar to how the GCP engine relies on environment variables for REGION. I validated this locally (e.g., AWS_REGION=us-east-2), and it works as expected.

PR #1997, opened 2025-10-31
Expose chat_template_kwargs param in ModelParams

shanghongsim:
# Description

This change is to enable users to switch `thinking` on/off for Qwen3 models by doing this:
```
inference_config.model.chat_template_kwargs = {"enable_thinking": False}
```

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/model_params.py
+++ b/src/oumi/core/configs/params/model_params.py
@@ -145,7 +145,7 @@ class ModelParams(BaseParams):
 
     chat_template_kwargs: dict[str, Any] = field(default_factory=dict)
     """Additional keyword args to pass to the chat template renderer.
-    Currently, it is being access by vLLM inference engine to pass additional kwargs
+    Currently, it is only used by the vLLM inference engine to pass additional kwargs
     for generation of assistant messages.
     """
 


ryan-arman:
Thanks for this PR. There is just one test that is failing

wizeng23:
1. I thought the Qwen models with toggleable thinking had worse quality, hence why later Qwen3 releases drop the hybrid thinking feature and are either instruct or reasoning models. Any reason you can't use the later models?
2. The new field only affects vLLM generation, but there's no indication in the docstring comment that this is the case. If you don't follow the point below, can you make sure the docstring is clear on where this argument will be ingested?
3. This change seems hyper-specific. Ideally, we extend this PR to support passing in chat template kwargs across all of training, inference, evaluation, etc. (Ex. OPE-1210 for supporting it in training). If this field is only in GenerationParams, we can't have it support training. ModelParams to me seems like a more suitable place. Thoughts?

shanghongsim:
I have shifted the `chat_template_kwargs` from GenerationParams to ModelParams. Would appreciate a second review! 🙏  Thank you! @wizeng23 @ryan-arman

ryan-arman:
> I have shifted the `chat_template_kwargs` from GenerationParams to ModelParams. Would appreciate a second review! 🙏 Thank you! @wizeng23 @ryan-arman

Thank you Shang Hong. Looks good to me. I would wait for Will to review too.

ryan-arman (approved):

.gitignore
```diff
@@ -1,6 +1,7 @@
 # Local experiments
 tmp/
 wandb/
+mlruns/
```
wizeng23, line 5: nit: Is this a folder name you created or does it come from a framework? If the latter, could you add a comment for which one for tracking purposes? For the former, is it possible to save the outputs somewhere else, ex. output/? Just so we don't clutter our gitignore with too many personal directory names.

src/oumi/core/configs/params/model_params.py
```diff
@@ -143,6 +143,12 @@ class ModelParams(BaseParams):
         documentation for the appropriate template to use.
     """
 
+    chat_template_kwargs: dict[str, Any] = field(default_factory=dict)
+    """Additional keyword args to pass to the chat template renderer.
+    Currently, it is being access by vLLM inference engine to pass additional kwargs
```
wizeng23:
```suggestion
    Currently, it is only used by the vLLM inference engine to pass additional kwargs
```

.gitignore
shanghongsim: Its for a framework (databricks MLflow telemetry)! Added the comment

PR #1999, opened 2025-10-31
Added dict support for image ids

oumiandy:
# Description

We encountered reliability issues running training within Docker containers on AWS. As a result, we switched to using AMIs instead of Docker images. To support multi-region deployments without hardcoding a single region, we implemented a solution that allows passing multiple region-specific image IDs at once.

## Related issues

[LOU-92](https://linear.app/oumi/issue/LOU-92/create-training-docker-image-scripts)

## Before submitting

- [] This PR only changes documentation. (You can ignore the following checks in that case)

## Reviewers

### DIFF
--- a/tests/unit/launcher/clients/test_sky_client.py
+++ b/tests/unit/launcher/clients/test_sky_client.py
@@ -147,13 +147,13 @@ def test_convert_job_to_task_with_dict_image_id(
                     job.resources.image_id = None
                     job.resources.image_id_map = {
                         "us-central1": "ami-0c1a2b3c4d5e6f701",
-                        "us-east1":    "ami-0d2e3f4g5h6i7j802",
-                        "us-west1":    "ami-0e3f4g5h6i7j8k903",
-                        "us-west2":    "ami-0f4g5h6i7j8k9l004",
-                        "us-west3":    "ami-0a5b6c7d8e9f0g105",
-                        "us-west4":    "ami-0b6c7d8e9f0g1h206",
-                        "us-west5":    "ami-0c7d8e9f0g1h2i307",
-                        "us-west6":    "ami-0d8e9f0g1h2i3j408",
+                        "us-east1": "ami-0d2e3f4g5h6i7j802",
+                        "us-west1": "ami-0e3f4g5h6i7j8k903",
+                        "us-west2": "ami-0f4g5h6i7j8k9l004",
+                        "us-west3": "ami-0a5b6c7d8e9f0g105",
+                        "us-west4": "ami-0b6c7d8e9f0g1h206",
+                        "us-west5": "ami-0c7d8e9f0g1h2i307",
+                        "us-west6": "ami-0d8e9f0g1h2i3j408",
                     }
 
                     _ = _convert_job_to_task(job)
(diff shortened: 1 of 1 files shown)


oelachqar:
@oumiandy out of curiosity what are the issues with docker on aws? anything we need to fix in the oss repo?

tests/unit/launcher/clients/test_sky_client.py
```diff
@@ -53,6 +52,7 @@ def _get_default_job(cloud: str) -> JobConfig:
         run="./hello_world.sh",
     )
 
+#
```
taenin: Nit: delete this line

min-oumi:
> @oumiandy out of curiosity what are the issues with docker on aws? anything we need to fix in the oss repo?

Currently when we launch a100s it does not have cuda drivers installed. Hence we must specify AMIs that do have these drivers, but the AMIs are different for each region. This change allows us to specify all AMIs for all regions we have, so that skypilot is able to choose whichever one available, compared to our current workflow where we are specifying a single region.

oumiandy:
> @oumiandy out of curiosity what are the issues with docker on aws? anything we need to fix in the oss repo?

We were having issues with torch being usable with cuda on the 8x A100 machines. We tried the torch images from docker hub and amazon's official images, but could not get it to work. We believe its an issue with AWS.

oelachqar:
> > @oumiandy out of curiosity what are the issues with docker on aws? anything we need to fix in the oss repo?
> 
> We were having issues with torch being usable with cuda on the 8x A100 machines. We tried the torch images from docker hub and amazon's official images, but could not get it to work. We believe its an issue with AWS.

Have you tried the official oumi image? Does it also have the same issue?

oumiandy:
> > > @oumiandy out of curiosity what are the issues with docker on aws? anything we need to fix in the oss repo?
> > 
> > 
> > We were having issues with torch being usable with cuda on the 8x A100 machines. We tried the torch images from docker hub and amazon's official images, but could not get it to work. We believe its an issue with AWS.
> 
> Have you tried the official oumi image? Does it also have the same issue?

We used the torch and amazon images as the base for a modified version of the official Oumi image (I had to remove the Oumi user portion because it did not have sudo privileges)

src/oumi/core/configs/job_config.py
```diff
@@ -120,7 +120,7 @@ class JobResources:
     disk tiers.
     """
 
-    image_id: Optional[str] = None
+    image_id: Union[dict[Optional[str], str], str, None] = None
```
oelachqar: IIRC omegaconf (the lib we use to parse configs) only supports union of primitive types (i.e. not dict). Does this work when loading a config from a file?

oumiandy: I tried switching it to the primitive type in the most recent commit but it seems to break some of the tests

oumiandy: Using Dict seemed to break the ruff formatting test. It suggested I use the primitive type instead

oumiandy: Looks like the issue is with OmegaConf not being compatible with a dict inside of a union. I will resolve on Monday!

tests/unit/launcher/clients/test_sky_client.py
```diff
@@ -129,6 +129,128 @@ def test_convert_job_to_task(
...
+                    job.resources.image_id = None
+                    job.resources.image_id_map = {
+                        "us-central1": "docker://ubuntu:latest",
+                        "us-east1": "docker://ubuntu:latest",
+                        "us-west1": "docker://ubuntu:latest",
+                        "us-west2": "docker://ubuntu:latest",
+                        "us-west3": "docker://ubuntu:latest",
+                        "us-west4": "docker://ubuntu:latest",
+                        "us-west5": "docker://ubuntu:latest",
+                        "us-west6": "docker://ubuntu:latest",
```
taenin: Super nit: for this test, could you change one of the values to be different? Otherwise you could change your code to keep the keys for this dictionary and set the value equal to the value of the first key and your tests would still pass.

oumiandy: Updated

src/oumi/core/configs/job_config.py
oumiandy: Fixed in latest commit

tests/unit/launcher/clients/test_sky_client.py
oumiandy: Done

PR #2021, opened 2025-11-18
enable custom master port for distributed training

monnetb:
# Description

The Pytorch distributed port is hard coded as 8007 ( _DEFAULT_MASTER_PORT ) and can not be changed when using Slurm or Polaris scheduler. However, due to some internal/entreprise restriction, this port might be unavailable.
It should be possible to use the 'classical' environement variable MASTER_PORT defined in Pytorch for more flexibility

# Solution
The changes in the code is to use the existing code from the local machine process

`master_port=int(env.get(_MASTER_PORT_ENV, _DEFAULT_MASTER_PORT))`

If the environement is not set, we fall back to the default value  = 8007

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/cli/distributed_run.py
+++ b/src/oumi/cli/distributed_run.py
@@ -437,7 +437,7 @@ def _detect_slurm_process_run_info(env: dict[str, str]) -> Optional[_ProcessRunI
         node_rank=node_rank,
         world_info=_WorldInfo(num_nodes=len(node_ips), gpus_per_node=gpus_per_node),
         master_address=node_ips[0],
-        master_port=_DEFAULT_MASTER_PORT,
+        master_port=int(env.get(_MASTER_PORT_ENV, _DEFAULT_MASTER_PORT)),
         node_ips=node_ips,
     )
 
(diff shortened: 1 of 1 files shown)


src/oumi/cli/distributed_run.py
```diff
@@ -437,7 +437,7 @@ def _detect_slurm_process_run_info(env: dict[str, str]) -> Optional[_ProcessRunI
         node_rank=node_rank,
         world_info=_WorldInfo(num_nodes=len(node_ips), gpus_per_node=gpus_per_node),
         master_address=node_ips[0],
-        master_port=_DEFAULT_MASTER_PORT,
+        master_port=_int(env.get(_MASTER_PORT_ENV, _DEFAULT_MASTER_PORT)),
```
oelachqar: Looks like there is a typo with the trailing underscore!

monnetb (commented):
Oh, Sorry. I missed this one.

src/oumi/cli/distributed_run.py
oelachqar: @monnetb it looks like the typo is still there!

PR #2060, opened 2025-11-26
Fix too many files open error

ryan-arman:
# Description
Cap TCPConnector connection limit at 200 to prevent file descriptor
exhaustion when requests_per_minute is set to 500 or 1000. HTTP
connections are reused, so we don't need one connection per concurrent
request. The semaphore still controls concurrency properly.

## Related issues

Towards LOU-283

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -60,6 +60,7 @@ from oumi.utils.http import (
 _AUTHORIZATION_KEY: str = "Authorization"
 _BATCH_PURPOSE = "batch"
 _BATCH_ENDPOINT = "/v1/chat/completions"
+_MAX_CONNECTION_LIMIT = 200
 
 
 class BatchStatus(Enum):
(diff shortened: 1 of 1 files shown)


src/oumi/inference/remote_inference_engine.py
```diff
@@ -60,6 +60,7 @@
 _AUTHORIZATION_KEY: str = "Authorization"
 _BATCH_PURPOSE = "batch"
 _BATCH_ENDPOINT = "/v1/chat/completions"
+_MAX_CONNECTION_LIMIT = 200
```
taenin, line 63: Any rationale for choosing 200?

min-oumi (approved):

src/oumi/inference/remote_inference_engine.py
ryan-arman: The default value is 100. The recommendation is to increase it slowly up for example to 400. I think 200 is a safe lower end that does work fast enough: https://proxiesapi.com/articles/optimizing-aiohttp-for-high-concurrency

PR #2061, opened 2025-11-27
[tests] consolidate models

oelachqar:
# Description

**Problem**
GPU tests currently take ~20min to run:
- Downloading all the models & datasets required to run the tests takes ~8minutes
- Setting up the docker container takes ~3-4min

**Fixes**
- This PR aims to reduce that by consolidating the number of models that need to be downloaded.
- We also switch to using smaller models (e.g. SmolVLM-256M vs QwenVL-2B)
- We also add a caching step for the models folder
- Revert to using the default ubuntu image for github actions + manually installing cuda
- No logic change, just swapping models + necessary fixes to make tests pass (e.g. updating chat templates)

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/.github/workflows/gpu_tests.yaml
+++ b/.github/workflows/gpu_tests.yaml
@@ -18,9 +18,6 @@ jobs:
     permissions:
       contents: "read"
     runs-on: linux-gpu-runner
-    # container:
-    #   image: nvidia/cuda:12.8.1-cudnn-devel-ubuntu24.04
-    #   options: --gpus all
 
     steps:
       - name: Checkout


tests/scripts/predownload_for_github_gpu_tests.sh
```diff
@@ -37,74 +37,31 @@ download_with_retry() {
...
-download_with_retry "bert-base-uncased"
-
-# Add delay between downloads to avoid rate limiting
-sleep 10
-
-download_with_retry "t5-small"
-
-# Add delay between downloads to avoid rate limiting
-sleep 10
+download_with_retry "Qwen/Qwen3-0.6B" \
```
wizeng23, line 53: Are we using this model in tests anywhere right now?

.github/workflows/gpu_tests.yaml
```diff
@@ -18,9 +18,9 @@ jobs:
     permissions:
       contents: "read"
     runs-on: linux-gpu-runner
-    container:
-      image: nvidia/cuda:12.8.1-cudnn-devel-ubuntu22.04
-      options: --gpus all
+    # container:
```
wizeng23: Delete commented block?

tests/scripts/predownload_for_github_gpu_tests.sh
oelachqar: Yes, to test the think/no-think chat template

.github/workflows/gpu_tests.yaml
oelachqar: Updated!

PR #2065, opened 2025-12-01
Enable TogetherAI api to output reasoning traces consistently

shanghongsim:
# Description

1. Consume `reasoning` argument of `ChatCompletions` object
Reasoning/hybrid models like Kimi K2, DeepSeek v3.1, GPT OSS 120B returns thinking traces in the `reasoning` argument of `ChatCompletionMessage` object. 

```
ChatCompletionMessage(content='Of course! The capital of France is **Paris**.', refusal=None, role='assistant', annotations=None, audio=None, function_call=None, tool_calls=[], reasoning="Hmm, the user is asking a straightforward factual question about the capital of France. This is a common knowledge query with a clear answer. \n\nI recall that Paris is the capital and has been for centuries, so I can confirm that directly. Since the question is simple, no additional context or elaboration is needed unless the user asks for more. \n\nI'll keep the response concise and accurate, just stating the answer with a brief mention of its historical status to add slight value without overcomplicating it.")
```

Currently, Oumi conversations only consumes the `content` argument of `ChatCompletionMessage` object. This PR modifies TogetherAI inference engine API to concatenate `reasoning` and `content` fields of the `ChatCompletionMessage` for the `content` field of oumi `Conversations` output.

2. Enable reasoning in API
Hybrid models like DeepSeek-V3.1 need `"reasoning": {"enabled": True},` to be passed in via API input to turn on reasoning. This PR adds optional `api_kwargs` to `RemoteParams` to allow users to pass in such arguments. 

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/inference/together_inference_engine.py
+++ b/src/oumi/inference/together_inference_engine.py
@@ -12,18 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
+import copy
 from typing import Any, Optional
 
-import pydantic
 from typing_extensions import override
 
 from oumi.core.configs import GenerationParams, ModelParams
-from oumi.core.types.conversation import Conversation, Message, Role
+from oumi.core.types.conversation import Conversation
 from oumi.inference.remote_inference_engine import RemoteInferenceEngine
-from oumi.utils.conversation_utils import (
-    convert_message_to_json_content_list,
-)
 from oumi.utils.logging import logger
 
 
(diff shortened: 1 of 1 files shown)


wizeng23:
I don't see any logic changes in this PR, only test changes

src/oumi/inference/together_inference_engine.py
```diff
@@ -33,3 +41,158 @@ def base_url(self) -> Optional[str]:
...
+                    content=message["content"],
+                    role=Role(message["role"]),
+                ),
+            ],
+            metadata=original_conversation.metadata,
+            conversation_id=original_conversation.conversation_id,
+        )
+
+    @override
+    def _convert_conversation_to_api_input(
```
wizeng23, line 87: Per [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), and to make sure this implementation doesn't diverge from the parent class over time, could you call `super._convert_conversation_to_api_input()` and then add any modifications you need after that? Ditto for `_convert_api_output_to_conversation`.

src/oumi/core/configs/params/remote_params.py
```diff
@@ -82,6 +82,13 @@ class RemoteParams(BaseParams):
     requests.
     """
 
+    api_kwargs: Optional[dict[str, Any]] = None
+    """Additional keyword arguments to pass to the API.
+
+    This allows for passing any API-specific parameters that are not
```
wizeng23, line 88: Could you add a comment here that this is currently only used for the together inference engine?

src/oumi/inference/together_inference_engine.py
```diff
@@ -33,3 +37,82 @@ def base_url(self) -> Optional[str]:
...
+        """Converts an API response to a conversation.
+
+        Args:
+            response: The API response to convert.
+            original_conversation: The original conversation.
+
+        Returns:
+            Conversation: The conversation including the generated response.
+        """
+        try:
```
wizeng23, line 54: I don't quite get what this try-except is trying to do. If the try-block fails, then super()._convert_api_output_to_conversation will also fail because it will also try to extract `message` the exact same way right?

src/oumi/inference/together_inference_engine.py
```diff
@@ -33,3 +37,82 @@ def base_url(self) -> Optional[str]:
...
+            message = response["choices"][0]["message"]
+        except (KeyError, IndexError, TypeError):
+            return super()._convert_api_output_to_conversation(
+                response, original_conversation
+            )
+
+        if "reasoning" in message and "</think>" not in message.get("content", ""):
```
wizeng23, line 61: Do you know if any other inference provider may include a "reasoning" field? If this is a common convention, we should move the logic in this function into the parent class RemoteInferenceEngine.

src/oumi/inference/together_inference_engine.py
```diff
@@ -33,3 +37,82 @@ def base_url(self) -> Optional[str]:
...
+        """
+        api_input = super()._convert_conversation_to_api_input(
+            conversation=conversation,
+            generation_params=generation_params,
+            model_params=model_params,
+        )
+
+        # Then layer on Together-specific / remote-specific kwargs
+        remote_params = self._remote_params
+        if remote_params.api_kwargs:
```
wizeng23, line 113: Ditto here; do you know of any other inference engines that could benefit from this? Should this logic be in RemoteInferenceEngine?

PR #2066, opened 2025-12-01
Make torchdata optional

oelachqar:
# Description

**Problem**
- torchdata >= 0.10 dropped support for datapipes
- torchdata < 0.10 does not have a python3.13 or arm64 wheels
- Since torchdata is a required dependency, this means we can't install oumi on those systems

**Solution**
- This PR removes torchdata from the required dependencies for oumi
- Two use cases require installing it:
  - Using stateful data loaders in `trainer_type=OUMI`
  - Using datapipes processing pipeline using `use_torchdata=True` 

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/packaging.py
+++ b/src/oumi/utils/packaging.py
@@ -174,5 +174,4 @@ def require_torchdata(feature_name: str = "This feature") -> None:
         raise ImportError(
             f"{feature_name} requires torchdata. "
             "Please install it with: pip install 'oumi[torchdata]' "
-            "or pip install torchdata"
         )


src/oumi/utils/packaging.py
```diff
@@ -149,3 +151,28 @@ def check_package_prerequisites(
...
+        return False
+
+
+def require_torchdata(feature_name: str = "This feature") -> None:
+    """Raises an ImportError if torchdata is not available."""
+    if not is_torchdata_available():
+        raise ImportError(
+            f"{feature_name} requires torchdata. "
+            "Please install it with: pip install 'oumi[torchdata]' "
+            "or pip install torchdata"
```
wizeng23: Should we specify version requirements for this command?

wizeng23 (approved):

src/oumi/utils/packaging.py
oelachqar: Good catch, updated!

PR #2100, opened 2025-12-17
[configs] Add train/infer/eval configs 

oelachqar:
# Description

- Add example train/eval/infer configs for recent models 
- Tested on 8xH100

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


wizeng23 (approved):

wizeng23:
nit: train and eval are in nested dirs for different model sizes, but not for inference. Could consider making it consistent between them. We usually nested train/eval in the past because we had GCP job configs but that's not the case here

PR #2319, opened 2026-03-26
Add config-specific exception hierarchy, validation checks, and CLI validation checks, and CLI error handling

idoudali:
# Description

- Introduce a new lightweight `oumi.exceptions` module with an
  `OumiConfigError` base class (kept import-cheap for the CLI).
- Move `HardwareException` to `oumi.exceptions`. It remains importable
  from both `oumi.core.types` and `oumi.core.types.exceptions`; the
  latter is now a thin back-compat shim that re-exports from the
  canonical location.
- Wrap file I/O in `BaseConfig.from_yaml`, the save path, and
  `_read_config_without_interpolation` to raise `OumiConfigError` on
  `FileNotFoundError`, `IsADirectoryError`, and `NotADirectoryError`
  instead of leaking raw OS errors.
- Wrap adapter config loading in `ModelParams` to raise
  `OumiConfigError` on `OSError` and `json.JSONDecodeError`, with
  file path and parse location in the message.
- Handle `OumiConfigError` in the CLI top-level handler so users see a
  concise red error line and exit code 1 instead of a full traceback.
- Update tests accordingly.

## Related issues

Fixes # (issue)

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [ ] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [ ] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

## Reviewers

At least one review from a member of `oumi-ai/oumi-staff` is required.

PR #2325, opened 2026-03-28
Replace config ValueError with OumiConfigValueError

idoudali:
Depends on https://github.com/oumi-ai/oumi/pull/2319

# Description

Update analyze, judge, and quantize unit tests to expect OumiConfigError
instead of ValueError after config validation moved to typed Oumi config
exceptions. This matches runtime behavior and fixes the failing
static-checks test cases.

Fixes OPE-1853

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/data_params.py
+++ b/src/oumi/core/configs/params/data_params.py
@@ -51,7 +51,7 @@ class MixtureStrategy(str, Enum):
         elif self.value == MixtureStrategy.ALL_EXHAUSTED:
             return "all_exhausted"
         else:
-            raise ValueError("Unsupported value for MixtureStrategy")
+            raise OumiConfigError("Unsupported value for MixtureStrategy")
 
 
 class TrainTarget(str, Enum):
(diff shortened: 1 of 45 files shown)


src/oumi/core/configs/params/data_params.py
```diff
@@ -49,7 +50,7 @@ def get_literal_value(self) -> Literal["first_exhausted", "all_exhausted"]:
         elif self.value == MixtureStrategy.ALL_EXHAUSTED:
             return "all_exhausted"
         else:
-            raise ValueError("Unsupported value for MixtureStrategy")
+            raise OumiConfigValueError("Unsupported value for MixtureStrategy")
```
oelachqar, line 53: What's the difference between OumiConfigError and OumiConfigValueError? Should we just use one?

idoudali: We could collapse into one type if need be.

I considered to "specialize" our exception "generic" OumiConfigError, so that it "easier"/obvious the rename the existing ValueError to the new exception type and also to make it hopefully "obvious" to the user what type of exception to raise from now on

idoudali:
rebased on top of latest main

src/oumi/core/configs/params/data_params.py
oelachqar: Let's just OumiConfigError

idoudali:
Hi @oelachqar,
Updated the PR to use OumiConfigError instead of ValueError

oelachqar (approved):

oelachqar (approved):

PR #2335, opened 2026-03-31
[infer] Add progress_callback to inference, judge, and synthesis APIs

rlehman221:
## Summary

- Adds an optional `progress_callback: Callable[[int, int], None]` parameter to `BaseInferenceEngine.infer()`, `BaseJudge.judge()`, and `AttributeSynthesizer.synthesize()`
- The callback fires `(completed_count, total_count)` after each item finishes, enabling callers to report granular progress during bulk operations
- `RemoteInferenceEngine` fires the callback per async task completion in the gather loop; other engines accept the parameter but don't use it yet

## Motivation

When the Oumi Enterprise worker calls bulk inference/judge/synthesis, it passes all rows at once. Without this callback, the worker can only update heartbeat progress at the start and end of the call — leaving users with no progress updates for potentially long-running operations.

## Changes

| File | Change |
|------|--------|
| `base_inference_engine.py` | `progress_callback` on `infer()` and `_infer_online()` |
| `remote_inference_engine.py` | Fires callback per task in async gather loop |
| `native_text_inference_engine.py` | Accepts param in `_infer_online()` |
| `vllm_inference_engine.py` | Accepts param in `_infer_online()` |
| `llama_cpp_inference_engine.py` | Accepts param in `_infer_online()` |
| `base_judge.py` | `progress_callback` on `judge()` and `_infer()`, forwarded to engine |
| `attribute_synthesizer.py` | `progress_callback` on `synthesize()`, forwarded to engine |

## Design

- Callback is optional (`None` default) — fully backwards compatible
- Thread-safe: `asyncio.gather()` runs on a single event loop thread, so the nonlocal counter won't race
- Best-effort: callback exceptions are silently caught to never crash inference
- Engine-level: judge and synthesis just forward the callback to the inference engine

### DIFF
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -19,6 +19,7 @@ import json
 import time
 import uuid
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -91,6 +92,7 @@ class BaseInferenceEngine(ABC):
         self,
         input: list[Conversation] | None = None,
         inference_config: InferenceConfig | None = None,
+        progress_callback: Callable[[int, int], None] | None = None,
     ) -> list[Conversation]:
         """Runs model inference.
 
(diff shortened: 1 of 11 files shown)


rlehman221:
Superseded by the partial-inference stack (#2498–#2502): progress reporting now flows through a cross-process JSON progress file on infer_partial() (progress_path), which also works across the SkyPilot job boundary, alongside per-row partial-failure support.

PR #2361, opened 2026-04-10
Improve MPS (Apple Silicon) logging and metrics reporting

splittingfield:
# Description

Training on MPS already worked, but logging was CUDA-centric — printing "CUDA is not available!" and "GPU runtime info: None" despite actively using the MPS backend. This updates the monitoring code to recognize MPS as a valid accelerator and report available metrics (memory allocated, driver reserved, recommended max).


src/oumi/utils/device_utils.py
```diff
@@ -338,9 +338,26 @@ def get_nvidia_gpu_runtime_info(
 
 
 def log_nvidia_gpu_runtime_info(device_index: int = 0, log_prefix: str = "") -> None:
```
idoudali, line 340: I think this function name is not representative of the functionaltiy as we do not support only nvidia gpu,

probablyt we need to drop `nvidia` from the function name

src/oumi/utils/torch_utils.py
```diff
@@ -78,61 +78,77 @@ def format_cudnn_version(v: int | None) -> str:
...
 
-    for i in range(num_devices):
-        device_name = torch.cuda.get_device_name(i)
-        mem_free, mem_total = torch.cuda.mem_get_info(i)
-        mem_allocated = torch.cuda.memory_allocated(i)
-        mem_reserved = torch.cuda.memory_reserved(i)
-        capability = torch.cuda.get_device_capability(i)
+    log_lines: list[str] = []
+    if torch.cuda.is_available():
+        ncpus = os.cpu_count()
```
idoudali, line 105: The number of CPUs is GPU independent and could move before the if statement

src/oumi/utils/torch_utils.py
```diff
@@ -78,61 +78,77 @@ def format_cudnn_version(v: int | None) -> str:
...
+                f"Free: {_mem_to_gib(mem_free)}GiB "
+                f"Allocated: {_mem_to_gib(mem_allocated)}GiB "
+                f"Cached: {_mem_to_gib(mem_reserved)}GiB]"
+            )
+    elif torch.backends.mps.is_available():
+        ncpus = os.cpu_count()
+        mem_allocated = torch.mps.current_allocated_memory()
+        mem_driver = torch.mps.driver_allocated_memory()
+        mem_recommended_max = torch.mps.recommended_max_memory()
+        log_lines.append(f"CPU cores: {ncpus} MPS devices: 1 (Apple Silicon)")
```
idoudali, line 128: Q: why is MPS devices assumed to be always 1?

src/oumi/utils/device_utils.py
```diff
@@ -338,9 +338,26 @@ def get_nvidia_gpu_runtime_info(
...
+        if torch.backends.mps.is_available():
+            allocated_mb = torch.mps.current_allocated_memory() / 1024**2
+            driver_mb = torch.mps.driver_allocated_memory() / 1024**2
+            logger.info(
+                f"{log_prefix.rstrip()} MPS memory: "
+                f"Allocated: {allocated_mb:.1f} MiB, "
+                f"Driver Reserved: {driver_mb:.1f} MiB."
+            )
+    except ImportError:
+        pass
```
idoudali, line 360: I think we should not suppress the exception here and let it propagate so that this error does not get ignored. 

Is there a chance that we use PyTorch to get the device information, but not actually make use of the pytorch backend ?

wizeng23:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/oumi/builders/callbacks.py
```diff
@@ -70,7 +70,7 @@ def build_training_callbacks(
 
     dtype = next(model.parameters()).dtype
     add_mfu_callbacks: bool = True
-    if not torch.cuda.is_available():
+    if not torch.cuda.is_available() and not torch.backends.mps.is_available():
```
wizeng23, line 73: Can you revert this change? MFU logging doesn't work for MPS because we don't have the peak tflops number here: https://github.com/oumi-ai/oumi/blob/6e4ea8888e01d19533a73d12517b69f2545c3ff9/src/oumi/performance/mfu.py#L20

PR #2369, opened 2026-04-14
Extend DataCollatorForCompletionOnlyLM to support correct tool result masking

shanghongsim:
# Description

## Context

Currently, the completions only collator does not consistently mask tool responses (Role.TOOL) correctly. For some chat templates (like Qwen), it accidentally gets it correct, while masking is consistently wrongly applied for llama chat templates. This PR enables the masking to be correctly applied for all chat templates, supporting all four roles (system, user, assistant and tool).

Without `instruction_template`, tool results are masked correctly. But with `instruction_template`, things get weird and some tool results are unmasked. Omitting `instruction_template` causes the collator to find the last `response_template` and masks everything before it (only the final assistant turn trains). With `instruction_template`, the collator searches for `instruction_template` and `response_template` pairs and masks everything in between.

**Case 1: without `instruction_template`**

```
RESPONSE_TEMPLATE = "<|im_start|>assistant\n"

old_no_inst = DataCollatorForCompletionOnlyLM(
    response_template=RESPONSE_TEMPLATE,
    instruction_template=None,
    tokenizer=tokenizer,
)
b = old_no_inst.torch_call([token_ids])
labels_A = b["labels"][0].tolist()
summarise("Case A", labels_A, N)
show_masking(b["input_ids"][0].tolist(), labels_A, tokenizer)
```

image

Without inst template, collator masks everything before the last resp template. So loss only sees

```
<|im_start|>assistant
The weather in Paris is sunny and 18°C.<|im_end|>
```

**Case 2: with `instruction_template`**

With `instruction_template`, it masks everything between a inst and resp template. Since there isn't an inst template before the tool result, it does not get masked properly. 

```
INSTRUCTION_TEMPLATE = "<|im_start|>user\n"

old_with_inst = DataCollatorForCompletionOnlyLM(
    response_template=RESPONSE_TEMPLATE,
    instruction_template=INSTRUCTION_TEMPLATE,
    tokenizer=tokenizer,
)
b2 = old_with_inst.torch_call([token_ids])
labels_B = b2["labels"][0].tolist()
summarise("Case B", labels_B, N)
show_masking(b2["input_ids"][0].tolist(), labels_B, tokenizer)
```

image

In case 2, the loss sees tool result, which is incorrect.

```
...
<|im_start|>assistant
<tool_call>
{"name": "get_weather", "arguments": {"location": "Paris"}}
</tool_call><|im_end|>
...
<|im_start|>assistant
The weather in Paris is sunny and 18°C.<|im_end|>
```

To confirm my hypothesis of no inst template before the tool result being the cause, I experimented with adding a user turn before the tool call and masking works correctly in that case.

image

## ToolAwareCompletionsCollator
With the new `ToolAwareCompletionsCollator`, tool results are masked properly

Without `instruction_template`
image

With `instruction_template`
image

## Change 1: Extend `DataCollatorForCompletionOnlyLM` with span-based masking
Extend `DataCollatorForCompletionOnlyLM` with span-based masking (detecting assistant turn boundaries via response + end-of-turn tokens) beyond instruction based masking (matching instruction/response string pairs) supported currently. 

## Change 2: `train_target` enum for explicit control of which part of response to train on
- **`assistant_turn`**: Masks everything, then unmasks each assistant response bounded by `response_template` .. `end_of_turn_template` (inclusive of EOT). 
- **`final_assistant_turn`**: Masks all tokens before the last `response_template` occurrence. 

When `train_target` is set, no matter what templates have been provided, `train_target` has final authority on which combination is used. When `train_target` is not set, it is inferred from the templates provided. Eventually, we want users to specify `train_target` instead of relying on combinations of templates to control behavior. However, since we have many configs in production that do not use `train_target` and set templates explicitly, this shall be the interim solution to ensure old configs do not break. Existing configs using `collator_kwargs` with `instruction_template` + `response_template` continue to work via the legacy path (with a deprecation warning). New configs should use `train_target` explicitly. Users can set `train_target` and override the templates manually by setting `collator_kwargs`.

**Before** — users must provide model-specific token strings:
```yaml
collator_name: "text_completions_only_with_padding"
collator_kwargs:
  response_template: "<|im_start|>assistant\n"
  end_of_turn_template: "<|im_end|>"
```

**After** — users express intent, templates auto-resolve from tokenizer:
```yaml
collator_name: "text_completions_only_with_padding"
train_target: "assistant_turn"
```

## Change 3: Auto detection of collator templates
Instead of fragile vocab and collator template matching (requires us to maintain collator templates for all popular models), shift to a more robust approach: apply the chat template to a known test conversation, then finds the assistant boundary strings in the rendered output.

## Migration and deprecation considerations:
1. Configs that **only** specify `collator_name` -> removed support for this with the  removal of the default llama template fallback. This should be ok as no configs in production specify `collator_name` without `collator_kwargs`. Only less than 10 configs in OSS fall in this edgecase and they are being updated in this PR.
2. Enterprise SFT configs -> they should still work with old legacy behavior. Will update them once OSS version is updated in API.

## Open issues for next time
- Move`train_target` into `collator_kwargs` -> I considered putting train_target inside collator_kwargs instead of making it a separate field, but collator_kwargs is an opaque dict. Users wouldn't know train_target exists unless they read the docs. A top-level field makes it more visible. This will eventually be solved when we shift to `CollatorParam`.
- `CollatorParams`: This would mean changing every YAML from:

```
  collator_name: "text_completions_only_with_padding"
  collator_kwargs:
    train_target: "all_assistant_turns"
```
to 

```
  collator:
    name: "text_completions_only_with_padding"
    train_target: "all_assistant_turns"
```

This is much cleaner but its a substantial refactor. In order to manage the scope of this PR, this shall be deferred to a future PR
- Vision specific concerns -> future PR to neaten the vision concerns that are now scattered throughout the LM specific things
- Per-split collator configuration -> `build_collator_from_config` builds a single collator from the training split's settings, which is reused for all splits (train, validation, test). Validation/test split-specific collator settings (e.g. a different `train_target`) are silently ignored. This is pre-existing behavior not introduced by this PR, but worth noting as a known limitation. Fixing it would require refactoring the training loop to build per-split collators.

# Related issues

Fixes https://linear.app/oumi/issue/OPE-1870

# Before submitting

- [ ] This PR only changes documentation.

# Reviewers

### DIFF
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -302,12 +302,15 @@ def build_data_collator(
                 "train_target in collator_kwargs."
             )
 
+        ignore_index = kwargs.pop(
+            "ignore_index",
+            label_ignore_index if label_ignore_index is not None else -100,
+        )
+
         return TextCompletionsCollatorWithPadding(
             tokenizer=tokenizer,
             debug=debug,
-            ignore_index=(
-                label_ignore_index if label_ignore_index is not None else -100
-            ),
+            ignore_index=ignore_index,
             **kwargs,
         )
     raise ValueError(f"Unknown data collator name: '{collator_name}'")
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -115,6 +115,7 @@ def _detect_response_template(
                 f"Extracted response_template is only a <think> block.\n{_FIX_HINT}"
             )
 
+    response_template = response_template.rstrip("\n")
     return response_template
 
 
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -292,12 +292,6 @@ def build_collator_from_config(
 
     # --- TrainTarget auto-resolution ---
     if train_split.train_target is not None:
-        if collator_name != "text_completions_only_with_padding":
-            raise ValueError(
-                f"`train_target` is only supported with the "
-                f"'text_completions_only_with_padding' collator, "
-                f"got '{collator_name}'."
-            )
         if tokenizer is None:
             raise ValueError("Tokenizer is required for `train_target` auto-detection.")
         response_template, end_of_turn_template = _resolve_collator_templates(tokenizer)
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections.abc import Callable
+from dataclasses import dataclass
 
 import oumi.core.constants as constants
 from oumi.core.collators.text_collator_with_padding import TextCollatorWithPadding
@@ -27,11 +28,81 @@ from oumi.core.configs import DatasetSplit, TrainingConfig
 from oumi.core.configs.internal.supported_models import (
     find_internal_model_config,
 )
+from oumi.core.configs.params.data_params import MaskingMethod
 from oumi.core.tokenizers.base_tokenizer import BaseTokenizer
 from oumi.utils.logging import logger
 
 # This is used to set the max input length for a model with infinite size input
 _VERY_LARGE_INTEGER = int(1e30)
+
+
+# ---------------------------------------------------------------------------
+# Template auto-detection for MaskingMethod
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class _CollatorTemplates:
+    """Chat-format template strings used by the completions collator."""
+
+    response_template: str
+    end_of_turn_template: str
+
+
+_CHATML_TEMPLATES = _CollatorTemplates(
+    response_template="<|im_start|>assistant\n",
+    end_of_turn_template="<|im_end|>",
+)
+
+_LLAMA3_TEMPLATES = _CollatorTemplates(
+    response_template="<|start_header_id|>assistant<|end_header_id|>\n\n",
+    end_of_turn_template="<|eot_id|>",
+)
+
+
+# Each entry is (detector_fn, templates).  The detector receives the
+# tokenizer vocabulary (dict[str, int]) and returns True when the format
+# matches.
+_COLLATOR_TEMPLATE_DETECTORS: list[
+    tuple[Callable[[dict[str, int]], bool], _CollatorTemplates]
+] = [
+    (lambda vocab: "<|im_start|>" in vocab, _CHATML_TEMPLATES),
+    (lambda vocab: "<|start_header_id|>" in vocab, _LLAMA3_TEMPLATES),
+]
+
+
+def _resolve_collator_templates(
+    tokenizer: "BaseTokenizer",
+) -> _CollatorTemplates:
+    """Auto-detect collator templates from the tokenizer vocabulary.
+
+    Raises:
+        ValueError: If no known chat format is detected.
+    """
+    vocab: dict[str, int] = tokenizer.get_vocab()
+    for detector, templates in _COLLATOR_TEMPLATE_DETECTORS:
+        if detector(vocab):
+            return templates
+    raise ValueError(
+        "Cannot auto-detect chat template format from the tokenizer "
+        "vocabulary. Please use `collator_kwargs` to specify templates "
+        "manually instead of `masking_method`."
+    )
+
+
+def _build_masking_kwargs(
+    masking_method: MaskingMethod,
+    templates: _CollatorTemplates,
+) -> dict:
+    """Build collator keyword arguments for the given masking method."""
+    kwargs: dict = {
+        "response_template": templates.response_template,
+        "masking_method": masking_method.value,
+    }
+    if masking_method == MaskingMethod.ASSISTANT_TURN:
+        kwargs["end_of_turn_template"] = templates.end_of_turn_template
+    # FINAL_ASSISTANT_TURN needs no extra kwargs beyond response_template
+    return kwargs
 
 
 def build_data_collator(
@@ -130,7 +201,9 @@ def build_data_collator(
         if not kwargs.get("response_template"):
             raise ValueError(
                 "'text_completions_only_with_padding' requires a "
-                "response_template. Provide it via collator_kwargs."
+                "response_template. Either set train_target in your config "
+                "(which auto-resolves templates from the tokenizer) or "
+                "provide response_template via collator_kwargs."
             )
 
         return TextCompletionsCollatorWithPadding(
@@ -197,6 +270,22 @@ def build_collator_from_config(
         collator_kwargs["trust_remote_code"] = collator_kwargs.get(
             "trust_remote_code", config.model.trust_remote_code
         )
+
+    # --- MaskingMethod auto-resolution ---
+    if train_split.masking_method is not None:
+        if collator_name != "text_completions_only_with_padding":
+            raise ValueError(
+                f"`masking_method` is only supported with the "
+                f"'text_completions_only_with_padding' collator, "
+                f"got '{collator_name}'."
+            )
+        if tokenizer is None:
+            raise ValueError(
+                "Tokenizer is required for `masking_method` auto-detection."
+            )
+        templates = _resolve_collator_templates(tokenizer)
+        masking_kwargs = _build_masking_kwargs(train_split.masking_method, templates)
+        collator_kwargs.update(masking_kwargs)
 
     # Merge collator_kwargs from config with the existing kwargs
     # Config kwargs take precedence over automatically determined kwargs
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -33,6 +33,7 @@ from oumi.core.tokenizers.base_tokenizer import BaseTokenizer
 from oumi.utils.logging import logger
 
 _VERY_LARGE_INTEGER = int(1e30)
+_SENTINEL_SYS = "<<__S__>>"
 _SENTINEL_USER = "<<__U__>>"
 _SENTINEL_ASST = "<<__A__>>"
 _FIX_HINT = (
@@ -131,21 +132,28 @@ def resolve_collator_templates(
     Raises:
         ValueError: If templates cannot be extracted.
     """
-    msgs = [
+    msgs_with_sys = [
+        {"role": "system", "content": _SENTINEL_SYS},
         {"role": "user", "content": _SENTINEL_USER},
         {"role": "assistant", "content": _SENTINEL_ASST},
         {"role": "user", "content": _SENTINEL_USER},
         {"role": "assistant", "content": _SENTINEL_ASST},
     ]
+    msgs_no_sys = msgs_with_sys[1:]
 
-    try:
-        rendered = tokenizer.apply_chat_template(
-            msgs, tokenize=False, add_generation_prompt=False
-        )
-    except Exception as exc:
+    rendered = None
+    for msgs in (msgs_with_sys, msgs_no_sys):
+        try:
+            rendered = tokenizer.apply_chat_template(
+                msgs, tokenize=False, add_generation_prompt=False
+            )
+            break
+        except Exception:
+            continue
+    if rendered is None:
         raise ValueError(
             f"Tokenizer has no chat template or it failed to render.\n{_FIX_HINT}"
-        ) from exc
+        )
 
     if not isinstance(rendered, str):
         raise ValueError(
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -295,6 +295,12 @@ def build_data_collator(
                 "Fix: set train_target in your data config (auto-resolves templates "
                 "from the tokenizer), or provide response_template in collator_kwargs."
             )
+        if not kwargs.get("train_target"):
+            raise ValueError(
+                "'text_completions_only_with_padding' requires a train_target.\n"
+                "Fix: set train_target in your data config, or provide "
+                "train_target in collator_kwargs."
+            )
 
         return TextCompletionsCollatorWithPadding(
             tokenizer=tokenizer,
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -96,7 +96,7 @@ def _resolve_collator_templates(
             break
         eot_len += 1
     eot_ids = after_ids[:eot_len]
-    end_of_turn = tokenizer.decode(eot_ids, skip_special_tokens=False)
+    end_of_turn_template = tokenizer.decode(eot_ids, skip_special_tokens=False)
 
     # Response template: strip the EOT prefix to get just the assistant header.
     resp_ids = tokenizer.encode(
@@ -121,7 +121,7 @@ def _resolve_collator_templates(
             "response_template manually."
         )
 
-    return response_template, end_of_turn
+    return response_template, end_of_turn_template
 
 
 def build_data_collator(
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -111,7 +111,7 @@ def _resolve_collator_templates(
     assert isinstance(_resp_decoded, str)
     response_template = _resp_decoded
 
-    if not response_template.strip():
+    if not response_template.strip() or not end_of_turn_template.strip():
         raise ValueError(_FALLBACK_MSG)
 
     # Qwen3 and similar reasoning models inject <think>...</think> into
@@ -315,6 +315,17 @@ def build_collator_from_config(
             except ValueError:
                 if config_collator_kwargs.get("response_template") is None:
                     raise
+
+            if (
+                train_split.train_target == TrainTarget.ALL_ASSISTANT_TURNS
+                and "end_of_turn_template" not in collator_kwargs
+                and config_collator_kwargs.get("end_of_turn_template") is None
+            ):
+                raise ValueError(
+                    "train_target='all_assistant_turns' requires an "
+                    "end_of_turn_template, but auto-detection failed and "
+                    "none was provided in collator_kwargs."
+                )
 
         elif config_collator_kwargs.get("response_template") is not None:
             # Path 2: train_target not set, templates provided manually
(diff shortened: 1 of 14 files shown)


src/oumi/builders/collators.py
```diff
@@ -127,33 +208,15 @@ def build_data_collator(
...
+            raise ValueError(
+                "response_template is required for "
+                "'text_completions_only_with_padding'. Provide it via "
+                "collator_kwargs or use masking_method for auto-resolution."
             )
-
-        if not response_template:
-            response_template = "<|start_header_id|>assistant<|end_header_id|>\n\n"
-
         return TextCompletionsCollatorWithPadding(
```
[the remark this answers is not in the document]
shanghongsim: fixed with new commit

src/oumi/builders/collators.py
```diff
@@ -127,33 +208,21 @@ def build_data_collator(
...
             tokenizer=tokenizer,
-            response_template=response_template,
-            instruction_template=instruction_template,
             debug=debug,
```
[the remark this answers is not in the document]
shanghongsim: Fixed. Now, if user only specifies `collator_name` without `collator_kwargs` or `masking_method`, ity will fallback to legacy behavior with llama defaults.

src/oumi/builders/collators.py
```diff
@@ -127,33 +208,39 @@ def build_data_collator(
...
+        # span-based masking (end_of_turn_template makes instruction_template
+        # unnecessary since masking is handled by response/eot spans).
+        if kwargs.get("end_of_turn_template") is None:
+            if not kwargs.get("instruction_template"):
+                logger.warning(
+                    "No instruction_template provided for "
+                    "'text_completions_only_with_padding'. "
+                    "Falling back to Llama-style default. "
+                    "Please set instruction_template via collator_kwargs "
+                    "or use masking_method for auto-resolution."
```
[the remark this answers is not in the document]
shanghongsim: Fixed. Defaults only applied when `masking_method` is `None`

src/oumi/builders/collators.py
```diff
@@ -127,33 +208,39 @@ def build_data_collator(
...
-        if end_of_turn_template is None:
-            instruction_template = (
-                instruction_template
-                if instruction_template
-                else "<|start_header_id|>user<|end_header_id|>\n\n"
-            )
+        # span-based masking (end_of_turn_template makes instruction_template
+        # unnecessary since masking is handled by response/eot spans).
+        if kwargs.get("end_of_turn_template") is None:
+            if not kwargs.get("instruction_template"):
```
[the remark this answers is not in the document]
shanghongsim: Fixed. Defaults only applied when `masking_method` is `None`

src/oumi/builders/collators.py
```diff
@@ -127,33 +208,42 @@ def build_data_collator(
...
+                        "<|start_header_id|>user<|end_header_id|>\n\n",
+                    )
 
-        if not response_template:
-            response_template = "<|start_header_id|>assistant<|end_header_id|>\n\n"
+            if not kwargs.get("response_template"):
+                logger.warning(
+                    "No response_template provided for "
+                    "'text_completions_only_with_padding'. "
+                    "Falling back to Llama-style default. "
```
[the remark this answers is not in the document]
shanghongsim: 1. There's no path where masking_method arrives in kwargs without response_template. _build_masking_kwargs always includes it 
2. masking_method is not situated in collator_kwargs, they are on the same level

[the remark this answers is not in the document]
shanghongsim: `build_data_collator` is only access by `builc_collator_from_config` and it never calls `build_data_collator` with `masking_method` and without `response_template`. If there is `masking_method`, there will be `response_template`. This is supplied by `_build_masking_kwargs`.

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,101 @@
...
+    between_ids = tokenizer.encode(
+        rendered[first_asst_end:second_user], add_special_tokens=False
+    )
+    eot_len = 0
+    for a, b in zip(after_ids, between_ids):
+        if a != b:
+            break
+        eot_len += 1
+    eot_ids = after_ids[:eot_len]
+    end_of_turn_template: str = tokenizer.decode(eot_ids, skip_special_tokens=False)
```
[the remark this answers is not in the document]
shanghongsim: fixed

src/oumi/builders/collators.py
```diff
@@ -27,11 +27,101 @@
...
+    between_ids = tokenizer.encode(
+        rendered[first_asst_end:second_user], add_special_tokens=False
+    )
+    eot_len = 0
+    for a, b in zip(after_ids, between_ids):
+        if a != b:
+            break
+        eot_len += 1
+    eot_ids = after_ids[:eot_len]
+    end_of_turn_template = tokenizer.decode(eot_ids, skip_special_tokens=False)
```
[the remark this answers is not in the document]
shanghongsim: fixed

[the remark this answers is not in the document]
shanghongsim: fixed

src/oumi/builders/collators.py
```diff
@@ -27,11 +27,101 @@
...
+    between_ids = tokenizer.encode(
+        rendered[first_asst_end:second_user], add_special_tokens=False
+    )
+    eot_len = 0
+    for a, b in zip(after_ids, between_ids):
+        if a != b:
+            break
+        eot_len += 1
+    eot_ids = after_ids[:eot_len]
+    end_of_turn = tokenizer.decode(eot_ids, skip_special_tokens=False)
```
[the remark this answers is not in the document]
shanghongsim: fixed

src/oumi/builders/collators.py
```diff
@@ -27,11 +27,100 @@
...
+
```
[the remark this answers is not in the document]
shanghongsim: fixed

src/oumi/builders/collators.py
```diff
@@ -198,6 +290,14 @@ def build_collator_from_config(
...
+    if train_split.train_target is not None:
+        response_template, end_of_turn_template = _resolve_collator_templates(tokenizer)
+        collator_kwargs["response_template"] = response_template
+        collator_kwargs["train_target"] = train_split.train_target.value
+        if train_split.train_target == TrainTarget.ALL_ASSISTANT_TURNS:
+            collator_kwargs["end_of_turn_template"] = end_of_turn_template
+
     # Merge collator_kwargs from config with the existing kwargs
     # Config kwargs take precedence over automatically determined kwargs
     config_collator_kwargs = train_split.collator_kwargs or {}
```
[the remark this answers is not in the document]
shanghongsim: Preexisting bug, will fix in next PR

src/oumi/builders/collators.py
```diff
@@ -206,9 +291,55 @@ def build_collator_from_config(
...
+            try:
+                response_template, end_of_turn_template = _resolve_collator_templates(
+                    tokenizer
+                )
+                collator_kwargs["response_template"] = response_template
+                if train_split.train_target == TrainTarget.ALL_ASSISTANT_TURNS:
+                    collator_kwargs["end_of_turn_template"] = end_of_turn_template
+            except ValueError:
+                if config_collator_kwargs.get("response_template") is None:
+                    raise
```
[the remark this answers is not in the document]
shanghongsim: Fixed

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+    Applies the chat template to a known test conversation, then finds
+    the assistant boundary strings in the rendered output.
+
+    Returns:
+        (response_template, end_of_turn_template)
+
+    Raises:
+        ValueError: If templates cannot be extracted.
+    """
+    msgs = [
```
jgreer013, line 58: This is really clever, but we should maybe include a system instruction here?

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+    except Exception as exc:
+        raise ValueError(_FALLBACK_MSG) from exc
+
+    if not isinstance(rendered, str):
+        raise ValueError(_FALLBACK_MSG)
+
+    # Locate boundaries around the second turn pair
+    # to avoid system-prompt effects on the first turn.
+    try:
+        a1 = rendered.index(_SENTINEL_ASST)
```
jgreer013, line 79: nit: first_asst_start, and add _start to second turns

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+        first_asst_end = a1 + len(_SENTINEL_ASST)
+        second_user = rendered.index(_SENTINEL_USER, first_asst_end)
+        second_user_end = second_user + len(_SENTINEL_USER)
+        second_asst = rendered.index(_SENTINEL_ASST, second_user_end)
+        second_asst_end = second_asst + len(_SENTINEL_ASST)
+    except ValueError:
+        raise ValueError(_FALLBACK_MSG)
+
+    # End-of-turn: common token-ID prefix of the two strings that
```
jgreer013, line 88: extract this to its own method

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+    for a, b in zip(after_ids, between_ids):
+        if a != b:
+            break
+        eot_len += 1
+    eot_ids = after_ids[:eot_len]
+    _eot_decoded = tokenizer.decode(eot_ids, skip_special_tokens=False)
+    assert isinstance(_eot_decoded, str)
+    end_of_turn_template = _eot_decoded
+
+    # Response template: strip the EOT prefix to get just the assistant header.
```
jgreer013, line 104: Same here, extract to its own method

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+    if eot_len > 0 and resp_ids[:eot_len] == eot_ids:
+        resp_ids = resp_ids[eot_len:]
+    _resp_decoded = tokenizer.decode(resp_ids, skip_special_tokens=False)
+    assert isinstance(_resp_decoded, str)
+    response_template = _resp_decoded
+
+    if not response_template.strip() or not end_of_turn_template.strip():
+        raise ValueError(_FALLBACK_MSG)
+
+    # Qwen3 and similar reasoning models inject <think>...</think> into
```
jgreer013, line 117: This description scares me, I wonder if we could have a better workaround or a louder error

src/oumi/core/collators/trl_data_collator_for_completion_only_lm.py
```diff
@@ -19,18 +19,64 @@
 import torch
 from transformers.data.data_collator import DataCollatorForLanguageModeling
 
+from oumi.core.configs.params.data_params import TrainTarget
+
 
 class DataCollatorForCompletionOnlyLM(DataCollatorForLanguageModeling):
```
jgreer013, line 25: It's not clear to me what this collator does differently than the other one.

jgreer013 (approved):

src/oumi/builders/collators.py
shanghongsim: Like modify it to extract system instruction? We are current no longer relying on system instruction for masking

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,114 @@
...
 _VERY_LARGE_INTEGER = int(1e30)
+_SENTINEL_USER = "<<__U__>>"
+_SENTINEL_ASST = "<<__A__>>"
+_FIX_HINT = (
+    "Fix: provide response_template (and end_of_turn_template for "
+    "all_assistant_turns) in collator_kwargs."
+)
+
+
+def _resolve_collator_templates(
```
lefft, line 44: Has this been tested with gpt-oss models? They use a non-standard system for marking boundaries within responses, so wondering if boundary detection will work as it does for Qwen/Llama style templates.

src/oumi/builders/collators.py
```diff
@@ -206,9 +303,66 @@ def build_collator_from_config(
...
+            # Path 2: train_target not set, templates provided manually
+            # via collator_kwargs. Infer train_target from which templates
+            # are present.
+            has_eot = config_collator_kwargs.get("end_of_turn_template") is not None
+            has_inst = config_collator_kwargs.get("instruction_template") is not None
+            if has_eot:
+                collator_kwargs["train_target"] = "all_assistant_turns"
+            elif has_inst:
+                warnings.warn(
+                    "Instruction-based masking is deprecated.\n"
```
lefft, line 348: Is there a plan/intention to eliminate support for instruction-based masking altogether, or will we just keep it in a "deprecated" state? 

Want to make sure we retain compatibility with existing enterprise configs (or verify that moving to this style doesn't change training dynamics before adapting enterprise configs).

lefft (approved):
Can we be sure to check whether template detection and collation happens properly when user or assistant turns begin with (one or more) leading `\n`s? We had issues with this breaking tokenization and leading to records being skipped in enterprise, would like to understand whether these changes will help, harm, or not impact that issue.

src/oumi/builders/collators.py
jgreer013: No I mean ensure that when we extract the different template portions, we do so using an example conversation that also includes a system instruction

Or rather, perhaps do it twice, once on a conversation with a system instruction, and once on a conversation without.

Why? Because I'm a bit concerned of the scenario where this logic doesn't use system instructions when extracting, but the user does in their data, and *somehow* the resulting templates we extract using this methodology don't wind up working when the data includes the presence of a system instruction for some reason.

src/oumi/core/collators/trl_data_collator_for_completion_only_lm.py
shanghongsim: Previously, it would `instruction_template` and `response_template` pairs then mask everything in between (see L324). However, there isn't an `instruction_template` before the tool result, so tool result content does not get masked properly. Some models (like Qwen) *accidentally* mask it correctly because their chat template stores tool result under role.user. But for most models (like Llama etc), tool result is stored under a role.tool with a separate tag so masking is incorrectly applied for the reasons above. The new approach basically masks everything then only unmask portions between the `response_template` and `end_of_turn_template`. This is more robust to the different roles and specifics of how tool results are formatted and stored by different chat templates.

image

oelachqar (approved):

tests/integration/builders/test_collator_template_detection.py
```diff
@@ -0,0 +1,240 @@
...
+) -> transformers.PreTrainedTokenizerBase:
+    tok = transformers.AutoTokenizer.from_pretrained(
+        model_name, trust_remote_code=trust_remote_code
+    )
+    if not tok.chat_template:
+        try:
+            jinja_path = hf_hub_download(model_name, "chat_template.jinja")
+            with open(jinja_path) as f:
+                tok.chat_template = f.read()
+        except Exception:
```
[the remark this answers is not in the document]
shanghongsim: This has been fixed

src/oumi/builders/collators.py
shanghongsim: Added code to strip the <think> in `_detect_response_template`

src/oumi/builders/collators.py
shanghongsim: Done!

src/oumi/builders/collators.py
shanghongsim: Done!

src/oumi/builders/collators.py
shanghongsim: Done!

src/oumi/builders/collators.py
shanghongsim: Yes, tested. There is an integration test in tests/integration/builders/test_collator_template_detection.py

src/oumi/builders/collators.py
shanghongsim: Plan is to eliminate instruction-based masking altogether. In the interim, while we do more testing, it will be in deprecated state.

src/oumi/builders/collators.py
shanghongsim: Ok, added!

shanghongsim:
> Can we be sure to check whether template detection and collation happens properly when user or assistant turns begin with (one or more) leading `\n`s? We had issues with this breaking tokenization and leading to records being skipped in enterprise, would like to understand whether these changes will help, harm, or not impact that issue.

- Added `.rstrip("\n")` in `_detect_response_template` (collators.py:119)
- Added two tests to verify correct behavior with leading `\n`: `test_leading_newline_bpe_merge_regression` and `test_span_masking_with_leading_newline_content`

cc @lefft

src/oumi/builders/collators.py
```diff
@@ -27,11 +28,105 @@
...
+        rendered[second_user_end:second_asst], add_special_tokens=False
+    )
+    if eot_len > 0 and resp_ids[:eot_len] == eot_ids:
+        resp_ids = resp_ids[eot_len:]
+    _resp_decoded = tokenizer.decode(resp_ids, skip_special_tokens=False)
+    assert isinstance(_resp_decoded, str)
+    response_template = _resp_decoded
+
+    if not response_template.strip() or not end_of_turn_template.strip():
+        raise ValueError(_FALLBACK_MSG)
```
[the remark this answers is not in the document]
shanghongsim: We should be able to extract `end_of_turn_template` even when `train_target` is `FINAL_ASSISTANT_TURN`

src/oumi/builders/collators.py
```diff
@@ -126,27 +280,19 @@ def build_data_collator(
...
         return TextCompletionsCollatorWithPadding(
             tokenizer=tokenizer,
-            instruction_prefix=instruction_prefix,
-            response_prefix=response_prefix,
             debug=debug,
+            ignore_index=(
+                label_ignore_index if label_ignore_index is not None else -100
+            ),
             **kwargs,
         )
```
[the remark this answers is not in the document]
shanghongsim: Fixed

src/oumi/builders/collators.py
```diff
@@ -126,27 +231,19 @@ def build_data_collator(
...
     raise ValueError(f"Unknown data collator name: '{collator_name}'")
```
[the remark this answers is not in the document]
shanghongsim: Fixed

PR #2376, opened 2026-04-17
refactor(analyze): split AnalyzerConfig into type/id/display_name

ryan-arman:
# Description

Part of the **analyze v2** split of #2370 (Phase 2 of 4).

Reshapes `AnalyzerConfig` into three fields with distinct responsibilities:

- `type` — analyzer kind (required, was `id` pre-#2376)
- `id` — stable identity; canonical key for results, caches, and test metric paths. Defaults to `display_name`.
- `display_name` — human-readable label. Defaults to `type`. Was `instance_id` pre-#2376.

Today `id` always defaults to `display_name`, so YAMLs that only set `display_name` keep working unchanged. The split exists so that in the future the API can mint stable asset ids for analyzers while `display_name` remains purely cosmetic (and need not be unique).

Key behavior changes:

- Uniqueness is now enforced on `id`, not `display_name`. Two analyzers may share a `display_name` as long as their `id` values differ.
- Test metric paths use `id`: `"{id}.{field_name}"`. When `id` is omitted, this still equals `"{display_name}.{field_name}"`.
- Legacy `instance_id` YAML key still maps to `display_name` with a `DeprecationWarning`.
- Pre-#2376 YAMLs that used `id` in place of `type` now fail loudly with "type is required" — `id` has new semantics.

Also:

- Adds an `AnalyzerType` enum for the built-in analyzers (`length`, `quality`, `turn_stats`).
- Updates CLI, example YAML, and analyze docs to the new shape.
- Expands tests to cover the id/display_name split, duplicate-id detection, and the legacy `instance_id` alias.

Design doc: `docs/superpowers/specs/2026-04-17-analyzer-config-id-design.md` (on `main`).

**Dependencies:** Stacked on #2375 (Phase 1 — CLI migration). Rebase to `main` once that lands.

## Related issues

Linear Issue: [OPE-1868](https://linear.app/oumi-team/issue/OPE-1868)
Towards OPE-1868, OPE-1869

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
- [ ] Did you read the [contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?
- [x] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2375, opened 2026-04-17
feat(analyze): migrate analyze CLI to typed analyzer framework

ryan-arman:
# Description

Migrates `oumi analyze` from the legacy `DatasetAnalyzer` onto the typed framework (`TypedAnalyzeConfig` + `AnalysisPipeline` + `TestEngine`) that already lives in `src/oumi/analyze/`. First PR of a 4-PR split of #2370.

**What changed**

- Rewrite `src/oumi/cli/analyze.py` on top of the typed framework. Delegates analyzer construction to `oumi.analyze.create_analyzer_from_config` instead of keeping its own duplicate.
- Restore CLI flags: `--config`, `--list`, `--list-metrics`, `--log-level`, `--output`, `--format`, `--verbose`, plus `--dataset-path` / `--dataset-name` / `--sample-count` (snake_case forms accepted as aliases).
- Emit a friendly migration error when a user passes a legacy v1 `AnalyzeConfig` YAML (detected via `dataset_source`, `processor_name`, `is_multimodal`, etc.).
- Wrap analyze as a nested Typer app in `src/oumi/cli/main.py` so subcommands and help panels render.
- Replace `configs/examples/analyze/analyze.yaml` with the three-analyzer, five-test default recipe used by the API's auto-analysis flow.
- Rewrite `docs/user_guides/analyze/{analyze,analyze_config}.md` to match the new shape; exclude the v2 example from the legacy-config sweep in `tests/unit/core/configs/test_parse_configs.py`.

**Out of scope (follow-up PRs)**

- `type`/`display_name` rename in `AnalyzerConfig` → #2376
- `testing/engine.py` refactor + index / MRO fixes → #2377
- `analyze/{base,pipeline,discovery,utils}` cleanup → #2378

**Verification**

- `pytest tests/unit/cli/test_cli_analyze.py` — 7/7 pass.
- `pytest tests/unit/core/configs/test_parse_configs.py` — 984/984 pass.
- Smoke test: `oumi analyze --config configs/examples/analyze/analyze.yaml --dataset-path <local.jsonl>` runs end-to-end and writes `analysis.csv`, `test_results.json`, `summary.json`.

Usage example:
analyze_cli_usage

Using the old config (errors):
analyze_cli_old_config

## Related issues

Linear Issue: OPE-1868
Fixes OPE-1868

## Before submitting

- [ ] This PR only changes documentation. (You can ignore the following checks in that case)
[...]

PR #2377, opened 2026-04-17
fix(analyze): preserve sample indices and fix threshold truthiness

ryan-arman:
# Description

Part of the **analyze v2** split of #2370 (Phase 3 of 4). Standalone fixes — not dependent on PRs #2375 / #2376.

Two related fixes to the analyze test engine.

## 1. Preserve original sample indices

`TestEngine._extract_metric_values` now returns `(original_index, value)` pairs so that `sample_indices` and `all_affected_indices` on the resulting `TestResult` point to real conversation positions in the dataset.

**Why a sample can be "missing" a metric** (i.e., why `_get_nested_value` returns `None`):

- The analyzer's result model declares the field as optional and leaves it `None` for that sample (e.g., couldn't compute on an empty conversation, or a conditional code path that doesn't produce the field).
- The metric lives in a `CustomMetricResult.values` dict whose keys vary per sample (custom metric functions are user-written and may emit different keys per input).
- A nested dict path has a missing intermediate key.

**The bug.** Previously, the code enumerated `analyzer_results` and dropped Nones, then re-numbered the survivors `0..N-1`. With 100 samples where #5 and #42 lacked the metric, indices `5..98` in the result list pointed to dataset samples `6..99`. Any downstream "drop these flagged rows" tooling consuming `all_affected_indices` deleted the wrong rows.

## 2. Fix threshold truthiness bug

The result-builder used:

```python
threshold=test.max_percentage or test.min_percentage,
```

`or` returns the first truthy operand, and **`0.0` is falsy in Python**. So a "zero tolerance" rule like:

```yaml
- id: no_pii_allowed
  type: threshold
  metric: pii.count
  operator: ">"
  value: 0
  max_percentage: 0.0   # fail if ANY sample has pii.count > 0
```

…produced `0.0 or None` → `None` on the `TestResult.threshold` field. The CLI table, `--output-json`, and any downstream tool reading the result then displayed `threshold: —` instead of `0.0`, hiding the user's actual policy.

The "both set" variant is sneakier: with `max_percentage=0.0` and `min_percentage=50.0`, `0.0 or 50.0` → `50.0` — confidently misattributing the threshold to the min rule when the max rule fired.

The pass/fail decision itself wasn't broken (that path uses explicit `is not None` checks). What was broken was the *reported* threshold. Replaced with explicit `is not None` in both the in-memory `TestEngine` and the incremental `BatchTestEngine`:

```python
threshold=(
    test.max_percentage
    if test.max_percentage is not None
    else test.min_percentage
),
```

## Tests

`tests/unit/analyze/test_testing_engine.py` adds three regression tests with docstrings that quote the buggy code so the intent survives future rewrites:

- `test_sample_indices_map_to_original_positions_when_values_are_missing`
- `test_zero_tolerance_threshold_is_reported_as_zero_not_none`
- `test_zero_tolerance_threshold_picks_max_when_both_are_set`

## Related issues

Linear Issue: [OPE-1868](https://linear.app/oumi-team/issue/OPE-1868)
Towards OPE-1868

## Before submitting

### DIFF
--- a/src/oumi/analyze/testing/batch_engine.py
+++ b/src/oumi/analyze/testing/batch_engine.py
@@ -289,7 +289,11 @@ class BatchTestEngine:
             affected_count=len(affected_ids),
             total_count=acc.total_count,
             affected_percentage=round(affected_pct, 2),
-            threshold=test.max_percentage or test.min_percentage,
+            threshold=(
+                test.max_percentage
+                if test.max_percentage is not None
+                else test.min_percentage
+            ),
             actual_value=None,
             sample_indices=[],  # Not meaningful for batch mode
             all_affected_indices=[],  # Not meaningful for batch mode
--- a/src/oumi/analyze/testing/engine.py
+++ b/src/oumi/analyze/testing/engine.py
@@ -113,7 +113,11 @@ class TestEngine:
             affected_count=len(affected_indices),
             total_count=total_count,
             affected_percentage=round(affected_pct, 2),
-            threshold=test.max_percentage or test.min_percentage,
+            threshold=(
+                test.max_percentage
+                if test.max_percentage is not None
+                else test.min_percentage
+            ),
             actual_value=actual_value,
             sample_indices=affected_indices[:MAX_SAMPLE_INDICES],
             all_affected_indices=affected_indices,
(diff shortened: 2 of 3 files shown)


ryan-arman:
Adding context on the sample-index preservation fix in response to a review question.

**Why a sample can be "missing" a metric** (i.e., why `_get_nested_value` returns `None`):

1. The analyzer's result model declares the field as optional and leaves it `None` for that sample — e.g., couldn't compute on an empty conversation, or a conditional code path that doesn't produce the field.
2. The metric lives in a `CustomMetricResult.values` dict whose keys vary per sample (custom metric functions are user-written and may emit different keys per input).
3. A nested dict path has a missing intermediate key.

Before this fix, those samples were dropped from the iteration but the surviving values were re-numbered 0..N-1, so `sample_indices` and `all_affected_indices` on the resulting `TestResult` no longer pointed at real dataset rows. With 100 samples where #5 and #42 are missing the metric, indices 5..98 in the result list pointed to samples 6..99 in the dataset — which broke any "drop these rows" downstream use case.

Pushed `d9b714714` expanding the `_extract_metric_values` docstring to make this explicit.

oelachqar (approved):

PR #2378, opened 2026-04-17
fix(analyze): handle cached dicts in dataframe conversion

ryan-arman:
# Description

Dataset-level cached results are raw dicts (loaded from the pipeline's JSON cache), not Pydantic instances. `to_analysis_dataframe` gated the dataset-level branch on `isinstance(result, BaseModel)`, so cached dataset metrics silently disappeared from the returned DataFrame. `results_to_dict` had the same blind spot on the list branch.

Both entry points now accept `dict` alongside `BaseModel`, and the type signatures are broadened to reflect that. Includes regression tests for dataset-level dicts, per-conversation list-of-dicts, and the `results_to_dict` passthrough.

Part of a 4-PR split of #2370. Standalone — does not depend on the other PRs.

## Related issues

Towards OPE-1868

## Before submitting

### DIFF
--- a/src/oumi/analyze/utils/dataframe.py
+++ b/src/oumi/analyze/utils/dataframe.py
@@ -28,7 +28,10 @@ logger = logging.getLogger(__name__)
 
 def to_analysis_dataframe(
     conversations: list[Conversation],
-    results: Mapping[str, Sequence[BaseModel] | BaseModel],
+    results: Mapping[
+        str,
+        Sequence[BaseModel | dict[str, Any]] | BaseModel | dict[str, Any],
+    ],
     message_to_conversation_idx: list[int] | None = None,
 ) -> pd.DataFrame:
     """Convert typed analysis results to a pandas DataFrame.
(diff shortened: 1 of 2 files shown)


ryan-arman:
Closing as YAGNI.

The fix is correct for the bug it describes, but the bug isn't reachable in main today:

1. No `DatasetAnalyzer` exists in the deployed codebase — only `ConversationAnalyzer`s (`LengthAnalyzer`, `DataQualityAnalyzer`, `TurnStatsAnalyzer`), which produce `list[BaseModel]`. The dataset-level branch (`BaseModel | dict`, single result) has no real producer.
2. The list branch already handles cached dicts via `_add_result_to_row`'s existing `isinstance(result, dict)` duck-typing.
3. No caller pipes `pipeline.load_cache()` into `pipeline.to_dataframe()` — the CLI doesn't reload from cache before producing a dataframe.

Will revisit when the first real `DatasetAnalyzer` or `load_cache → to_dataframe` caller lands and exercises this path for real.

PR #2392, opened 2026-04-24
Support tool definitions and structured tool calls in SFT pipeline

shanghongsim:
# Description

Adds first-class support for OpenAI-format tool definitions and structured tool calls in the Oumi `Conversation` / `Message` schema, so TRL's `SFTTrainer` can forward them to `tokenizer.apply_chat_template(messages, tools=...)`. The target model's chat template then renders tools in its native format (Qwen `<tools>` XML, Llama JSON, Mistral `[TOOL_CALLS]`, etc.), which lets the **same training data be portable across tool-capable models** instead of being locked to one pre-baked format.

## Motivation

Today, training on tool-call data in Oumi requires manual pre-formatting tool definitions and tool calls into message `content` as text. This is very error prone as users will need to understand the idiosyncrasies of each model's chat template.

After this change, users will store a universal format (JSONL row can carry OpenAI-wire-format structured tools and calls). This will be formatted into the model's native representation string by `apply_chat_template`. Flow: `TextSftJsonLinesDataset` with `return_conversations=True, format="dict"` → TRL `SFTTrainer` → `apply_chat_template(messages, tools=example["tools"])`

```json
{
  "messages": [
    {"role": "user", "content": "Weather in SF?"},
    {"role": "assistant", "content": null,
     "tool_calls": [{"id": "call_abc", "type": "function",
                     "function": {"name": "get_weather", "arguments": "{\"city\": \"SF\"}"}}]},
    {"role": "tool", "content": "{\"temp\": 65}", "tool_call_id": "call_abc"},
    {"role": "assistant", "content": "It's 65 in SF."}
  ],
  "tools": [{"type": "function", "function": {"name": "get_weather", "...": "..."}}]
}
```

## What changes

**`src/oumi/core/types/conversation.py`**
- `Conversation.tools: list[dict | Callable] | None` — new top-level field. Accepts OpenAI-format dicts; callables are coerced to dicts at validation time via `transformers.utils.chat_template_utils.get_json_schema` (Pydantic `field_validator(mode="before")`). After validation every entry is a dict, so persistence is uniform.
- `Message.tool_calls: list[dict] | None` — structured calls on assistant messages.
- `Message.tool_call_id: str | None` — links tool responses to the call they answer.
- `Message.content` is now `str | list[ContentItem] | None`; a message is valid when it has at least one of `content` or `tool_calls`.
- `Conversation.to_dict()` post-processes the dump so `content: null` is preserved on messages with `tool_calls`. Without this, `exclude_none=True` drops the key and some chat templates choke on the missing field.

**`src/oumi/core/datasets/base_sft_dataset.py`**
- `transform()` with `return_conversations=True, format="dict"` always emits a top-level `tools` key (None when absent) so HF Dataset schema stays uniform when mixing tool and non-tool rows.

**Defensive updates for `content=None`**
- `src/oumi/analyze/base.py`, `src/oumi/analyze/utils/dataframe.py`, `src/oumi/utils/conversation_utils.py`: return empty string on `None` content in text-extraction helpers rather than erroring.

## Backward compatibility

This change does not break older data formats, `to_dict()` consumers, and `Conversation`/`Message` constructions and is backward compatible.  

| Concern | Status | Why |
|---|---|---|
| Old JSONL loads correctly | ✅ | `Conversation.model_validate` populates known fields; missing `tools`/`tool_calls`/`tool_call_id` default to `None`. |
| Old `to_dict()` output shape preserved | ✅ | `exclude_none=True` strips the new optional fields when they're `None`. The post-process loop only restores `content: null` when `msg.tool_calls` is truthy, so plain messages are untouched. |
| Old `Conversation(messages=[…])` / `Message(role=…, content=…)` constructions | ✅ | New fields are all `Optional` with `None` defaults. `model_post_init` requires *one of* `content` or `tool_calls`, which old messages always satisfy because they always have content. |
| Old code passing `list[dict]` as `Conversation.tools` | ✅ | Validator's first branch is `if isinstance(tool, dict): pass through`. No transformation. |
| Old TRL training pipelines | ✅ | Per-row HF dataset gains a `tools: null` column for non-tool rows; `example.get("tools")` returns `None`; `apply_chat_template` is called without tools → identical rendering to before. |
| Type-hint compatibility | ✅ | `Message.content` widened from `str \| list[ContentItem]` to `str \| list[ContentItem] \| None`. Existing callers pass valid subtypes. We also widened `Conversation.tools` to `list[dict \| Callable] \| None`; old `list[dict]` callers still satisfy. |

Concrete legacy-format tests in the suite that we kept passing unmodified:
- `test_conversation_to_dict_legacy`
- `test_conversation_from_dict_legacy`
- `test_roundtrip_dict_legacy`
- `test_roundtrip_json_legacy`
- `test_conversation_to_dict_plain_messages_have_no_extra_null_keys`

The four test files we *did* modify needed only one-line `isinstance` / None narrowing because pyright (correctly) tightened type inference once `Message.content` became `Optional`. None of them changed behavior.

**Test runs to validate compatibility**:
- `tests/unit/core/types/test_conversation.py`: 72 passed (57 pre-existing + 15 new)
- Broader unit suite (`tests/unit/core/`, `tests/unit/utils/`, `tests/unit/datasets/`, `tests/unit/analyze/`): 2371 passed, 0 failed.

## A note on data heterogeneity (column vs list-item level)

Why does the pipeline need `setdefault("tools", None)` to make the top-level `tools` column uniform across rows, but **not** force `tool_calls: null` on plain messages?

This is forced by **HuggingFace Dataset's schema inference rules**, which are different at the column level vs the list-item level:

| Level | Schema mode | Heterogeneity allowed? |
|---|---|---|
| Top-level columns | Strict arrow struct — one schema per column, valid for every row | **No** — column must exist in every row with the same type |
| Items inside a `List(Json)` column | Dynamic JSON — each item is a JSON-serializable dict, no fixed inner schema | **Yes** — each item can have different keys |

**Practical implication for users**: a single `Conversation` can have a mix of message shapes (system, user, plain assistant, tool-call assistant, tool response, plain assistant, …). All within one row. The pipeline handles it natively — verified end-to-end with the glaive multi-turn split where assistant messages alternate between tool-call and text within the same conversation.

## What's out of scope

- **Oumi-direct tokenization path** (`BaseSftDataset.tokenize()`) does not yet pass `tools=` to `apply_chat_template`. This PR only wires the TRL_SFT + `return_conversations=True` path end-to-end. Can follow up.
- **TRL_GRPO and other trainers** similarly unchanged.
- **Format conversion** is NOT done inside Oumi — users supply data already in OpenAI wire format. Callable inputs are normalized to OpenAI-format dicts via `get_json_schema`; no other coercion.

## Verification done

| # | Check | Result |
|---|---|---|
| 1 | **Unit tests** — 15 new + entire pre-existing suite | **2371 passed, 0 failed** |
| 2 | **Byte-diff** of `to_dict()` on 100 real legacy rows | **Single difference**: `tools: null` added at top level. Every other byte is identical. |
| 3 | **50-step smoke training** on legacy data, branch | **Converges normally** (loss 1.062 → 0.812). No errors. |
| 4 | **Side-by-side parity** vs main, identical seeds | **Max loss diff = 0.0004** across 11 logged steps. Grad-norm parity. |

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/analyze/utils/dataframe.py
+++ b/src/oumi/analyze/utils/dataframe.py
@@ -21,7 +21,8 @@ from typing import Any
 import pandas as pd
 from pydantic import BaseModel
 
-from oumi.core.types.conversation import ContentItem, Conversation
+from oumi.analyze.base import BaseAnalyzer
+from oumi.core.types.conversation import Conversation
 
 logger = logging.getLogger(__name__)
 
@@ -155,16 +156,7 @@ def to_message_dataframe(
                 "role": message.role.value,
             }
 
-            if isinstance(message.content, str):
-                row["text_content"] = message.content
-            elif message.content is None:
-                row["text_content"] = ""
-            else:
-                text_parts = []
-                for item in message.content:
-                    if isinstance(item, ContentItem) and isinstance(item.content, str):
-                        text_parts.append(item.content)
-                row["text_content"] = " ".join(text_parts)
+            row["text_content"] = BaseAnalyzer.get_text_content(message)
 
             # Add results from each message-level analyzer
             for analyzer_name, analyzer_results in results.items():
--- a/src/oumi/analyze/utils/dataframe.py
+++ b/src/oumi/analyze/utils/dataframe.py
@@ -157,6 +157,8 @@ def to_message_dataframe(
 
             if isinstance(message.content, str):
                 row["text_content"] = message.content
+            elif message.content is None:
+                row["text_content"] = ""
             else:
                 text_parts = []
                 for item in message.content:
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -162,11 +162,18 @@ class BaseSftDataset(BaseMapDataset, ABC):
                 return {"conversation_json": conversation_json}
             elif self._return_conversations_format == "dict":
                 data = conversation.to_dict()
-                # Ensure `tools` is always present as a top-level key so the
-                # HF dataset schema stays consistent across rows. Without this,
-                # rows whose Conversation has no tools end up missing the key
-                # (exclude_none=True strips it), and `datasets.Dataset.from_generator`
-                # fails with KeyError when mixing tool and non-tool rows.
+                # HF Datasets infers a single arrow struct schema across rows
+                # and requires every top-level key to exist on every row with
+                # the same type. Without this, rows whose Conversation has no
+                # tools end up missing the key (`exclude_none=True` strips it),
+                # and `datasets.Dataset.from_generator` fails with KeyError
+                # when mixing tool and non-tool rows.
+                #
+                # Note that this rule applies at the column level only —
+                # inside a list-of-dicts column like `messages`, items can
+                # have heterogeneous keys (e.g., assistant messages with
+                # `tool_calls`, tool messages with `tool_call_id`, plain
+                # messages with neither), and that's fine.
                 data.setdefault("tools", None)
                 return data
             else:
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -161,7 +161,14 @@ class BaseSftDataset(BaseMapDataset, ABC):
                 conversation_json = conversation.to_json()
                 return {"conversation_json": conversation_json}
             elif self._return_conversations_format == "dict":
-                return conversation.to_dict()
+                data = conversation.to_dict()
+                # Ensure `tools` is always present as a top-level key so the
+                # HF dataset schema stays consistent across rows. Without this,
+                # rows whose Conversation has no tools end up missing the key
+                # (exclude_none=True strips it), and `datasets.Dataset.from_generator`
+                # fails with KeyError when mixing tool and non-tool rows.
+                data.setdefault("tools", None)
+                return data
             else:
                 raise ValueError(
                     f"Invalid return_conversations_format: "
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -161,21 +161,7 @@ class BaseSftDataset(BaseMapDataset, ABC):
                 conversation_json = conversation.to_json()
                 return {"conversation_json": conversation_json}
             elif self._return_conversations_format == "dict":
-                data = conversation.to_dict()
-                # HF Datasets infers a single arrow struct schema across rows
-                # and requires every top-level key to exist on every row with
-                # the same type. Without this, rows whose Conversation has no
-                # tools end up missing the key (`exclude_none=True` strips it),
-                # and `datasets.Dataset.from_generator` fails with KeyError
-                # when mixing tool and non-tool rows.
-                #
-                # Note that this rule applies at the column level only —
-                # inside a list-of-dicts column like `messages`, items can
-                # have heterogeneous keys (e.g., assistant messages with
-                # `tool_calls`, tool messages with `tool_call_id`, plain
-                # messages with neither), and that's fine.
-                data.setdefault("tools", None)
-                return data
+                return self._conversation_to_hf_compatible_dict(conversation)
             else:
                 raise ValueError(
                     f"Invalid return_conversations_format: "
@@ -183,6 +169,25 @@ class BaseSftDataset(BaseMapDataset, ABC):
                     "Supported formats are 'json' and 'dict'."
                 )
         return self.tokenize(conversation)
+
+    def _conversation_to_hf_compatible_dict(self, conversation: Conversation) -> dict:
+        """Convert a Conversation to a dict with an HF-Datasets-stable schema.
+
+        HF Datasets infers a single arrow struct schema across rows and
+        requires every top-level key to exist on every row with the same
+        type. Without forcing `tools` to be present, rows whose Conversation
+        has no tools end up missing the key (`exclude_none=True` strips it),
+        and `datasets.Dataset.from_generator` fails with KeyError when mixing
+        tool and non-tool rows.
+
+        This rule applies at the column level only — inside a list-of-dicts
+        column like `messages`, items can have heterogeneous keys (e.g.,
+        assistant messages with `tool_calls`, tool messages with
+        `tool_call_id`, plain messages with neither), and that's fine.
+        """
+        data = conversation.to_dict()
+        data.setdefault("tools", None)
+        return data
 
     def tokenize(
         self,
--- a/src/oumi/core/types/conversation.py
+++ b/src/oumi/core/types/conversation.py
@@ -537,10 +537,18 @@ class Conversation(pydantic.BaseModel):
         data = self.model_dump(
             mode="json", exclude_unset=True, exclude_defaults=False, exclude_none=True
         )
-        # Preserve `content: null` on assistant messages that only carry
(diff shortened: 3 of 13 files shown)


src/oumi/analyze/utils/dataframe.py
```diff
@@ -157,6 +157,8 @@ def to_message_dataframe(
 
             if isinstance(message.content, str):
                 row["text_content"] = message.content
+            elif message.content is None:
```
lefft: Minor: If `content` is null but there are tool calls, do we not count tool turns toward analyses like token counts? Should we? (e.g. could include a stringified form of the tool fields so that they get counted)

lefft (approved):

src/oumi/core/types/conversation.py
```diff
@@ -228,31 +229,57 @@ class Message(pydantic.BaseModel):
...
     For multimodal messages, `content` should be a list of content items of
     potentially different types e.g., text and image.
+    May be ``None`` on assistant messages that only contain ``tool_calls``
+    (OpenAI tool-calling wire format).
     """
 
     role: Role
     """The role of the entity sending the message (e.g., user, assistant, system)."""
 
+    tool_calls: list[dict] | None = None
```
wizeng23: Have you considered having tool calls be a subclass of ContentItem instead? What's the advantages of your current approach?

src/oumi/core/types/conversation.py
```diff
@@ -228,31 +229,57 @@ class Message(pydantic.BaseModel):
         Optional[str]: The unique identifier of the message, if set; otherwise None.
     """
 
-    content: str | list[ContentItem]
+    content: str | list[ContentItem] | None = None
```
wizeng23, line 234: Is it ever possible have content not be None when tool_calls is present? Or are tool calls usually in a separate Message from the assistance response of the same turn?

src/oumi/core/types/conversation.py
wizeng23: Is it possible to have Tool calls be a dataclass instead of dict? If we expect certain keys (id, type, etc.), it's better to have these codified instead of having arbitrary keys.

src/oumi/core/types/conversation.py
```diff
@@ -383,6 +410,55 @@ class Conversation(pydantic.BaseModel):
     in a key-value format. It can be used to include any relevant contextual data.
     """
 
+    tools: list[dict | Callable] | None = None
```
wizeng23: ditto, can the dict be a Dataclass?

src/oumi/core/types/conversation.py
shanghongsim: Yes, possible to have content be `None` if `tool_calls` are present. Say the model just directly responds with a tool call. Raw output from the model will be text. The model specific tool call parser will then extract the tool call and place it in this field. If the model responds with some preamble or explanation or thinking before the tool call, the preamble will remain in content while the tool call will be extracted out to `tool_call`. See example data from https://huggingface.co/datasets/interstellarninja/hermes_reasoning_tool_use

```
{
  "role": "assistant",
  "content": null,
  "tool_calls": [
    {
      "id": "call_2_0",
      "type": "function",
      "function": {
        "name": "create_feedback_form",
        "arguments": {
          "fields": [
            "Employee Name",
            "Department",
            "Feedback Type",
            "Comments"
          ]
        }
      }
    },
    {
      "id": "call_2_1",
      "type": "function",
      "function": {
        "name": "get_all_feedback",
        "arguments": {
          "sort_by": "Department"
        }
      }
    }
  ]
}
```

src/oumi/core/types/conversation.py
shanghongsim: Done! There was a change going out to make this into pydantic dataclasses while the review was being done

src/oumi/core/types/conversation.py
shanghongsim: same as above!

src/oumi/core/types/conversation.py
shanghongsim: I went for with a sibling field primarily for wire-format parity with HF chat templates / TRL `SFTTrainer`, which all expect `tool_calls` as a top-level message key. The key disadvantage, implementation wise, of putting it inside content would be forcing a flatten/unflatten on every render and dataset round-trip. Also, on a logical front, we do not expect the tool set to be changing between turns in a single conversation, hence it makes sense for it to live outside of the conversation turn level.

src/oumi/analyze/utils/dataframe.py
shanghongsim: Good point. I added a static method is base Analyzer (`get_text_content`) to stringify the entire message (content and tool calls). Since the repsentation is model agnostic (no chat template is involved), the token-count analyses is approximate. This is noted in the code. We could implement a more complex system with chat template when the need arises for it.

src/oumi/core/datasets/base_sft_dataset.py
```diff
@@ -161,7 +161,21 @@ def transform(self, sample: pd.Series) -> dict:
                 conversation_json = conversation.to_json()
                 return {"conversation_json": conversation_json}
             elif self._return_conversations_format == "dict":
-                return conversation.to_dict()
+                data = conversation.to_dict()
+                # HF Datasets infers a single arrow struct schema across rows
```
jgreer013: If a single line of code needs a comment this long, it's probably worth it to move it out into its own method.

src/oumi/core/types/tool_call.py
```diff
@@ -0,0 +1,129 @@
...
+    length of 64.
+    """
+
+    description: str | None = None
+    """A description of what the function does.
+
+    Used by the model to choose when and how to call the function.
+    """
+
+    parameters: dict[str, Any] | None = None
```
jgreer013: If this var is strictly a JSON Schema object, we should have the type match that instead of just a generic dict

src/oumi/core/types/conversation.py
```diff
@@ -452,9 +534,22 @@ def filter_messages(
 
     def to_dict(self):
         """Converts the conversation to a dictionary."""
-        return self.model_dump(
+        data = self.model_dump(
             mode="json", exclude_unset=True, exclude_defaults=False, exclude_none=True
         )
+        # Preserve `content: null` on assistant messages that only carry
```
jgreer013: Same here - very long comment means we should probably extract this to a method

src/oumi/core/types/conversation.py
```diff
@@ -383,6 +416,55 @@ class Conversation(pydantic.BaseModel):
     in a key-value format. It can be used to include any relevant contextual data.
     """
 
+    tools: list[ToolDefinition] | None = None
```
jgreer013, line 419: @aniruddh-alt FYI, we'll want to ensure data synthesis ultimately outputs data with these definitions.

jgreer013 (approved):

src/oumi/core/types/conversation.py
shanghongsim: Done

src/oumi/core/types/tool_call.py
shanghongsim: Done

src/oumi/core/datasets/base_sft_dataset.py
shanghongsim: Done

wizeng23 (approved):

PR #2393, opened 2026-04-24
feat(deploy)!: typed exception hierarchy for deployment provider errors

brian-nguyen:
# Description

`raise_api_error` now raises typed `DeployApiError` subclasses selected by HTTP status (`DeployNotFoundError` / `DeployRateLimitError` / `DeployTransientError` / `DeployInvalidRequestError`), instead of a bare `ValueError` that leaked method, URL, and account ID into the message.

`str(exc)` is `"Failed to {context}: {detail}"`. Method, URL, body stay on structured attrs + DEBUG logs.

Provider-specific 4xx sub-types (e.g. `FireworksUnsupportedHardwareError`) live next to their client in `oumi.deploy.fireworks_errors` and are wired in via the `classify_4xx` hook on `raise_api_error`. Adding a new Fireworks-specific type is a local change to `fireworks_errors.py` — no edits to shared code required.

**BREAKING:** catch `oumi.deploy.errors.DeployApiError` (or a subclass) instead of `ValueError`.

## Related issues

Fixes LOU-1616

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/deploy/errors.py
+++ b/src/oumi/deploy/errors.py
@@ -1,0 +1,95 @@
+# Copyright 2026 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Provider-neutral typed exception hierarchy for deployment API errors.
+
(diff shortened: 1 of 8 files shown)


oelachqar (approved):

src/oumi/deploy/errors.py
```diff
@@ -0,0 +1,95 @@
...
+client implementation — see :mod:`oumi.deploy.fireworks_errors` — and
+are plumbed in via the ``classify_4xx`` hook on
+:func:`~oumi.deploy.utils.raise_api_error`. Do NOT re-export
+provider-specific classes from this module; consumers wanting a
+provider-specific catch should import from the provider's module
+explicitly.
+"""
+
+
+class DeployApiError(Exception):
```
idoudali, line 45: @oelachqar we do want to be generating an error message at the top-level of the CLI in case of this error or should this exception print the full stack ? 

if we want to try to generate a error probably we need to create a top-level exception under https://github.com/oumi-ai/oumi/blob/main/src/oumi/exceptions.py 

That all our other exceptions inherit from

PR #2400, opened 2026-04-29
Wire tool calling through RemoteInferenceEngine

shanghongsim:
# Description

PR #2397 added tool calling to `VLLMInferenceEngine`. The remote OpenAI-compatible engines do not yet send `Conversation.tools` on the request or parse `tool_calls` from the response, so tool fields are silently dropped on both sides. This PR adds tool-calling support to `RemoteInferenceEngine` (the OpenAI-compatible base class), so the same `Conversation` works on `engine: VLLM` and `engine: OPENAI`.

Providers that do not override `_convert_conversation_to_api_input` — Together, Fireworks, DeepSeek, Cerebras, OpenRouter, Parasail — pick up the new behavior automatically through the base class. `OpenAIInferenceEngine` overrides the method but calls `super()` after tweaking reasoning-model fields, so it inherits too.

Three OpenAI-compatible engines are *not* wired up by this PR and will silently drop tool fields until a follow-up: `SambanovaInferenceEngine` and `RemoteVLLMInferenceEngine` both override `_convert_conversation_to_api_input` without calling `super()` (their request bodies are minor variations of OpenAI chat completions, so a `super()`-based refactor is the natural fix). `SGLangInferenceEngine` uses SGLang's Native API (`text` + `sampling_params`) rather than chat completions, so it needs its own translation layer.

Anthropic, Gemini, Bedrock, and GCP Vertex use different wire formats and are out of scope here. Anthropic tool calling and a related Anthropic multimodal image-shape bug fix ship as stacked follow-ups (see Related issues below).

## Motivation

A previous attempt at remote tool calling landed in commit [`e4f7054`](https://github.com/oumi-ai/oumi/commit/e4f7054c7c2a359e48095e7b6a780689865774e1) on `oelachqar/infer-tool-calling`. Two of its three design choices conflict with where the codebase landed after PR #2397, so we keep one and replace the other two:

- `e4f7054` put tools on `GenerationParams.tools`. PR #2397 puts them on `Conversation.tools`, matching `tokenizer.apply_chat_template(messages, tools=...)` and letting the vLLM engine group conversations by identical tool definitions. We keep `Conversation.tools`.
- `e4f7054` hand-rolled OpenAI message-list construction inside `OpenAIInferenceEngine`. We extend `RemoteInferenceEngine._convert_conversation_to_api_input` once and reuse the existing `create_list_of_message_json_dicts` helper, which already forwards `tool_calls` and `tool_call_id`.
- `e4f7054` introduced typed Pydantic classes in `src/oumi/core/types/tool_call.py`. The current branch already adopts that pattern (commit `de351040`), so we use it. We dump typed objects to JSON via `model_dump(mode="json", exclude_none=True)` at API boundaries and let Pydantic validate response-side tool calls into typed `ToolCall` instances.

### `GenerationParams`

- Added `tool_choice: Any | None = None`. Accepts the OpenAI value space (`"none"`, `"auto"`, `"required"`, or a `{"type": "function", "function": {"name": "..."}}` dict). Engines targeting non-OpenAI providers translate.
- Added `parallel_tool_calls: bool = True`. Default matches the OpenAI default; only sent on the request when explicitly set to `False`.
- `tool_choice` is annotated `Any` because omegaconf does not support unions of containers. Same workaround as `GuidedDecodingParams.json`.

### `RemoteInferenceEngine` (base class)

- Request side: replaces the inline message-list comprehension with `create_list_of_message_json_dicts(group_adjacent_same_role_turns=False)`. That helper already forwards `tool_calls` and `tool_call_id` and emits `content: null` for tool-only assistant turns.
- When `Conversation.tools` is non-empty, dumps each `ToolDefinition` to its OpenAI-format dict and adds it to the request.
- When `tool_choice` is set, forwards it verbatim. When `parallel_tool_calls` is `False`, forwards it (otherwise omits to keep the request minimal).
- Response side: reads `message.tool_calls` and lets Pydantic coerce into `list[ToolCall]`. Propagates `Conversation.tools` onto the result.
- `get_supported_params` extended with `tool_choice` and `parallel_tool_calls`.

`OpenAIInferenceEngine` requires no code change. Its reasoning-model override mutates `temperature` and `logit_bias` and then calls `super()`, so the base-class change flows through.

### Examples

- `configs/apis/openai/infer_gpt_4o_tool_calling.yaml`

### Tests

- 5 new tests on `RemoteInferenceEngine`: request body shape (tools, tool_choice, parallel_tool_calls), assistant tool-call round-trip, response parsing into typed `Message.tool_calls`.
- 1 new test on `OpenAIInferenceEngine` confirming the reasoning-model temperature override does not drop tool fields.
- Existing tests inspecting the message wire shape are updated for the new primitive-string content emitted by the helper.

## Out of scope

Anthropic, Gemini, Bedrock, GCP Vertex, SGLang custom protocols, tool streaming, batch-API tool-calling tests (the OpenAI batch path inherits the new request shape).

## Related issues

Stacks on top of #2397. The Anthropic side ships as two stacked follow-ups:

- Anthropic multimodal image block-shape bug fix (will be opened next, stacked on this PR).
- Anthropic tool calling (will be opened after, stacked on the image-fix PR).

Not closing a specific tracked issue.

## Before submitting

## Reviewers

### DIFF
--- a/configs/apis/openai/infer_gpt_4o_tool_calling.yaml
+++ b/configs/apis/openai/infer_gpt_4o_tool_calling.yaml
@@ -1,0 +1,35 @@
+# OpenAI Chat Completions inference config with tool calling.
+#
+# Sends ``Conversation.tools`` (OpenAI function-calling schema) on every
+# request and parses ``message.tool_calls`` out of the response into
+# structured ``Message.tool_calls``.
+#
+# Requirements:
+#   - Set the ``OPENAI_API_KEY`` environment variable.
+#
+# Usage:
+#   oumi infer -i -c configs/apis/openai/infer_gpt_4o_tool_calling.yaml
+#
+# The input JSONL should carry a top-level ``tools`` field on each
+# conversation (OpenAI function-calling schema). Output messages will
+# populate ``tool_calls`` when the model emits a structured call, and
+# ``metadata.finish_reason`` will be set to ``tool_calls``.
+#
+# See Also:
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/infer/infer.html
+#   - Config class: oumi.core.configs.InferenceConfig
+#   - Other inference configs: configs/**/inference/
+
+model:
+  model_name: "gpt-4o"
+
+engine: OPENAI
+
+generation:
+  max_new_tokens: 4096
+  temperature: 0.3
+  # Let the model decide whether to call a tool. Other valid values:
+  # "none" (force a text reply), "required" (force a tool call), or
+  # {"type": "function", "function": {"name": "..."}} to force a
+  # specific tool.
+  tool_choice: "auto"
--- a/src/oumi/core/configs/params/generation_params.py
+++ b/src/oumi/core/configs/params/generation_params.py
@@ -130,6 +130,25 @@ class GenerationParams(BaseParams):
     format (e.g., reasoning tokens, tool call markers).
     """
 
+    # Should be Union[str, dict[str, Any], None], but omegaconf does not
+    # support unions of containers, so we widen to Any.
+    tool_choice: Any | None = None
+    """Controls how the model selects tools when ``Conversation.tools`` is set.
+
+    Uses the OpenAI value space: ``"none"``, ``"auto"``, ``"required"``, or
+    ``{"type": "function", "function": {"name": "..."}}``. Default ``None``
+    omits the field from the request, letting the API apply its own default
+    (``"auto"`` when tools are present). Engines targeting non-OpenAI providers
+    (e.g. Anthropic) translate this value to the provider's wire shape.
+    """
+
+    parallel_tool_calls: bool = True
+    """Whether the model may emit multiple tool calls in one response.
+
+    OpenAI-specific. Default ``True`` matches the OpenAI API default; the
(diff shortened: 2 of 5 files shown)


src/oumi/inference/anthropic_inference_engine.py
```diff
@@ -24,18 +24,65 @@
...
 from oumi.utils.logging import logger
 
 _CONTENT_KEY: str = "content"
 
+_OPENAI_TO_ANTHROPIC_TOOL_CHOICE: dict[str, dict[str, str]] = {
+    "auto": {"type": "auto"},
+    "required": {"type": "any"},
+}
+
+# Anthropic accepts image/jpeg, image/png, image/gif, image/webp for `image`
```
oelachqar: @shanghongsim is the image stuff related to this PR?

oelachqar (approved):

src/oumi/inference/anthropic_inference_engine.py
shanghongsim: It fixes an existing bug for multimodal Anthropic requests. Previously, the engine sent OpenAI-shape image blocks which Anthropic does not accept. But I agree this probably bloats the PR. Will make it into a separate PR.

PR #2410, opened 2026-05-01
[Synth Env] Phase 1: Stateless tool simulation via batched step()

aniruddh-alt:
## Summary

**1. Batched `step()` API** — `BaseEnvironment.step` now takes `list[(tool_id, args)] -> list[ToolResult]` instead of single-call. `DeterministicEnvironment` implements its own loop internally; `SyntheticEnvironment` overrides `step` to fold batches into one inference (impl lands in commit 2). The synthesizer's per-call `_run_tool_call` is replaced by `_dispatch_tool_calls`, which validates each call (args parse, type, known tool), groups surviving calls by environment, and batch-dispatches each group via `env.step(...)`. If a batched `step()` raises, the dispatcher falls back to per-call dispatch within that group so individual errors stay attributed.

**2. Stateless synthetic-env step implementation** — `SyntheticEnvironment.step` now executes simulator inference instead of `NotImplementedError`. `attach_inference(engine, base_config)` wires the synthesizer's already-built engine + base config into the env (no separate `InferenceConfig` on `SyntheticEnvironmentKwargs`). Per-tool simulator configs derive from `base_config` by overlaying `GuidedDecodingParams(json=tool.output_schema)`, mirroring `_planner_inference_config`. The synthesizer's tool-dispatch construction grew one `isinstance` check that calls `attach_inference` after `build_environment`. Sanitization is permissive but bounded: `extract_json` salvages a JSON object from prose-wrapped simulator output; empty / unparseable / schema-mismatched outputs raise `ToolError`, which the dispatcher converts into a structured TOOL error message so the agent can self-correct.

**3. ToolRouter** – Unifies tool routing and dispatching of tool calls. Moves logic from conversation_synthesizer.py to tool_router.py.

Adds `configs/examples/synthesis/mcp_docs_lookup_synth.yaml` — an MCP-style stateless docs-lookup config with three typed tools (`resolve_library_id`, `query_docs`, `search_examples`), exercising the new synthetic env end-to-end.

New unit tests for `SyntheticEnvironment.step` (cache hit/miss/mixed, batched dispatch, JSON-from-prose extraction, schema-validation pass/fail, no-output-schema fallback, attach-inference-not-called error) + a wiring test in the synthesizer test file confirming `attach_inference` is invoked for synthetic envs.

### DIFF
--- a/configs/examples/synthesis/mcp_docs_lookup_synth.yaml
+++ b/configs/examples/synthesis/mcp_docs_lookup_synth.yaml
@@ -21,11 +21,15 @@
 #   - Set OPENAI_API_KEY in the environment.
 #
 # Usage:
-#   oumi synth -c configs/examples/synthesis/mcp_docs_lookup_synth.yaml
+#   oumi synth -c oumi://configs/examples/synthesis/mcp_docs_lookup_synth.yaml
 #
 # See Also:
-#   - oumi.environments.synthetic_environment.SyntheticEnvironment
-#   - oumi.core.configs.EnvironmentConfig
+#   - Documentation: https://oumi.ai/docs/en/latest/user_guides/synth.html
+#   - Config class: oumi.core.configs.SynthesisConfig
+#   - Params class: oumi.core.configs.params.synthesis_params.GeneralSynthesisParams
+#   - Env config class: oumi.core.configs.environment_config.EnvironmentConfig
+#   - SyntheticEnvironment: oumi.environments.synthetic_environment.SyntheticEnvironment
+#   - Other synthesis configs: configs/**/*synth.yaml
 
 strategy: GENERAL
 num_samples: 150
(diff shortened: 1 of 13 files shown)


configs/examples/synthesis/mcp_docs_lookup_synth.yaml
```diff
@@ -0,0 +1,362 @@
+# MCP-style Documentation Lookup — Stateless Synthetic Tool-Use Synthesis
+#
+# Generates multi-turn conversations between a developer asking how to use a
+# library and an assistant that answers via three MCP-style stateless tools:
+#
+#   1. `resolve_library_id` — turns a library NAME (e.g. "react") into a
```
jgreer013, line 12: Do we need this tool? This feels like a tool for a multi-library MCP instead of a single library

jgreer013 (approved):

configs/examples/synthesis/mcp_docs_lookup_synth.yaml
aniruddh-alt: This was supposed to be an MCP inspired by Context-7 which provides access to multiple libraries, and is resolved using a library id. I can add more information to clarify the intention of this dataset and the MCP it tries to simulate.
https://context7.com/

PR #2421, opened 2026-05-05
[Det Env v2] Phase 1: Tool schema unification & validation

aniruddh-alt:
# Description

Part of a new **Deterministic Environment v2** PR series — replacing the original 8-PR chain (#2386–#2411) with a more logical decomposition. **Phase 1 of 7.**

**What changed**: Replace the chain's `ToolSchema` dataclass with main's Pydantic `JSONSchema` (#2392), add a tool-error exception hierarchy, JSON-schema argument validation helpers, an OpenAI-wire-format projection, and relocate `ToolResult` to `core.types.tool_call` to co-locate with the rest of the tool-related types.

**Why**: The skeleton (#2358) shipped `ToolSchema` before #2392 unified tool types on main. This PR eliminates the duplicate and gives the chain a single, shared validation surface — every subsequent phase consumes these types, so doing this first keeps later PRs focused.

## What's in scope

- **Type unification**: `ToolSchema` deleted; `ToolParams.parameters` / `output_schema` are typed as `Any` so OmegaConf carries them as plain dicts through YAML, then `__post_init__` coerces to `JSONSchema`.
- **`_coerce_json_schema(raw)`** helper — handles `None | JSONSchema | Mapping`; replaces `ToolSchema.create()`.
- **`ToolError` / `ToolArgumentError` / `ToolLookupError`** exception hierarchy.
- **`_validate_json_schema_value`** moved from `synthetic_environment.py` to `tool_params.py` so other modules can share it without crossing the environments-package boundary.
- **`validate_arguments_against_schema(arguments, schema)`** helper, with `ToolParams.validate_arguments()` delegating to it.
- **`ToolParams.to_tool_definition()`** — projects to OpenAI-wire-format `ToolDefinition` (drops chain-internal fields like `output_schema`, `read_only`, display name).
- **`ToolResult` relocated** from `tool_params.py` to `core/types/tool_call.py` (co-located with `ToolDefinition` / `ToolCall` / `FunctionCall`); drops unused `BaseParams` inheritance.

## Notes

- The skeleton's `required ⊆ properties` check on `ToolSchema` is dropped — JSON Schema spec doesn't require this and standard validators don't enforce it.
- 398 tests pass across `tests/unit/core/configs/params/`, `tests/unit/core/synthesis/`, `tests/unit/environments/`, `tests/unit/builders/test_environments.py`.

## Series context

This is PR 1 of a 7-PR series replacing the original chain:

1. **Tool schema unification & validation infrastructure** ← this PR
[...]

PR #2411, opened 2026-05-01
[Det Env] Phase 8: Unify tool schema types with main and project Conversation.tools on output

aniruddh-alt:
# Description

Part of the **Deterministic Environment** PR chain (Phase 8) — Stage: refactor / wire-format integration with main's recently-landed structured tool calling.

**What changed**: Eliminates the chain's dataclass `ToolSchema` in favor of the Pydantic `JSONSchema` from `oumi.core.types.tool_call` (Shang Hong's #2392 wire format), adds `ToolParams.to_tool_definition()` projection, populates `Conversation.tools` on synthesizer output, projects in-conversation tool messages to the structured `Message.tool_calls` / `Role.TOOL` form, relocates `ToolResult` to `oumi.core.types.tool_call`, and renames the in-text marker from `<tool_result>` to `<tool_response>` to match the Hermes / Qwen chat-template convention.

**Why**: Datasets generated by the chain need to flow into the structured SFT pipeline (`apply_chat_template(tools=...)` and the `all_assistant_turns` collator) without a downstream re-projection step. The chain previously had a parallel `ToolSchema` dataclass duplicating the type Shang Hong already shipped — consolidating removes the duplicate and lets the chain speak the same wire format as the rest of the codebase, end-to-end (declarations + calls + results).

## What's in scope

- **Type unification**: chain's `ToolSchema` deleted; `ToolParams.parameters` / `output_schema` now coerce to `JSONSchema`. Validation logic extracted as a standalone `validate_arguments_against_schema(args, schema)` helper.
- **Tool-definition projection**: new `ToolParams.to_tool_definition() -> ToolDefinition` (drops chain-internal `output_schema` / `read_only` / `display_name` — they have no slot in the OpenAI wire format).
- **`Conversation.tools` on output**: `_synthesize_all_samples` populates `Conversation.tools` via the projection.
- **Structured tool-message projection**: a final pass in `_synthesize_all_samples` rewrites the chain's text-format tool exchange to OpenAI wire format before output:
  - `Role.ASSISTANT + '<tool_call>{...}</tool_call>'` text → `Message(role=ASSISTANT, content=prose|None, tool_calls=[ToolCall(id, function=FunctionCall(...))])`.
  - `Role.USER + '<tool_response>...</tool_response>'` text → `Message(role=TOOL, content=..., tool_call_id=...)`.
  - Synthetic tool-call IDs are opaque (`call_<8-hex>`), matched FIFO between each assistant turn's calls and the tool-result messages that follow.
[...]

PR #2386, opened 2026-04-22
[Det Env] Phase 1: Planner guided decoding & tool schema

aniruddh-alt:
# Description

Part of the **Deterministic Environment** PR chain (Phase 1 of 6) - Stage: planner improvements

**What changed**: Add JSON schema guided decoding for planner calls, inject tool context into assistant personas, and refactor `to_llm_schema()` to use tool IDs instead of display names.

**Why**: The planner needs structured JSON output (not markdown-fenced) and assistants need tool awareness in their persona prompts for coherent tool-use conversations.

## PR Chain

1. **Planner guided decoding & tool schema** (this PR)
2. Grounding types & config
3. Grounding in environments
4. Grounding synthesis pipeline
5. Tool use robustness & polish
6. JSON repair & planner refinement

## Related issues

Part of deterministic environment grounding work (replaces #2384)

## Before submitting
- [x] Did you link the issue(s) related to this PR in the section above?
- [x] Did you add / update tests where needed?

PR #2422, opened 2026-05-05
[Det Env v2] Phase 2: Planner guided JSON decoding

aniruddh-alt:
# Description

Part of the **Det Env v2** PR series — **Phase 2 of 7**.

**What changed**: Switch the planner to JSON guided decoding against a Pydantic-derived `PLANNER_JSON_SCHEMA`. Planner output now arrives as a structured `{"turns": [...]}` object instead of markdown-fenced JSON.

**Why**: Markdown-fenced output is brittle — truncated fences and trailing prose force a re-parse loop and add latency. Guided decoding lets the inference engine constrain the output directly so the planner returns valid JSON on the first attempt.

## What's in scope

- **`PlannerOutput` / `PlannedTurn`** Pydantic models in `core.types.conversation`, with a derived `PLANNER_JSON_SCHEMA: dict` constant. The schema enforces a `{"turns": [{"turn": int, "instruction": str}, ...]}` shape.
- **`ConversationSynthesizer._planner_inference_config()`** — returns a copy of the user's inference config with `GuidedDecodingParams(json=PLANNER_JSON_SCHEMA)` set on the planner call only. The original config is never mutated, so per-turn (non-planner) inference stays unconstrained.
- **`_create_planner_prompt`** updated to emit the wrapped object form with no markdown fences, matching the schema.
- **`_parse_plan`** reads the wrapped object form first and falls back to a bare list for backward compatibility.
- **`_generate_plan`** calls `_planner_inference_config()` instead of passing the raw config.
- **Test fixture** `mock_inference_config` switched from `Mock(spec=InferenceConfig)` to a real `InferenceConfig` so `dataclasses.replace` works in the new helper.
- New tests: `test_parse_plan_unwraps_object_form`, `test_generate_plan_uses_planner_only_guided_decoding` (verifies planner gets guided decoding, turn calls don't, and the original config is left untouched).

## Notes

- 447 tests pass across `tests/unit/core/synthesis/`, `tests/unit/core/configs/params/`, `tests/unit/utils/test_str_utils.py`.
- `_planner_inference_config` is a 4-line `dataclasses.replace` chain — both `InferenceConfig.generation` and `GenerationParams` are guaranteed to be present (non-optional fields with defaults), so no `getattr` / isinstance defensive code.

## Series context

This is PR 2 of a 7-PR series (rebooted from the original 8-PR chain #2386–#2411). Each PR targets `main` directly, opened one at a time as predecessors land.

1. ~~Tool schema unification & validation infrastructure~~ → **#2421 (merged)**
2. **Planner: guided JSON decoding** ← this PR
3. Grounding system: types + environment integration
4. Grounding in the synthesis pipeline
7. Library tool-use synthesis example + end-to-end test

## Related issues

None.

## Before submitting
- [x] Tests added (`test_parse_plan_unwraps_object_form`, `test_generate_plan_uses_planner_only_guided_decoding`).

### DIFF
--- a/src/oumi/core/synthesis/conversation_synthesizer.py
+++ b/src/oumi/core/synthesis/conversation_synthesizer.py
@@ -12,19 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
 import random
 
 from oumi.builders.inference_engines import build_inference_engine
 from oumi.core.configs.environment_config import EnvironmentConfig
 from oumi.core.configs.inference_config import InferenceConfig
 from oumi.core.configs.inference_engine_type import InferenceEngineType
+from oumi.core.configs.params.guided_decoding_params import GuidedDecodingParams
 from oumi.core.configs.params.synthesis_params import (
     GeneralSynthesisParams,
     MultiTurnAttribute,
 )
 from oumi.core.configs.params.tool_params import ToolParams
 from oumi.core.synthesis.attribute_formatter import AttributeFormatter
-from oumi.core.types.conversation import Conversation, Message, Role
+from oumi.core.types.conversation import (
+    PLANNER_JSON_SCHEMA,
+    Conversation,
+    Message,
+    Role,
+)
 from oumi.utils.logging import logger
 from oumi.utils.str_utils import extract_json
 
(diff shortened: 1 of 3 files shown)


jgreer013 (approved):

shanghongsim:
May I don't have enough context on the synthesis portions but what is the planner for?

shanghongsim (commented):
so is markdown wrapped json still allowed as a parsable synthesis output?

aniruddh-alt:
> May I don't have enough context on the synthesis portions but what is the planner for?

A conversation planner is an LLM call that plans out how a conversation should go. It guides the user and assistant turns by providing turn-wise instructions on what either of them should do in their respective turns. This is essentially a meta-planning approach that ensures consistent quality of data samples.

aniruddh-alt:
> so is markdown wrapped json still allowed as a parsable synthesis output?

No, markdown wrapped json should not be an allowed output from this PR

PR #2424, opened 2026-05-05
[Det Env v2] Phase 3: Grounding system — types + environment integration

aniruddh-alt:
# Description

Phase 3 of 7 in the Det Env v2 series.

This PR sets up the env-side primitives for grounding: types, environment hooks, and validation. The synth-side consumer lands in PR4 (#2428).

## The headline change

**The data model for deterministic envs flipped.** Previously, `DeterministicTool` was a `ToolParams` subclass that carried its own lookup data (`deterministic_outputs`) and grounding config. This PR removes the subclass entirely — tools become plain `ToolParams` (just contracts), and the env owns the data:

```yaml
env:
  env_type: deterministic
  tools:
    - id: lookup_book_status
      parameters: {...}
      output_schema: {...}
  env_kwargs:
    lookup_table:
      lookup_book_status:
        - input: { book_id: B001 }
          output: { title: "...", status: "..." }
  grounding:
    sample_size: 3
    seed: 42
    tools:
      lookup_book_status:
        fields: [book_id, title, status]
```

This brings deterministic envs in line with synthetic envs (and future executable envs): tools own contracts, env owns data.

## What's new

**Types** (`oumi/core/configs/params/grounding_params.py`):
- `GroundingConfig` (per-env): `sample_size`, `seed`, `tools: dict[str, ToolGroundingConfig]`.
- `ToolGroundingConfig` (per-tool whitelist): just `fields: list[str]`. No `key` field — earlier drafts had one, it was unused.
- `GroundingFact`: env-agnostic flat `data: dict[str, Any]`.

**Env hooks** (`base_environment.py`, `utils.py`):
- `BaseEnvironment.sample_grounding(n, *, rng, tool_ids)` — default returns `[]`.
- `BaseEnvironment.describe_grounding(facts)` — default delegates to `describe_grounding_default`, which renders a bulleted markdown block.

**Deterministic env runtime** (`deterministic_environment.py`):
- `ToolLookupEntry` — what `DeterministicToolOutput` used to be. Has `input`, `output`, `matches(args)`, and `input_key()` (canonical JSON for matching/dedup).
- `DeterministicEnvironmentKwargs.lookup_table: dict[str, list[ToolLookupEntry]]`.
- `step()` resolves via `self._kwargs.lookup_table[tool_id]`. Raises `ToolLookupError` on no-match (with the configured inputs surfaced for self-correction); raises `ValueError` for unknown tool ids.
- `sample_grounding` walks `params.grounding.tools`, projects each entry's `{**input, **output}` through the configured `fields` whitelist, samples without replacement.

**Validators**:
- `EnvironmentParams._validate_grounding_has_tools` — env declares grounding but `grounding.tools` is empty → hard error.
- `EnvironmentParams._warn_on_stale_grounding_tool_ids` — `grounding.tools` names a tool not in `env.tools` → warning (dormant config, not invalid).
- `DeterministicEnvironment._validate_lookup_table`:
  - Tool declared with no entries in `lookup_table` → hard error.
  - `lookup_table` key for an unknown tool → warning (dormant data, not invalid).
  - Duplicate inputs within a tool's entries → hard error.
- `DeterministicEnvironment.from_params` rejects unknown `env_kwargs` keys with a message naming the typo and listing the known keys.

## Migration for existing YAML configs

Anything using the old `DeterministicTool` / `deterministic_outputs` shape needs to move data from per-tool to env-level:

```yaml
# Before
tools:
  - id: lookup_book_status
    deterministic_outputs:
      - input: { book_id: B001 }
        output: { ... }
    grounding:
      key: book_id
      fields: [book_id, title, status]

# After
tools:
  - id: lookup_book_status
    parameters: {...}        # contract only
    output_schema: {...}
env_kwargs:
  lookup_table:
    lookup_book_status:
      - input: { book_id: B001 }
        output: { ... }
grounding:
  sample_size: 3
  tools:
    lookup_book_status:
      fields: [book_id, title, status]   # no `key` field anymore
```

The library example YAML migrates in PR7.

## Public API changes

`oumi.environments` exports:
- **Removed**: `DeterministicTool`, `DeterministicToolOutput`.
- **Added**: `ToolLookupEntry`, `GroundingConfig`, `GroundingFact`, `ToolGroundingConfig`.

## Notes

- 229 tests pass across `tests/unit/core/configs/params/`, `tests/unit/environments/`, `tests/unit/builders/`.

## Series context

3. **Grounding system: types + env-owned lookup tables** ← this PR
4. Grounding in the synthesis pipeline — #2428 (stacked on this)
7. Library tool-use synthesis example + e2e

## Before submitting

- [x] Tests for `GroundingConfig` / `ToolGroundingConfig` validation, env-grounding-without-tools detection, `lookup_table` validation (tool-without-entries error, stale-key warning, unknown-env-kwargs error, duplicate-input error), `step()` lookup-error path, `sample_grounding` projection and scoping, default base hooks.


src/oumi/core/configs/params/tool_params.py
```diff
@@ -88,6 +91,12 @@ def create(cls, raw: Any) -> ToolParams:
                 else None
             ),
             read_only=raw.get("read_only", True),
+            grounding=(
+                grounding_raw
+                if grounding_raw is None
+                or isinstance(grounding_raw, ToolGroundingConfig)
+                else ToolGroundingConfig(**grounding_raw)
+            ),
```
aniruddh-alt: **Slop / double-defensive parsing:** `create()` does the dict→`ToolGroundingConfig` coercion here, and `__post_init__` does it again at lines 121-124. Since `__post_init__` runs on every `__init__`, this branch in `create()` is dead — the dataclass will coerce the raw dict regardless.

**Suggestion:** drop the conditional in `create()` (and the `grounding_raw` local at line 81) and pass `grounding=raw.get("grounding")` directly; let `__post_init__` own the coercion. That matches the pattern already used for `parameters` / `output_schema` (which delegate to `_coerce_json_schema` only at one site each).

```python
return cls(
    ...
    read_only=raw.get("read_only", True),
    grounding=raw.get("grounding"),
)
```

Same comment applies to `DeterministicTool.create` (lines 88-93 of `deterministic_tool.py`).

(agent)

src/oumi/environments/deterministic_tool.py
```diff
@@ -83,5 +85,11 @@ def create(cls, raw: Any) -> DeterministicTool:
                 else None
             ),
             read_only=raw.get("read_only", True),
+            grounding=(
+                grounding_raw
+                if grounding_raw is None
+                or isinstance(grounding_raw, ToolGroundingConfig)
+                else ToolGroundingConfig(**grounding_raw)
+            ),
```
aniruddh-alt: **Same double-coercion as in `tool_params.py`:** `__post_init__` (inherited from `ToolParams`) coerces a raw `dict` for `grounding` already, so this branch is dead. Drop the `grounding_raw` local at line 76 and pass `grounding=raw.get("grounding")` directly.

(agent)

src/oumi/environments/deterministic_environment.py
```diff
@@ -62,12 +65,23 @@ def __init__(
...
         tool = self._lookup_tool(tool_id)
         for entry in tool.deterministic_outputs:
             if entry.matches(arguments):
                 return ToolResult(output=entry.output)
-        return ToolResult(output={})
+        available = [entry.input for entry in tool.deterministic_outputs]
+        raise ToolLookupError(
+            f"No deterministic output matches arguments "
+            f"{json.dumps(arguments, sort_keys=True)} for tool '{tool_id}'. "
+            f"Configured inputs: {json.dumps(available, sort_keys=True)}"
```
aniruddh-alt, line 114: **Behavior change worth a mention in the chain context:** previously `step()` returned `ToolResult(output={})` on a no-match, now it raises `ToolLookupError`. I checked `origin/main` and there are zero callers of `DeterministicEnvironment.step()` in the synthesizer (the tool-call loop lands in PR5), so this is safe to land here without coordinated changes elsewhere.

Question: should the error string `json.dumps(available, sort_keys=True)` be capped? For a tool with hundreds of `deterministic_outputs` entries, this error message will balloon and end up in the planner prompt verbatim via the tool-error projection. The library example tops out around \~25 entries today so it's fine in practice, but if any future env wires up a larger lookup table the prompt cost gets surprising. Maybe truncate at e.g. 20 entries with "... and N more"?

(agent)

src/oumi/environments/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    if not facts:
+        return ""
+    lines: list[str] = []
+    for fact in facts:
+        parts = [
+            f"{key}={_format_grounding_value(value)}"
+            for key, value in fact.data.items()
+        ]
+        lines.append(f"- {', '.join(parts)}")
+    return "\n".join(lines)
```
aniruddh-alt, line 47: **Edge case:** a fact with an empty `data` dict renders as `"- "` (bullet with trailing space and no content). The empty-fact-list case is handled at line 39, but an empty data dict per fact is not. Probably not reachable from `DeterministicEnvironment.sample_grounding` (every row produces at least the configured `fields` ∩ `{**input, **output}`), but if a custom subclass emits a `GroundingFact(data={})` it'd produce a malformed bullet.

Minor — flagging in case it's worth a guard or a test.

Separate nit: `_format_grounding_value` doesn't escape embedded `"` in string values (e.g. `title='Foo "the" Bar'` → `title="Foo "the" Bar"`). Not parsed downstream so it's cosmetic, but a passing user-supplied string with quotes could read oddly in the planner prompt.

(agent)

src/oumi/environments/deterministic_environment.py
```diff
@@ -17,100 +17,169 @@
...
+        Raises:
+            ValueError: If ``tool_id`` is not declared in this env's tools list.
+            ToolLookupError: If no entry in the env's lookup table matches
+                the provided arguments.
+        """
+        known_tool_ids = {tool.id for tool in self._params.tools}
+        if tool_id not in known_tool_ids:
+            raise ValueError(
+                f"Tool '{tool_id}' not found in environment '{self._params.id}'. "
+                f"Available tools: {sorted(known_tool_ids)}"
```
aniruddh-alt: **Performance:** `known_tool_ids = {tool.id for tool in self._params.tools}` recomputes the tool-id set on every `step()` call. For a planner that calls dozens of tools per conversation, this is O(N_tools) per call when it could be O(1) lookup against a cached set.

**Suggestion:** cache on `__init__`:

```python
def __init__(self, params, kwargs):
    self._params = params
    self._kwargs = kwargs
    self._tool_ids = {tool.id for tool in params.tools}
    self._validate_lookup_table()

def step(self, tool_id, arguments):
    if tool_id not in self._tool_ids:
        raise ValueError(...)
    ...
```

`_validate_lookup_table` can also use `self._tool_ids` instead of recomputing it.

(agent)

src/oumi/core/configs/params/grounding_params.py
```diff
@@ -0,0 +1,84 @@
...
+@dataclass
+class ToolGroundingConfig(BaseParams):
+    """Per-tool field whitelist for grounding projection.
+
+    Tools listed in ``GroundingConfig.tools`` are projected through
+    ``fields`` when the env samples grounding facts. Tools not listed
+    contribute nothing.
+    """
+
+    fields: list[str] = field(default_factory=list)
```
aniruddh-alt: **Defaults / validation:** `fields: list[str] = field(default_factory=list)` declares an empty-list default, but `__post_init__` raises `ValueError("fields must be non-empty")` immediately. So the default is *invalid* — instantiating `ToolGroundingConfig()` always fails.

**Suggestion:** make `fields` a required positional with no default — `fields: list[str]` — so attempting to construct without it fails at the type-system level, not via post-init. Matches the convention in `ToolParams` for required fields like `id`, `name`, `description`.

(agent)

src/oumi/environments/deterministic_environment.py
```diff
@@ -17,100 +17,169 @@
...
 
-    @staticmethod
-    def _validate_tools(tools: list[DeterministicTool]) -> None:
-        for tool in tools:
-            if not tool.deterministic_outputs:
+    def _validate_lookup_table(self) -> None:
+        """Validate the env's lookup_table against its tool list.
+
+        - Stale ``lookup_table`` keys (no matching tool): log a warning;
+          entries are dormant.
```
aniruddh-alt, line 170: **Backwards compat / error message regression:** the prior `from_params` raised a clear `"DeterministicEnvironment does not accept env_kwargs, got: ['x']"` when callers passed unknown kwargs. The new code just splats into the dataclass, so a typo (e.g. `lookup_tabel`) raises a less obvious `TypeError: __init__() got an unexpected keyword argument 'lookup_tabel'` instead.

**Suggestion:** check unknown keys explicitly before construction — same shape as the old guard:

```python
known = {f.name for f in dataclasses.fields(DeterministicEnvironmentKwargs)}
unknown = set(raw_kwargs) - known
if unknown:
    raise ValueError(
        f"DeterministicEnvironment got unknown env_kwargs: {sorted(unknown)}. "
        f"Known: {sorted(known)}"
    )
kwargs = DeterministicEnvironmentKwargs(**raw_kwargs)
```

**Why:** users hand-edit YAML; one validation error pointing at the typo is much friendlier than a TypeError that requires inspecting the dataclass.

(agent)

src/oumi/environments/deterministic_environment.py
```diff
@@ -17,100 +17,169 @@
...
                 raise ValueError(
-                    f"Deterministic tool '{tool.id}' must have at least one "
-                    "deterministic_output entry."
+                    f"Tool '{tool.id}' has no entries in lookup_table for "
+                    f"environment '{self._params.id}'."
                 )
             seen: set[str] = set()
-            for entry in tool.deterministic_outputs:
+            for entry in entries:
                 key = json.dumps(entry.input, sort_keys=True)
```
aniruddh-alt: **Code dup:** `json.dumps(entry.input, sort_keys=True)` is computed here AND in `LookupEntry.matches()` (line 38). Same canonical key for the same purpose.

**Suggestion:** factor a `LookupEntry.input_key()` returning the JSON form, and call it from both sites:

```python
def matches(self, arguments):
    return self.input_key() == json.dumps(arguments, sort_keys=True)

def input_key(self):
    return json.dumps(self.input, sort_keys=True)
```

Minor / flagging for symmetry, not blocking.

(agent)

jgreer013 (approved):

PR #2428, opened 2026-05-06
[Det Env v2] Phase 4: Grounding in the synthesis pipeline

aniruddh-alt:
# Description

Phase 4 of 7 in the Det Env v2 series. **Stacked on #2424** — the base of this branch is `aniruddh-alt/det-env-v2-03-grounding`. Once #2424 merges, this PR's base auto-rebases to `main`.

This PR is the synthesizer-side consumer of the grounding hooks introduced in #2424. The flow:

1. Env owns the data (lookup tables, state).
2. `ConversationSynthesizer` queries the env via `env.sample_grounding(...)` per conversation.
3. Sampled facts get rendered and injected into the planner prompt.

Without this PR, the grounding hooks in #2424 are unused infrastructure. This is what closes the loop.

## What's new on `ConversationSynthesizer`

- **`_make_grounding_rng(seed, sample_index)`** — per-sample deterministic `Random` when `GroundingConfig.seed` is set; unseeded `Random()` otherwise. The `(seed, sample_index)` pair gives stable but distinct RNGs across samples in the same run.
- **`_attach_grounding_facts(samples, attribute)`** — walks every env in scope (honoring `available_environments` and `available_tools` filters), calls `env.sample_grounding(...)` per env, writes the concatenated facts to `sample["grounding_facts"]`. Logs one warning per env per `synthesize()` call when `sample_size > pool_size`.
- **Planner prompt grounding injection** — when `sample["grounding_facts"]` is non-empty, `_create_planner_prompt` appends:
  - The bulleted fact list via `describe_grounding_default(...)`.
  - A "Grounding rules (role-aware)" block: identifiers belong in user turns, status fields are preconditions (plan a recovery flow when a fact contradicts the conversation intent), assistant turns must NOT pre-resolve tool output.
- **`_warn_on_grounding_placeholder`** — warns once if a user/assistant persona template inlines `{grounding_facts}`. Grounding is planner-only; leaking the fact list to user/assistant personas defeats its purpose.

## Tests

20 new tests in `test_conversation_synthesizer.py`:
- `_make_grounding_rng` reproducibility (seeded/unseeded, varies-across-samples).
- `_attach_grounding_facts` no-op paths (no env config, no grounded env), scoping by environments and tools, multi-env concatenation, truncation warnings.
- Planner prompt injection: block present when facts exist, omitted when absent or empty, state-aware branching nudge included.
- Integration through `synthesize()`.
- `_warn_on_grounding_placeholder` for user/assistant personas (warns) vs no placeholder (silent).

275 tests pass across the affected directories.

## Open follow-ups (from review comments)

- The "Grounding rules (role-aware)" block has library-domain examples baked in (`book B007`, `borrowed/overdue`, `borrower_name`). Worth generalizing or parameterizing before a non-library env lands.
- The grounding rules prose is a 30-line string literal in `_create_planner_prompt`. Worth extracting to a module-level constant for testability and prompt iteration.

## Series context

3. Grounding system: types + env-owned lookup tables — #2424 (in review)
4. **Grounding in the synthesis pipeline** ← this PR
7. Library tool-use synthesis example + e2e

## Before submitting

- [x] Tests for `_make_grounding_rng`, `_attach_grounding_facts` (no-op paths, scoping, multi-env, truncation), planner prompt injection (block present/absent, state-aware nudge), integration through `synthesize()`, `{grounding_facts}` placeholder warning.

### DIFF
--- a/src/oumi/core/synthesis/conversation_synthesizer.py
+++ b/src/oumi/core/synthesis/conversation_synthesizer.py
@@ -15,10 +15,13 @@
 import dataclasses
 import random
 
+from oumi.builders.environments import build_environment
 from oumi.builders.inference_engines import build_inference_engine
 from oumi.core.configs.environment_config import EnvironmentConfig
 from oumi.core.configs.inference_config import InferenceConfig
 from oumi.core.configs.inference_engine_type import InferenceEngineType
+from oumi.core.configs.params.environment_params import EnvironmentParams
+from oumi.core.configs.params.grounding_params import GroundingConfig
 from oumi.core.configs.params.guided_decoding_params import GuidedDecodingParams
 from oumi.core.configs.params.synthesis_params import (
     GeneralSynthesisParams,
@@ -32,6 +35,9 @@ from oumi.core.types.conversation import (
     Message,
     Role,
 )
+from oumi.environments import GroundingFact
+from oumi.environments.base_environment import BaseEnvironment
+from oumi.environments.utils import describe_grounding_default
 from oumi.utils.logging import logger
 from oumi.utils.str_utils import extract_json
 
@@ -128,6 +134,8 @@ class ConversationSynthesizer:
                 [tool.id for tool in available_tools],
             )
 
+        self._warn_on_grounding_placeholder(multiturn_attributes)
+        self._attach_grounding_facts(samples, multiturn_attributes)
         samples = self._plan_samples(samples, multiturn_attributes)
         conversations = self._synthesize_all_samples(samples, multiturn_attributes)
 
@@ -427,6 +435,39 @@ class ConversationSynthesizer:
         if role_context:
             base_prompt += f"\nRole context:\n{role_context}\n"
 
+        grounding_facts = sample.get("grounding_facts") or []
+        if grounding_facts:
+            block = describe_grounding_default(grounding_facts)
+            base_prompt += (
+                "\nGround this plan in these specific entities:\n"
+                f"{block}\n"
+                "Grounding rules (role-aware):\n"
+                "- USER turn instructions MAY inline concrete identifiers "
+                "from the list above (e.g. 'order ORD-4421 is late', "
+                "'book B007'). The user persona cannot see this list, so "
+                "identifiers the user should mention must be written into "
+                "their turn instruction.\n"
+                "- Treat each entity's non-identifier fields (e.g. status, "
+                "due_date, return_date) as preconditions. If a field's value "
+                "contradicts the conversation intent -- for example trying "
+                "to borrow a book whose status is 'borrowed' or 'overdue', "
+                "or trying to return one that is 'available' -- plan a "
+                "recovery flow that handles the conflict (offer an "
+                "alternative entity from the list, explain the conflict, ask "
+                "a clarifying question) instead of a happy-path that the "
+                "tool will reject.\n"
+                "- ASSISTANT turn instructions MUST NOT pre-resolve or "
+                "pre-state any tool output — no identifiers, statuses, "
+                "borrower names, due dates, or other facts the assistant "
+                "would normally look up. Reference entities by what the "
+                "user said (e.g. the title) and describe which TOOL the "
+                "assistant should call to resolve or verify. Example — "
+                "write 'call lookup_book_status with the book_id from the "
+                "catalog', not 'tell the user book B007 is checked out'.\n"
+                "- The planner's job for assistant turns is to probe the "
+                "right tool usage, not to do the tool's work.\n"
+            )
+
         if multiturn_attribute.conversation_planner:
             formatted_planner = self._formatter.format(
                 sample,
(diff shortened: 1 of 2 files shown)


aniruddh-alt:
## Review

### `conversation_synthesizer.py:437-438` — Inline imports (slop)

```python
from oumi.builders.environments import build_environment
from oumi.environments.utils import describe_grounding_default
```

I checked: `oumi.builders.environments` doesn't import the synthesizer, and neither does `oumi.environments.utils` — no circular dep that requires inline imports. Hoist both to the top of the file.

Same comment applies to the inline `from oumi.builders.environments import build_environment` in `_attach_grounding_facts` at line 732.

### `conversation_synthesizer.py:443-456` — Per-sample rebuild + forward-looking comment

`_create_planner_prompt` is called once per sample (and per planning retry). `build_environment(describer_env_params)` runs every call. `build_environment` reads the registry and instantiates a runtime — not free. `_attach_grounding_facts` already builds env runtimes for sampling, so this duplicates work.

**Suggestion:** build the describer env once at the start of `synthesize()` (or memoize on `self`) and thread it through.

The comment `# In v1 every env uses the default describer; future envs with custom describers should be revisited here.` is the same forward-looking slop pattern as the `ToolGroundingConfig` v1/v2 docstring on PR3 (which we trimmed). Either drop the comment or, better, drop the per-env-describer-pick branch entirely and always use `describe_grounding_default` until a custom-describer env actually exists. YAGNI.

**Multi-env correctness:** picking the *first* grounded env's describer means facts from env B render through env A's logic if A is listed first in the config. Moot today (all envs use default), but worth flagging alongside the cleanup.

### `conversation_synthesizer.py:678-687` — Verbose docstring

The cross-reference `matching the non-reproducible behavior used by DatasetPlanner for sampled attributes` will rot if `DatasetPlanner` evolves and stops using `Random()`. Trim to something self-contained:

```python
"""Build the per-sample RNG for grounding.

Seeded mode makes facts deterministic from `(seed + sample_index)`;
unseeded uses OS entropy.
"""
```

### `conversation_synthesizer.py:757` — Defensive assert

```python
grounding = env_params.grounding
assert grounding is not None  # filtered above
```

The list comprehension at lines 745-746 already filters out `env_params.grounding is None`, so by construction every entry has non-None grounding. Drop the assert; the `# filtered above` comment is the only thing it adds.

If you want pyright type-narrowing, bake the non-None config into the tuple type: `list[tuple[EnvironmentParams, GroundingConfig, BaseEnvironment]]`.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -427,6 +435,39 @@ def _create_planner_prompt(
...
+                "their turn instruction.\n"
+                "- Treat each entity's non-identifier fields (e.g. status, "
+                "due_date, return_date) as preconditions. If a field's value "
+                "contradicts the conversation intent -- for example trying "
+                "to borrow a book whose status is 'borrowed' or 'overdue', "
+                "or trying to return one that is 'available' -- plan a "
+                "recovery flow that handles the conflict (offer an "
+                "alternative entity from the list, explain the conflict, ask "
+                "a clarifying question) instead of a happy-path that the "
+                "tool will reject.\n"
```
aniruddh-alt, line 458: **Domain coupling:** the grounding rules block embeds library-specific examples — `book B007`, `order ORD-4421`, statuses `borrowed`/`overdue`/`available`, fields `borrower_name`/`due_date`/`return_date`/`borrow_date`. These were probably tuned for the library example, but for an env with a different domain (returns, billing, calendar, etc.) the planner sees irrelevant examples that may pull it off-task.

**Suggestion:** either parameterize the examples (sample one or two real fact identifiers from `grounding_facts` and substitute them in), or generalize the prose to drop concrete IDs/statuses and just describe the rules abstractly. The state-aware-branching nudge (treat status fields as preconditions, plan recovery flow on conflict) is the load-bearing part — the library examples are decoration that has hidden cost when the env isn't library-shaped.

Not blocking PR4 since it's the only env in flight, but worth a follow-up before this generalizes.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -427,6 +435,39 @@ def _create_planner_prompt(
...
+                "borrower names, due dates, or other facts the assistant "
+                "would normally look up. Reference entities by what the "
+                "user said (e.g. the title) and describe which TOOL the "
+                "assistant should call to resolve or verify. Example — "
+                "write 'call lookup_book_status with the book_id from the "
+                "catalog', not 'tell the user book B007 is checked out'.\n"
+                "- The planner's job for assistant turns is to probe the "
+                "right tool usage, not to do the tool's work.\n"
+            )
+
```
aniruddh-alt, line 470: **Code structure:** the multi-paragraph grounding rules prose is a 30-line string literal embedded in `_create_planner_prompt`. Hard to scan, hard to test in isolation, hard to A/B if you want to iterate on the prompt.

**Suggestion:** lift it to a module-level constant:

```python
_GROUNDING_RULES = (
    "Grounding rules (role-aware):\n"
    "- USER turn instructions MAY inline ...\n"
    ...
)
```

Then the planner-prompt code is `f"{block}\n{_GROUNDING_RULES}"` — clearly separates the *facts* (per-sample) from the *rules* (constant), and lets a unit test assert the rules text without exercising `_create_planner_prompt`'s full plumbing.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -427,6 +435,39 @@ def _create_planner_prompt(
         if role_context:
             base_prompt += f"\nRole context:\n{role_context}\n"
 
+        grounding_facts = sample.get("grounding_facts") or []
+        if grounding_facts:
+            block = describe_grounding_default(grounding_facts)
```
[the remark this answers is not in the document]
jgreer013: +1 here

jgreer013 (approved):

PR #2435, opened 2026-05-07
[Det Env v2] Phase 5: Native tool-call loop in synthesizer

aniruddh-alt:
# Description

Phase 5 of 6 in the Det Env v2 series.

This PR wires native tool calling into the conversation synthesizer. The
inference-engine layer (#2400 RemoteInferenceEngine, #2420 AnthropicInferenceEngine)
already round-trips structured `tool_calls`; this PR teaches the synth pipeline
to drive the agentic loop with that capability and projects results back into
`Conversation.tools` / `Message.tool_calls` / `Role.TOOL` natively.

## The headline change

Tool synthesis no longer round-trips through text-format `<tool_call>` blocks.
Engines that support native tool calling (OpenAI, Anthropic, vLLM, Remote, etc.)
get a structured loop: assistant turns produce a `Message` with `tool_calls`,
each call is dispatched to its owning environment, and `Role.TOOL` messages
are appended with `tool_call_id` matching. Engines that don't support it
(`NATIVE`, `LLAMACPP`, `BEDROCK`, `GOOGLE_GEMINI`) hard-error at init when an
environment exposes tools — fail fast over silent garbage.

## What's new

**Native tool-calling capability allowlist** (`oumi/inference/native_tool_calling.py`):
- `NATIVE_TOOL_CALLING_ENGINES`: frozenset of engine types known to round-trip
  `Conversation.tools` ↔ `Message.tool_calls` on the wire.
- `supports_native_tool_calling(engine_type) -> bool`: single source of truth.
- A test asserts the allowlist plus a manually-listed unsupported set covers
  the full `InferenceEngineType` enum, so adding a new engine forces an
  explicit classification.

**Synthesizer init-time gate** (`ConversationSynthesizer.__init__`):
- When `environment_config.all_tools` is non-empty and the configured engine
  isn't in the allowlist, raises `ValueError` listing the supported engines.
- Builds a `tool_id -> environment` dispatch table once; per-sample tool
  dispatch is then `self._tool_dispatch[name].step(name, args)`.

**Native tool-call loop** (`_run_assistant_turn_native`):
- Per-sample state (`base_msgs`, `staging`, `tool_count`, `done`) lets the
  loop progress different samples in lockstep without conflating their
  histories. `staging[idx]` is flushed to the real conversation only when
  the sample converges or hits the cap.
- The cap (`MultiTurnAttribute.max_tool_calls_per_turn`) is enforced by
  clamping each round's `tool_calls` slice to the remaining budget — not
  by truncating after the fact, so the assistant message and the tool
  responses stay paired.
- Stragglers (samples that still want to call tools at cap) get a single
  `_STRAGGLER_NUDGE` user message asking for a final answer, then one more
  inference round; whatever they produce is taken as the assistant turn.

**Tool dispatch** (`_run_tool_call`):
- Decodes `tool_call.function.arguments` as JSON; malformed JSON, non-dict
  args, unknown tool ids, and env exceptions all become `Role.TOOL` messages
  carrying a JSON `{"error": ...}` payload, so the model can see the failure
  and self-correct.
- Tool results with `dict` `output` are JSON-serialized; string outputs
  pass through as-is.

**Per-turn prompt construction**:
- Assistant prompts get `Conversation.tools` populated from
  `multiturn_attribute.available_environments` + `available_tools`.
- User prompts never get tools attached — the user persona shouldn't see
  the tool schemas.
- Tool resolution is hoisted out of the per-sample turn loop (was O(N_envs *
  N_tools) per call; now O(1) lookup via the dispatch table).

**`_has_empty_messages`** now treats an assistant message with `tool_calls`
as non-empty even when `content is None` — that's the OpenAI tool-wire shape.

## Public API changes

- New module `oumi.inference.native_tool_calling` with two public symbols.
- `ConversationSynthesizer.__init__` raises `ValueError` for an
  unsupported-engine + tools configuration. Existing callers without tools
  are unaffected.

## Notes

- 2134 unit tests pass, 82 skipped, 0 failed across `tests/unit/core/synthesis/`,
  `tests/unit/inference/`, `tests/unit/environments/`, `tests/unit/builders/`,
  `tests/unit/core/configs/`.
- Synth-test fixture switched from `InferenceEngineType.NATIVE` to `OPENAI`
  so the grounding tests added in #2424 (which build env_configs with tools)
  pass the new init-time check. The engine type was incidental — the
  inference engine is patched out via `build_inference_engine`.

## Series context

3. ~~Grounding system: types + env integration~~ — #2424 (merged, includes Phase 4)
4. ~~Grounding in the synthesis pipeline~~ — folded into #2424
5. **Native tool-call loop in synthesizer** ← this PR
6. Library tool-use synthesis example + e2e test — next

## Before submitting

- [x] Tests for the capability allowlist (parametrized over supported and
  unsupported engines, exhaustive enum coverage check).
- [x] Tests for the init-time engine-validation gate (raises on unsupported
  engine with tools, no error on supported engine with tools, no error on
  unsupported engine without tools).
- [x] Tests for the native tool-call loop, tool-call cap clamping, straggler
  nudge path, dispatch error projection.
- [x] Existing synth + grounding tests still pass.


src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -80,6 +115,57 @@ def _resolve_available_tools(
...
+                tool_call_id=tool_call.id,
+                content=json.dumps({"error": f"Unknown tool '{name}'"}),
+            )
+        try:
+            result = environment.step(name, arguments)
+        except Exception as exc:
+            return Message(
+                role=Role.TOOL,
+                tool_call_id=tool_call.id,
+                content=json.dumps({"error": f"Tool '{name}' raised: {exc}"}),
```
aniruddh-alt: **Slop (code duplication):** This method has four near-identical error returns at lines 130-134, 138-143, 145-149, and 153-157. Each is the same `Message(role=Role.TOOL, tool_call_id=tool_call.id, content=json.dumps({"error": ...}))` shape with a different message string.

Can we extract a small helper?

```python
def _tool_error(self, tool_call: ToolCall, msg: str) -> Message:
    return Message(
        role=Role.TOOL,
        tool_call_id=tool_call.id,
        content=json.dumps({"error": msg}),
    )
```

Each error path then becomes one line:

```python
return self._tool_error(tool_call, f"Malformed tool_call arguments: {exc}")
```

**Why:** The four error paths have a single common shape — a typo or accidental drift between them (e.g., wrong key name) is hard to spot today. With a helper the boundary contract lives in one place.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -621,6 +738,95 @@ def _synthesize_all_samples(
...
+                    for tc in calls_to_dispatch:
+                        staging[idx].append(self._run_tool_call(tc))
+                    tool_count[idx] += len(calls_to_dispatch)
+                else:
+                    content = (
+                        assistant_msg.content
+                        if assistant_msg is not None and assistant_msg.content
+                        else ""
+                    )
+                    if not isinstance(content, str):
```
aniruddh-alt: **Slop (double-defensive parsing + duplication):** This 7-line content-extraction block appears here in the main loop and again at lines 818-823 in the straggler branch. Both are double-defensive: the truthy check `if assistant_msg is not None and assistant_msg.content` would already make `content` falsy if the value was `None`/`""`/`[]`, then `if not isinstance(content, str)` re-checks. The two checks dodge the same edge cases.

Can we simplify and dedupe with a tiny helper?

```python
def _final_assistant_text(msg: Message | None) -> str:
    if msg is None or not isinstance(msg.content, str):
        return ""
    return msg.content
```

Both call sites then become:

```python
staging[idx].append(
    Message(role=Role.ASSISTANT, content=_final_assistant_text(assistant_msg))
)
```

**Why:** Less doubled defensive logic, single place to evolve when we want to handle multimodal `list[ContentItem]` content properly. Today the second-arm `if not isinstance(...)` silently throws away non-string content; if we ever decide that should be preserved, one place to change.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -621,6 +738,95 @@ def _synthesize_all_samples(
...
+        sample_indices: list[int],
+        histories: list[list[Message]],
+        max_tool_calls: int,
+        assistant_tools: list[ToolDefinition] | None,
+    ) -> None:
+        """Run an assistant turn using native tool calling, mutating histories."""
+        # Each round's prompt for sample ``idx`` is ``base_msgs[idx] + staging[idx]``;
+        # staging is flushed to histories only after the sample converges.
+        base_msgs: dict[int, list[Message]] = {
+            idx: list(prompt.messages) for idx, prompt in zip(sample_indices, prompts)
```
aniruddh-alt: **Slop (ceremonial shallow copy):** `list(prompt.messages)` defensively copies the message list into `base_msgs[idx]`, but `base_msgs[idx]` is never mutated — its only used as the left operand of `base_msgs[idx] + staging[idx]`, which builds a fresh list each time.

Can we just use `prompt.messages` directly?

```python
base_msgs: dict[int, list[Message]] = {
    idx: prompt.messages for idx, prompt in zip(sample_indices, prompts)
}
```

**Why:** The shallow copy doesnt protect against anything real — `infer()` doesnt mutate input prompt messages, and the concatenation never mutates the left side either. Its extra allocation per sample with no behavioral effect.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -583,12 +681,31 @@ def _synthesize_all_samples(
...
             if not prompts:
                 break
 
+            uniform_role = roles_for_turn[0]
+            if not all(r == uniform_role for r in roles_for_turn):
+                raise RuntimeError(
+                    "Mixed roles in a single turn are not supported by "
+                    "_synthesize_all_samples; the default turn order should "
+                    "produce uniform per-turn roles."
+                )
```
aniruddh-alt: **Slop (defensive check for impossible scenario):** `roles_for_turn` is built a few lines up by iterating `sample_indices` and assigning `role = turn_order[turn_idx % len(turn_order)]` to each — the value of `role` doesnt depend on the sample, so by construction every entry is identical. The `if not all(...)` here can only trip if someone changes the loop body to vary `role` per sample, in which case the surrounding logic (`uniform_role = roles_for_turn[0]`, the `if uniform_role == Role.ASSISTANT` dispatch below) would already be wrong.

Per the project preference to "trust internal code and framework guarantees" rather than validate scenarios that cant happen, can we drop the check and just use `uniform_role = roles_for_turn[0]` directly? If were worried about future drift, an inline comment on the role assignment ("# uniform across samples within a turn") is cheaper than a runtime validation.

I noticed this was previously an `assert` and got upgraded to `RuntimeError` — totally fair if the intent is to keep an active runtime guard. In that case ignore this. But if its defending against a hypothetical future regression, the value is low and the noise is real.

(agent)

tests/unit/core/synthesis/test_conversation_synthesizer.py
```diff
@@ -1469,3 +1512,750 @@ def test_warn_on_grounding_placeholder_no_warning_when_placeholder_absent(
...
+    synth.synthesize(
+        samples=[{"target_turns": 2, "parsed_turn_plans": []}],
+        multiturn_attributes=multiturn_attr,
+    )
+
+    assert all(p.tools is None for p in captured_prompts)
+
+
+# ---------------------------------------------------------------------------
+# Tests for native tool-call loop (Task 5)
```
aniruddh-alt: **Slop (planning artifact in comment):** "(Task 5)" is a leftover label from the implementation plan — it has no meaning in the merged codebase and will rot fast (the next refactor that splits this section wont know what "Task 5" was).

Can we drop the parenthetical?

```python
# ---------------------------------------------------------------------------
# Tests for native tool-call loop
# ---------------------------------------------------------------------------
```

**Why:** Per the project preference, comments shouldnt reference the task/fix/PR that produced them — that context belongs in the commit message and PR description, not the source.

(agent)

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: Fixed in 96ea016c9.

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: Fixed in 96ea016c9.

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: Fixed in 96ea016c9.

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: Fixed in 96ea016c9.

tests/unit/core/synthesis/test_conversation_synthesizer.py
aniruddh-alt: Fixed in 96ea016c9.

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -69,6 +80,30 @@ def __init__(
...
+        if self._environment_config is not None:
+            envs_by_id: dict[str, BaseEnvironment] = {
+                env_params.id: build_environment(env_params)
+                for env_params in self._environment_config.environments
+            }
+            for (
+                tool_id,
+                env_id,
+            ) in self._environment_config.tool_environment_map.items():
+                self._tool_dispatch[tool_id] = envs_by_id[env_id]
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 9f6801b24.

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -621,6 +738,95 @@ def _synthesize_all_samples(
...
+            for idx, result in zip(stragglers, results):
+                assistant_msg = result.messages[-1] if result.messages else None
+                content = (
+                    assistant_msg.content
+                    if assistant_msg is not None and assistant_msg.content
+                    else ""
+                )
+                if not isinstance(content, str):
+                    content = ""
+                staging[idx].append(Message(role=Role.ASSISTANT, content=content))
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 9f6801b24.

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -311,6 +399,8 @@ def _has_empty_messages(self, conversation: Conversation) -> bool:
         for message in conversation.messages:
             if message.role == Role.SYSTEM:
                 continue
+            if message.tool_calls:
+                continue
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 9f6801b24.

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -583,12 +681,31 @@ def _synthesize_all_samples(
                 )
                 prompt_messages.append(Message(role=Role.USER, content=turn_info))
 
-                prompts.append(Conversation(messages=prompt_messages))
+                tools = assistant_tools if role == Role.ASSISTANT else None
+                prompts.append(Conversation(messages=prompt_messages, tools=tools))
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 9f6801b24.

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -621,6 +738,95 @@ def _synthesize_all_samples(
...
+                    messages=base_msgs[idx] + staging[idx],
+                    tools=assistant_tools,
+                )
+                for idx in active
+            ]
+            results = self._inference_engine.infer(
+                active_prompts,
+                inference_config=self._inference_config,
+            )
+            for idx, result in zip(active, results):
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 9f6801b24.

src/oumi/core/configs/params/synthesis_params.py
```diff
@@ -485,9 +485,11 @@ class MultiTurnAttribute:
     available_tools: list[str] = field(default_factory=list)
     """List of tool ids available in this conversation."""
 
-    max_tool_calls_per_turn: int = 50
-    """Safety ceiling for tool calls per ASSISTANT turn. The agent naturally stops
-    when it decides no more tools are needed. This only prevents runaway loops."""
+    max_tool_rounds_per_turn: int = 50
+    """Safety ceiling on assistant→tool→assistant rounds within a single assistant
```
aniruddh-alt: **Slop (bloated docstring):** This is 4 lines (with line breaks counted) for what was previously a 2-line docstring. The "Each round is one inference call that emits tool_calls (parallel batches count as one round)" sentence repeats what the first line already implies.

Can we tighten?

```python
max_tool_rounds_per_turn: int = 50
"""Circuit breaker on assistant→tool round-trips per turn (parallel batches = one round)."""
```

Or if we want to keep the WHY explicit:

```python
max_tool_rounds_per_turn: int = 50
"""Cap on assistant→tool round-trips per turn; parallel batches count as one.
A high circuit breaker, not an active limit — the agent stops naturally when no more tools are needed."""
```

**Why:** The semantic detail ("parallel batches count as one round") is worth keeping in one line, but the third clause about "the agent naturally stops" is restating standard agent-loop behavior — readers dont need that context here.

(agent)

tests/unit/core/synthesis/test_conversation_synthesizer.py
```diff
@@ -1469,3 +1510,715 @@ def test_warn_on_grounding_placeholder_no_warning_when_placeholder_absent(
...
+    )
+    with patch.object(
+        synth, "_resolve_available_tools", return_value=env_config.all_tools
+    ):
+        synth.synthesize(
+            samples=[{"target_turns": 2, "parsed_turn_plans": []}],
+            multiturn_attributes=multiturn_attr,
+        )
+
+    # Model emitted 5 parallel calls in one round; all 5 dispatched (no clamping).
```
aniruddh-alt: **Slop (restating-code comment):** This comment restates the assertion on the next line (`assert fake_env.step.call_count == 5`) and the test name (`test_assistant_turn_dispatches_parallel_batch_unrestricted`). Both already convey "5 parallel calls, all dispatched, no clamping."

Can we drop it?

```python
    assert fake_env.step.call_count == 5
```

**Why:** Per project preference, comments shouldnt restate what an assertion or function name already says.

(agent)

src/oumi/core/configs/params/synthesis_params.py
aniruddh-alt: Fixed in 647790253.

tests/unit/core/synthesis/test_conversation_synthesizer.py
aniruddh-alt: Fixed in 647790253.

src/oumi/core/configs/params/synthesis_params.py
```diff
@@ -485,9 +485,8 @@ class MultiTurnAttribute:
     available_tools: list[str] = field(default_factory=list)
     """List of tool ids available in this conversation."""
 
-    max_tool_calls_per_turn: int = 50
-    """Safety ceiling for tool calls per ASSISTANT turn. The agent naturally stops
-    when it decides no more tools are needed. This only prevents runaway loops."""
+    max_tool_rounds_per_turn: int = 50
```
jgreer013: nit: max_consecutive_tool_turns

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -621,6 +715,154 @@ def _synthesize_all_samples(
...
+                )
+            else:
+                staging[idx].append(
+                    Message(
+                        role=Role.ASSISTANT,
+                        content=self._final_assistant_text(assistant_msg),
+                    )
+                )
+
+    def _run_assistant_turn(
```
jgreer013: nit: _run_assitant_agentic_loop

single "turn" isn't quite right since this feeds back into itself I think?

jgreer013 (approved):

PR #2444, opened 2026-05-10
[Det Env v2] Phase 6: Library tool-use synthesis example

aniruddh-alt:
# Description

Part of the **Det Env v2** PR chain (Phase 6 of 6) — Stage: example + e2e.

**What changed**: adds `configs/examples/synthesis/library_tool_use_synth.yaml`, an end-to-end library-assistant synthesis config that exercises the deterministic environment + native tool-calling stack from Phases 1–5.

**Why**: gives users a runnable example of the full det-env v2 surface — planner with guided JSON, deterministic env with lookup tables, and native `tool_calls`/`Role.TOOL` round-tripping — and serves as a smoke test for the chain.

## What the example covers

- **Two tools** — `list_book_catalog` (no args) and `lookup_book_status(book_id)`, served by a deterministic env's `lookup_table`.
- **Three personas** — `anxious_parent`, `enthusiastic_reader`, `busy_professional`, sampled per conversation.
- **Multi-turn dialog** — `library_dialog` with 4–6 turns, `max_tool_rounds_per_turn: 6`.
- **Planner-driven** — the synth pipeline plans turn-by-turn instructions before running the user/assistant simulation.

## Prompt rules driven by judge findings

I ran the example at n=150 with gpt-4o-mini, then judged each conversation on four axes (tool-call correctness, persona fidelity, conversation coherence, plan adherence) using gpt-4o. Two prompt rules in the committed config were chosen to address the failure modes I saw:

- **ASSISTANT system prompt** forbids `lookup_book_status` with a `book_id` not seen in a recent `list_book_catalog` response. Without this, LibBot fabricated `B001` lookups when patrons asked for off-catalog titles ("The Silent Patient", "Gone Girl"), causing ~6 of 11 tool-call failures and clustering with coherence failures.
- **`conversation_planner`** receives `{user_persona.description}` and an explicit instruction to match plan length/style to persona. Without this, the planner generated chitchat-heavy plans that terse personas couldn't follow — `busy_professional` failed plan-adherence at 76%, vs. 12–18% for the other two personas.

## PR Chain Navigation

- **Previous PR**: #2435 (Phase 5: Native tool-call loop in synthesizer)
- _This is the final PR in the v2 chain._

## Series context

| Phase | Title | Status |
|---|---|---|
| 1 | Tool schema unification & validation | ✅ #2421 |
| 2 | Planner: guided JSON decoding | ✅ #2422 |
| 3+4 | Grounding system | ✅ #2424 |
| 5 | Native tool-call loop | DRAFT #2435 |
| 6 | Library tool-use example (this PR) | DRAFT |

## Before submitting

- [ ] Did you add / update tests where needed? — e2e test pending in a follow-up; this PR is config-only.

> [!NOTE]
> **Liberate**
> Risk: low


aniruddh-alt (commented):
## Overall

Clean, focused PR — one new example config that ties together the v2 chain (deterministic env + native tool calling + planner). Nice that the prompt rules are anchored to specific judge findings; that's better than guessing.

A few structural things worth following up:

**Missing README update**: `configs/examples/synthesis/README.md` lists 7 examples; this PR adds an 8th but doesn't update the README. The other examples have a numbered entry with **Purpose** / **What it does** / **Key features** / Run-with block, plus a folded `<details>` example output. Worth adding before this lands. Probably a separate small PR or folded into this one.

**E2E test deferred**: PR body acknowledges this; just flagging that the chain memory had `~700-800 LOC, mostly config data` for PR6 with an e2e test expected. Worth opening a follow-up issue or scoping a Phase 7 if the test actually needs to land.

Inline comments below on a few smaller items.

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,165 @@
...
+inference_config:
+  engine: OPENAI
+  model:
+    model_name: gpt-4o-mini
+  generation:
+    max_new_tokens: 512
+    temperature: 0.4
+  remote_params:
+    num_workers: 10
+    politeness_policy: 60
```
aniruddh-alt: **Default polish:** `politeness_policy: 60` is very conservative as an example default — empirically a 150-sample run takes ~50 min because it bottlenecks adaptive concurrency to one update per minute.

**Suggestion:** lower this to `0` (or omit it entirely) for the example. Users hitting real rate-limit issues can raise it themselves; the default shouldn't be slower than necessary.

**Why:** examples are the first thing new users run. A sub-5-minute synth run lands much better than a 50-minute one, and the comment-block already points to docs for tuning.

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,165 @@
...
+#
+# Requirements:
+#   - Set OPENAI_API_KEY in the environment.
+#   - Or change the inference_config to your preferred native-tool-calling engine
+#     (Anthropic, vLLM, Remote, Fireworks, etc. — see oumi/inference/native_tool_calling.py).
+#
+# Usage:
+#   oumi synth -c oumi://configs/examples/synthesis/library_tool_use_synth.yaml
+#
+# See Also:
```
aniruddh-alt, line 16: **Sibling conformance:** the other synthesis example configs (e.g. `multiturn_conversation_synth.yaml:15-21`) have a 6-entry `See Also` block with `Documentation` / `Config class` / `Params class` / `Config source` (GitHub) / `Params source` (GitHub) / `Other synthesis configs`.

**Suggestion:** match that shape — keep the `EnvironmentConfig` reference (it's load-bearing for this PR's semantics) but add the missing source links so the docs surface is consistent across the directory.

**Why:** the `eng:sibling-study` instinct here — when adding a new variant of a system, the closest siblings define the conventions. Easier to grep across configs when the metadata blocks line up.

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,165 @@
+# Synthesize multi-turn library-assistant conversations that exercise tool calling.
+#
+# A library patron (one of three personas) chats with LibBot, an assistant that
+# uses two tools to help: list_book_catalog and lookup_book_status. The
+# deterministic environment serves the tool calls from a small lookup table.
+#
+# Requirements:
+#   - Set OPENAI_API_KEY in the environment.
+#   - Or change the inference_config to your preferred native-tool-calling engine
+#     (Anthropic, vLLM, Remote, Fireworks, etc. — see oumi/inference/native_tool_calling.py).
```
aniruddh-alt: **Engine list:** the parenthetical "(Anthropic, vLLM, Remote, Fireworks, etc.)" is a stale-prone enumeration — `oumi/inference/native_tool_calling.py:NATIVE_TOOL_CALLING_ENGINES` is the actual source of truth and Phase 5's whole point.

**Suggestion:** drop the example list; just say "see `NATIVE_TOOL_CALLING_ENGINES` in `oumi/inference/native_tool_calling.py` for supported engines." Then there's exactly one place to update when an engine is added or removed.

**Why:** if someone adds an engine to the allowlist tomorrow, this comment silently goes stale. The current text says "etc." which technically covers it, but that's the kind of comment we'd flag as slop in a code review.

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
aniruddh-alt: Fixed in e418a9247.

configs/examples/synthesis/library_tool_use_synth.yaml
aniruddh-alt: Fixed in e418a9247.

configs/examples/synthesis/library_tool_use_synth.yaml
aniruddh-alt: Fixed in e418a9247.

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,169 @@
...
+            properties: {}
+        - id: lookup_book_status
+          name: lookup_book_status
+          description: Look up the borrow status of a book by book_id.
+          parameters:
+            type: object
+            properties:
+              book_id:
+                type: string
+                description: "Catalog ID of the book, e.g. 'B001'."
```
aniruddh-alt: **Subtle catalog leak:** the `e.g. 'B001'` in this parameter description gets serialized into the tool schema sent to the assistant model on every turn. So the assistant *sees* `B001` before it has called `list_book_catalog`.

Recall the judge findings from the 150-sample run — the most common tool-call failure was LibBot calling `lookup_book_status({"book_id": "B001"})` when the patron asked for an off-catalog title. The ASSISTANT prompt rules now block that, but this example value is plausibly the seed.

**Suggestion:** drop the example: `description: "Catalog ID of the book (from list_book_catalog)."`

**Why:** with no example, the assistant has no priming toward a specific ID. Pointer to the source-of-truth (`list_book_catalog`) reinforces the prompt rule.

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,169 @@
...
+      tools:
+        - id: list_book_catalog
+          name: list_book_catalog
+          description: List all books in the library catalog with their book_ids.
+          parameters:
+            type: object
+            properties: {}
+        - id: lookup_book_status
+          name: lookup_book_status
+          description: Look up the borrow status of a book by book_id.
```
aniruddh-alt: **Understated output:** the tool actually returns `{title, status, due_date?}` (see `env_kwargs.lookup_table.lookup_book_status` below) but the description only mentions `status`. The assistant has to infer that `title` and `due_date` are part of the response.

**Suggestion:** "Look up borrow status for a book by book_id. Returns title, status, and (when checked out) due_date."

**Why:** tool descriptions are the model's only spec for the response shape — undersold descriptions push the model toward generic responses ("the book is available") instead of grounded ones ("'The Great Gatsby' is checked out, due 2026-05-15").

(agent)

configs/examples/synthesis/library_tool_use_synth.yaml
```diff
@@ -0,0 +1,169 @@
...
+    - user_persona
+    - library_dialog
+    - library_dialog_plan
+
+inference_config:
+  engine: OPENAI
+  model:
+    model_name: gpt-4o-mini
+  generation:
+    max_new_tokens: 512
```
aniruddh-alt: **Tight headroom:** `max_new_tokens: 512` is on the low end for multi-turn with tool calls — a single assistant turn can include 1–2 tool calls (each call's JSON arguments count) plus a natural-language summary. The 150-sample run finished fine, but the headroom is thin.

**Suggestion:** bump to `1024` (or `2048`) — matches `multiturn_conversation_synth.yaml` and avoids silent truncation on chattier personas.

**Why:** silent truncation in the middle of a tool-call response is hard to spot in logs but degrades downstream quality (the assistant message ends mid-sentence and the next turn starts from that broken state).

(agent)

aniruddh-alt (commented):
## Re-review (post PR5 merge)

PR is in good shape:
- ✅ CI green (`static-checks` + CodeQL)
- ✅ Base correctly retargeted to `main` now that PR5 (#2435) merged
- ✅ All 3 prior review threads still resolved (politeness_policy, See Also, engine list)
- ✅ Two clean files, +197/-0, both contained to `configs/examples/synthesis/`

Three new observations from a fresh pass — all minor, focused on tool-schema details that subtly shape the assistant's behavior. Inline above.

The skipped `<details>` example output in the README entry is a defensible omission (the sibling entries include one but it's not strictly required by README's "Available Examples" enumeration). Worth a follow-up if we want full sibling conformance, but not blocking.

(agent)

jgreer013 (approved):

PR #2452, opened 2026-05-12
feat: add LiteLLM as AI gateway inference engine

RheagalFire:
# Description

Adds LiteLLM as a new inference engine, giving users access to 100+ LLM providers (Cohere, AI21, Perplexity, SambaNova, etc.) through a single engine type.

LiteLLM is a lightweight Python SDK that calls provider APIs directly using your own keys. Unlike adding a dedicated inference engine for each provider, LiteLLM routes requests based on the model string (e.g. `anthropic/claude-sonnet-4-5`, `bedrock/anthropic.claude-v2`, `vertex_ai/gemini-pro`). This is useful for:

- Providers without dedicated Oumi inference engines (Cohere, AI21, Perplexity, etc.)
- Swapping providers without changing code (just change the model string)
- Self-hosted endpoints (vLLM, TGI, custom OpenAI-compatible servers)

### Changes

- `src/oumi/inference/litellm_inference_engine.py` - new `LiteLLMInferenceEngine` extending `BaseInferenceEngine`
- `src/oumi/inference/__init__.py` - registered in exports
- `src/oumi/builders/inference_engines.py` - registered in `ENGINE_MAP`
- `src/oumi/core/configs/inference_engine_type.py` - added `LITELLM` enum value
- `pyproject.toml` - added `litellm>=1.55.0,<1.85` as optional dependency (`pip install oumi[litellm]`)

### Implementation

Extends `BaseInferenceEngine` directly (not `RemoteInferenceEngine`) since LiteLLM is an SDK, not an HTTP endpoint. The engine implements:

- `_build_api_input()` - converts Oumi `Conversation` objects to the `litellm.acompletion()` kwargs format using the existing `create_list_of_message_json_dicts` utility
- `_parse_response()` - converts the OpenAI-compatible response back to a `Conversation` with usage and finish_reason metadata
- `_infer_online()` - async execution via `litellm.acompletion()` with `drop_params=True` to silently drop provider-unsupported kwargs

### Example usage

```python
from oumi.core.configs import ModelParams, GenerationParams, InferenceEngineType
from oumi.builders import build_inference_engine
from oumi.core.types.conversation import Conversation, Message, Role

engine = build_inference_engine(
    engine_type=InferenceEngineType.LITELLM,
    model_params=ModelParams(model_name='anthropic/claude-sonnet-4-5'),
    generation_params=GenerationParams(max_new_tokens=500, temperature=0.7),
)

conv = Conversation(messages=[
    Message(content='What is machine learning?', role=Role.USER),
])
results = engine.infer([conv])
print(results[0].messages[-1].content)
```

Or via config YAML:

```yaml
inference:
  engine: LITELLM
  model:
    model_name: anthropic/claude-sonnet-4-5
  generation:
    max_new_tokens: 500
```

### Tests

Live E2E test against Anthropic Claude (`anthropic/claude-sonnet-4-5`):

```python
>>> engine = LiteLLMInferenceEngine(
...     model_params=ModelParams(model_name='anthropic/claude-sonnet-4-5'),
...     generation_params=GenerationParams(max_new_tokens=100, temperature=0.0),
... )
>>> conv = Conversation(messages=[Message(content='What is 2+2? Answer with just the number.', role=Role.USER)])
>>> results = engine.infer([conv])
>>> results[0].messages[-1].content
'4'
>>> results[0].metadata
{'usage': {'prompt_tokens': 20, 'completion_tokens': 5, 'total_tokens': 25}, 'finish_reason': 'stop'}
```

Builder path also verified:

```python
>>> engine = build_inference_engine(
...     engine_type=InferenceEngineType.LITELLM,
...     model_params=ModelParams(model_name='anthropic/claude-sonnet-4-5'),
...     generation_params=GenerationParams(max_new_tokens=100, temperature=0.0),
... )
>>> type(engine).__name__
'LiteLLMInferenceEngine'
>>> conv = Conversation(messages=[Message(content='Say hello in French. One word only.', role=Role.USER)])
>>> engine.infer([conv])[0].messages[-1].content
'Bonjour'
```

## Related issues

N/A - new feature addition.

## Before submitting


oelachqar:
thank you @RheagalFire for the contribution! Will review shortly

src/oumi/inference/litellm_inference_engine.py
```diff
@@ -0,0 +1,229 @@
...
+
+_FINISH_REASON_MAP = {
+    "stop": FinishReason.STOP,
+    "length": FinishReason.LENGTH,
+    "tool_calls": FinishReason.TOOL_CALLS,
+    "content_filter": FinishReason.CONTENT_FILTER,
+}
+
+
+class LiteLLMInferenceEngine(BaseInferenceEngine):
```
aniruddh-alt: could we inherit RemoteInferenceEngine instead of BaseInferenceEngine since LiteLLM is openai compatible format?

ccing @oelachqar for thoughts!

RheagalFire:
Thanks @aniruddh-alt for the suggestion. I considered RemoteInferenceEngine, but the challenge is that LiteLLM uses its own SDK (litellm.completion()) rather than raw HTTP requests, so inheriting  RemoteInferenceEngine would mean overriding _query_api entirely. Happy to refactor if @oelachqar confirms that's the preferred approach.

aniruddh-alt:
> Thanks @aniruddh-alt for the suggestion. I considered RemoteInferenceEngine, but the challenge is that LiteLLM uses its own SDK (litellm.completion()) rather than raw HTTP requests, so inheriting RemoteInferenceEngine would mean overriding _query_api entirely. Happy to refactor if @oelachqar confirms that's the preferred approach.

Thanks for the question. You're right that LiteLLM uses its SDK rather than raw HTTP but that does not need to be a blocker. BedrockInferenceEngine is a good precedent: it extends RemoteInferenceEngine and overrides _query_api to call boto3 instead of HTTP. Overriding _query_api is expected and fine.
I'd still recommend inheriting from RemoteInferenceEngine so we get parallel requests (num_workers), retries, rate limiting, and shared request/response conversion for free.

aniruddh-alt:
Hi @RheagalFire wanted to follow up on this PR.

RheagalFire:
@aniruddh-alt thanks for the heads up, will get this done today.

RheagalFire:
@aniruddh-alt Thanks for the follow-up and sorry for the delay! I've refactored the engine to inherit from RemoteInferenceEngine following BedrockInferenceEngine as the template, addressing all review feedback.

src/oumi/inference/litellm_inference_engine.py
```diff
@@ -0,0 +1,428 @@
...
+            for attempt in range(remote_params.max_retries + 1):
+                try:
+                    if attempt > 0:
+                        delay = min(
+                            remote_params.retry_backoff_base * (2 ** (attempt - 1)),
+                            remote_params.retry_backoff_max,
+                        )
+                        await asyncio.sleep(delay)
+
+                    response = await asyncio.to_thread(
```
aniruddh-alt, line 301: This override never calls `self._rate_limiter`, so `requests_per_minute`, `input_tokens_per_minute`, and `output_tokens_per_minute` from `RemoteParams` get built in the base `__init__` and then ignored. The base `_query_api` calls `wait_if_needed()` before the request and `record_usage()` after (`remote_inference_engine.py` L787-788 and L847-850).

Please add the same two calls here, each guarded by `if self._rate_limiter is not None`.

src/oumi/inference/litellm_inference_engine.py
```diff
@@ -0,0 +1,428 @@
...
+            api_input: The keyword arguments for litellm.completion().
+
+        Returns:
+            Dict[str, Any]: The response as a JSON-serializable dictionary.
+        """
+        response = litellm.completion(**api_input)
+        return response.model_dump(mode="json")
+
+    @override
+    def _convert_api_output_to_conversation(
```
aniruddh-alt: LiteLLM returns OpenAI-compatible responses, so this whole method can be removed and the base `_convert_api_output_to_conversation` used instead. The base version keeps two fields this one drops. `reasoning_content` is needed for reasoning models like o1 and DeepSeek-R1, and `cached_tokens` is used for prompt-cache accounting.

Removing it also lets you delete the module-level `_FINISH_REASON_MAP`.

aniruddh-alt:
Hi @RheagalFire following up on this!

RheagalFire:
@aniruddh-alt, I have addressed both points. For the rate limiter, _query_api now calls self._rate_limiter.wait_if_needed() before the request and record_usage(...) after it, each guarded by if self._rate_limiter is not None, matching the base _query_api. 

I also removed the _convert_api_output_to_conversation override and _FINISH_REASON_MAP, so the engine now uses the base conversion and keeps reasoning_content and cached_tokens. Ready for another look.

aniruddh-alt (commented):
A few targeted follow-ups from the latest CI run and current-main comparison:

pyproject.toml
```diff
@@ -134,6 +134,10 @@ gpu = [
     "kernels>=0.11,<0.15",                      # For compute kernel implementations (e.g., attn_implementation in gold trainer)
 ]
 
+litellm = [
+    "litellm>=1.55.0,<1.85",
```
aniruddh-alt: Could we refresh and justify this range? `>=1.55.0` admits known compromised/vulnerable releases, while `<1.85` excludes current stable LiteLLM. Please use the oldest version we actually test (at least `1.84.10`) and an upper bound matching the latest validated minor. References: https://github.com/BerriAI/litellm/issues/24518 and https://github.com/BerriAI/litellm/releases/tag/v1.95.0

tests/unit/inference/test_litellm_inference_engine.py
```diff
@@ -0,0 +1,143 @@
...
+    """Create LiteLLMInferenceEngine with mocked litellm."""
+    model_params = ModelParams(model_name="anthropic/claude-sonnet-4-5")
+    remote_params = RemoteParams(api_key="test_api_key")
+    return LiteLLMInferenceEngine(
+        model_params=model_params,
+        remote_params=remote_params,
+    )
+
+
+@pytest.mark.skipif(litellm_import_failed, reason="litellm not available")
```
aniruddh-alt: These tests mock LiteLLM, but every test is skipped when the optional package is absent and the CPU CI extra does not install it. Could we remove the skip from mock-only tests and add direct coverage for `_query_api` (completion call, retries, rate limiting, and response conversion)? `test_infer_online` currently mocks `_infer` itself, so it does not exercise the new engine.

src/oumi/inference/litellm_inference_engine.py
```diff
@@ -0,0 +1,371 @@
...
+    ) -> dict[str, Any]:
+        """Synchronously invokes litellm.completion.
+
+        Args:
+            api_input: The keyword arguments for litellm.completion().
+
+        Returns:
+            Dict[str, Any]: The response as a JSON-serializable dictionary.
+        """
+        response = litellm.completion(**api_input)
```
aniruddh-alt: `RemoteParams` accepts `api_key`, `api_key_env_varname`, `api_url`, and `connection_timeout`, but none are forwarded to `litellm.completion()`. This makes YAML credentials and endpoint overrides silently ineffective. Could we map them to LiteLLM arguments (`api_key`, `api_base`, and `timeout`) and test those overrides?

src/oumi/inference/litellm_inference_engine.py
```diff
@@ -0,0 +1,371 @@
...
+                        usage = self._extract_usage_from_response(response)
+                        if usage:
+                            await self._rate_limiter.record_usage(
+                                input_tokens=usage.get("prompt_tokens", 0),
+                                output_tokens=usage.get("completion_tokens", 0),
+                            )
+                    result = self._convert_api_output_to_conversation(
+                        response, conversation
+                    )
+                    if output_path:
```
aniruddh-alt: Could this call be unconditional, matching `RemoteInferenceEngine`? The base helper selects a temporary scratch path when `output_path` is `None`; this guard disables resume/checkpoint behavior and can repeat completed paid requests after a partial failure.

src/oumi/core/configs/inference_engine_type.py
```diff
@@ -72,6 +72,9 @@ class InferenceEngineType(str, Enum):
     CEREBRAS = "CEREBRAS"
     """The inference engine for Cerebras API."""
 
+    LITELLM = "LITELLM"
```
aniruddh-alt, line 75: This new enum needs downstream capability classification. Installation CI currently fails because `LITELLM` is absent from the native-tool-calling supported/unsupported sets. The request converter also drops tools and guided decoding, while `SimpleJudge` assumes strict guided decoding for every non-Bedrock engine. Could we either implement those OpenAI-compatible arguments or explicitly classify LiteLLM as unsupported in both paths?

src/oumi/inference/litellm_inference_engine.py
aniruddh-alt: Thanks for adding this. One remaining difference from current `RemoteInferenceEngine`: `wait_if_needed()` is inside its retry loop immediately before each request, whereas this calls it once before all attempts and holds the semaphore while sleeping. Could you mirror the base placement so retries are paced and do not occupy a worker during backoff?

RheagalFire:
@aniruddh-alt Thanks for the review. Pushed addressing every point.


- Implemented the args instead of marking it unsupported. Converter delegates to `RemoteInferenceEngine`, so `tools` and `response_format` go through, `LITELLM` is in `NATIVE_TOOL_CALLING_ENGINES` (fixes install CI), and `drop_params=True` drops what a backend cannot take. SimpleJudge stays as is.
- `RemoteParams` `api_key` / `api_url` / `connection_timeout` now map to LiteLLM's `api_key` / `api_base` / `timeout`.
- `_query_api` saves unconditionally, so resume works when `output_path` is None.
- `wait_if_needed()` moved before each attempt, backoff sleeps outside the semaphore.
- Pin is now `>=1.84.10,<1.96`.
- Dropped the `skipif` and added real `_query_api` tests (completion, retries, rate limiting, conversion, creds, guided decoding).
- pyright and formatting fixed.

PR #2457, opened 2026-05-14
[Synth Env] Phase 2 (core): Stateful tool execution in SyntheticEnvironment

aniruddh-alt:
## Description

Part of the **Stateful Tool Execution** PR chain (Wave 1 of 2) — core implementation.

Extends `SyntheticEnvironment` so the same class handles both stateless
LLM-simulated tools and stateful tools whose outputs are produced by
user-supplied Python executors. Selecting between the two modes is now
a config decision: pass `state_params` in `env_kwargs` to opt into
stateful behavior; omit it for the existing batched-LLM behavior from
Phase 1.

### What changes

- `ToolParams.executor` (new, optional): dotted import path to a
  Python callable. `SyntheticEnvironment` dispatches calls to it
  instead of LLM-simulating. Tools without an executor stay LLM-simulated.
- `SyntheticEnvironment.step()` routes per-call: stateful executor
  receives `(arguments, state)` with state threaded sequentially across
  the batch; stateless executor receives `(arguments,)`; everything
  else falls back to the existing batched simulator path.
- `ToolGroundingConfig` gets optional `key` + `state_path`. When
  `state_path` is set, the env projects grounding facts from
  `initial_state[state_path]` instead of a deterministic lookup table.
- Removes the standalone `StatefulEnvironment`, `StatefulTool`, and
  `EnvironmentStateParams` introduced earlier on this branch — their
  capabilities now live on `SyntheticEnvironment` / `ToolParams` /
  `SyntheticStateParams`. The env_type `stateful` no longer resolves;
  configs that used it should switch to `synthetic` with `state_params`.

### Validation

- `read_only=True` tools may not return `updated_state` (raises `ToolError`).
- `updated_state` is validated against `state_schema` before commit.
- Stateless executors returning `updated_state` are rejected.
- Grounding `state_path` is checked against `initial_state` at env
  construction time.

### Example usage

A worked example (EHR workflow + executors) lands in the follow-up PR — see the chain navigation below.

## PR Chain Navigation

- **Next PR**: #2458 (Examples: EHR stateful synthesis config + executors)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: medium

### DIFF
--- a/src/oumi/core/configs/params/grounding_params.py
+++ b/src/oumi/core/configs/params/grounding_params.py
@@ -43,16 +43,14 @@ class ToolGroundingConfig(BaseParams):
 class StateGroundingConfig(BaseParams):
     """Per-state-pool grounding for stateful synthetic environments.
 
-    Projects rows from ``initial_state[state_path]`` through ``fields``;
-    ``key`` names the primary-key field and must appear in ``fields``.
+    Projects rows from ``initial_state[state_path]`` through ``fields``.
     """
 
     state_path: str
     fields: list[str]
-    key: str
 
     def __post_init__(self) -> None:
-        """Validate ``state_path``, ``fields``, and ``key`` invariants."""
+        """Validate ``state_path`` and ``fields`` invariants."""
         if not self.state_path:
             raise ValueError(f"{type(self).__name__}.state_path must be non-empty.")
         if not self.fields:
@@ -61,13 +59,6 @@ class StateGroundingConfig(BaseParams):
             raise ValueError(
                 f"{type(self).__name__}.fields contains duplicate entries: "
                 f"{self.fields!r}."
-            )
-        if not self.key:
-            raise ValueError(f"{type(self).__name__}.key must be non-empty.")
-        if self.key not in self.fields:
-            raise ValueError(
-                f"{type(self).__name__}.fields must include 'key' "
-                f"({self.key!r}); got {self.fields!r}."
             )
 
 
--- a/src/oumi/core/configs/params/grounding_params.py
+++ b/src/oumi/core/configs/params/grounding_params.py
@@ -111,6 +111,12 @@ class GroundingConfig(BaseParams):
             else StateGroundingConfig(**cfg)
             for cfg in self.state
         ]
+        state_paths = [cfg.state_path for cfg in self.state]
+        if len(set(state_paths)) != len(state_paths):
+            raise ValueError(
+                f"{type(self).__name__}.state contains duplicate state_path "
+                f"entries: {state_paths!r}."
+            )
 
 
 @dataclass
--- a/src/oumi/core/configs/params/grounding_params.py
+++ b/src/oumi/core/configs/params/grounding_params.py
@@ -73,7 +73,12 @@ class StateGroundingConfig(BaseParams):
 
 @dataclass
 class GroundingConfig(BaseParams):
-    """Per-environment grounding configuration."""
+    """Per-environment grounding configuration.
+
+    Both sub-blocks are optional; envs read whichever applies. Deterministic
+    envs project from ``tools`` (per-tool lookup-table entries); stateful
+    synthetic envs project from ``state`` (per-pool ``initial_state`` rows).
+    """
 
     sample_size: int = 3
     """Number of grounding facts sampled per conversation."""
@@ -84,8 +89,11 @@ class GroundingConfig(BaseParams):
     tools: dict[str, ToolGroundingConfig] = field(default_factory=dict)
     """Per-tool field whitelists, keyed by tool id."""
 
+    state: list[StateGroundingConfig] = field(default_factory=list)
+    """Per-state-pool projections for stateful synthetic envs."""
+
     def __post_init__(self) -> None:
-        """Validate ``sample_size`` and coerce ``tools`` entries."""
+        """Validate ``sample_size`` and coerce ``tools``/``state`` entries."""
         if self.sample_size < 1:
             raise ValueError(
                 f"{type(self).__name__}.sample_size must be >= 1, "
@@ -97,6 +105,12 @@ class GroundingConfig(BaseParams):
             else ToolGroundingConfig(**cfg)
             for tool_id, cfg in self.tools.items()
         }
+        self.state = [
+            cfg
+            if isinstance(cfg, StateGroundingConfig)
+            else StateGroundingConfig(**cfg)
+            for cfg in self.state
+        ]
 
 
 @dataclass
--- a/src/oumi/core/configs/params/grounding_params.py
+++ b/src/oumi/core/configs/params/grounding_params.py
@@ -24,23 +24,50 @@ from oumi.core.configs.params.base_params import BaseParams
 
 @dataclass
 class ToolGroundingConfig(BaseParams):
-    """Per-tool field whitelist for grounding projection.
-
-    Tools listed in ``GroundingConfig.tools`` are projected through
-    ``fields`` when the env samples grounding facts. Tools not listed
-    contribute nothing.
-    """
+    """Per-tool field whitelist for deterministic-env grounding projection."""
 
     fields: list[str]
 
     def __post_init__(self) -> None:
-        """Validate ``fields`` invariants."""
+        """Validate ``fields`` is non-empty and de-duplicated."""
         if not self.fields:
             raise ValueError(f"{type(self).__name__}.fields must be non-empty.")
         if len(set(self.fields)) != len(self.fields):
             raise ValueError(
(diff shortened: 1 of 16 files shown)


src/oumi/core/configs/params/grounding_params.py
```diff
@@ -26,22 +26,28 @@
...
-    Tools listed in ``GroundingConfig.tools`` are projected through
-    ``fields`` when the env samples grounding facts. Tools not listed
-    contribute nothing.
+    When ``state_path`` is set, the env projects rows from
+    ``initial_state[state_path]`` instead of its lookup table.
     """
 
     fields: list[str]
+    key: str | None = None
+    state_path: str | None = None
```
aniruddh-alt: **Architecture:** State grounding mixed into ``ToolGroundingConfig`` couples two unrelated projection sources.

``ToolGroundingConfig`` now serves two callers with disjoint behavior:

1. **Deterministic env**: projects from ``lookup_table[tool_id]``, no ``state_path``.
2. **Synthetic + state**: projects from ``initial_state[state_path]``, ``state_path`` required.

The dispatch lives in ``SyntheticEnvironment.sample_grounding`` as ``if cfg.state_path is None: continue`` and the symmetric branch in ``DeterministicEnvironment.sample_grounding`` ignores ``state_path`` entirely. Two consumers, two type contracts on the same dataclass — a ``ToolGroundingConfig(fields=[...])`` on a synthetic env is silently a no-op, not a config error.

**Suggestion:** Split it. Keep ``ToolGroundingConfig`` as a deterministic-grounding leaf (``fields`` only) and reintroduce a focused ``StateGroundingConfig`` (``fields`` + ``key`` + required ``state_path``) for the stateful side. Either:

```python
# Option A — union on GroundingConfig.tools
class GroundingConfig:
    tools: dict[str, ToolGroundingConfig | StateGroundingConfig]
```

```python
# Option B — state grounding lives on the state, not on the env-grounding
@dataclass
class SyntheticStateParams:
    state_schema: dict | None = None
    initial_state: dict | None = None
    grounding: list[StateGroundingConfig] | None = None
```

I lean Option B — state grounding is part of the state definition, not a generic "tool grounding" concern, and it lets ``SyntheticEnvironment`` stop reaching into ``params.grounding.tools`` from inside ``_validate_state_grounding`` / ``sample_grounding``.

**Why:** Separation of concerns + "config typed wrong silently no-ops" is the kind of failure that bites long after merge.

(agent)

src/oumi/environments/synthetic_environment.py
```diff
@@ -163,57 +214,186 @@ def _lookup_tool(self, tool_id: str) -> ToolParams:
...
+                    jsonschema.validate(validated.updated_state, self._state_schema)
+                except jsonschema.ValidationError as e:
+                    raise ToolError(
+                        f"Tool '{tool_id}' updated_state failed state_schema "
+                        f"validation: {e}"
+                    ) from e
+            self._state = copy.deepcopy(validated.updated_state)
+        return validated
+
+    def _validate_executor_result(
```
aniruddh-alt: **Slop:** Boolean-flag parameter ``expect_state`` on ``_validate_executor_result`` branches a 4-line check; only two call sites with hardcoded values.

The method body has two responsibilities (output_schema validation + stateless-rejects-state check) gated by a boolean. The caller already knows which mode it's in, so the flag just makes the call site harder to scan.

**Suggestion:** Inline the ``updated_state``-must-be-None check into ``_step_executor_one`` and keep ``_validate_executor_result`` focused on output_schema:

```python
def _step_executor_one(self, tool_id, arguments):
    tool = self._lookup_tool(tool_id)
    tool.validate_arguments(arguments)
    result = self._executors[tool_id](arguments=arguments)
    self._validate_output(tool, result)
    if result.updated_state is not None:
        raise ToolError(f"Tool '{tool.id}' returned updated_state on stateless env.")
    return result
```

Saves one branch and one boolean argument; the two callers stay readable.

(agent)

src/oumi/environments/synthetic_environment.py
```diff
@@ -163,57 +214,186 @@ def _lookup_tool(self, tool_id: str) -> ToolParams:
...
+                continue
+            whitelist = set(cfg.fields)
+            for row in rows:
+                if not isinstance(row, dict):
+                    continue
+                projected = {k: v for k, v in row.items() if k in whitelist}
+                pool.append(GroundingFact(data=projected))
+        return rng.sample(pool, min(n, len(pool)))
+
+    def _validate_state_grounding(self) -> None:
```
aniruddh-alt, line 401: **Code duplication:** ``_validate_state_grounding`` and ``sample_grounding`` re-walk the same ``grounding.tools.items()`` filtered by ``cfg.state_path is None``.

The two methods share the same iteration shape:

```python
for tool_id, cfg in grounding.tools.items():
    if cfg.state_path is None:
        continue
    ...  # validate-or-project rows from self._state[cfg.state_path]
```

If state grounding moves to its own type (see other comment), the iteration becomes a single list walk and the ``state_path is None`` filter goes away entirely. If state grounding stays here, factor out:

```python
def _state_grounded_tools(self) -> Iterator[tuple[str, ToolGroundingConfig]]:
    grounding = self._params.grounding
    if grounding is None:
        return
    for tool_id, cfg in grounding.tools.items():
        if cfg.state_path is not None:
            yield tool_id, cfg
```

Either path collapses one of these methods to ~5 lines.

(agent)

docs/dev/tool-envs.md
aniruddh-alt, line 1: **Slop / Stale doc:** ``docs/dev/tool-envs.md`` pitches a three-env design that no longer exists.

After this PR ``StatefulEnvironment`` is gone, ``DeterministicTool.deterministic_outputs`` is gone, and ``env_type: stateful`` no longer resolves. The doc still has:

- Line 25: ``deterministic_outputs:`` on a tool (schema doesn't exist; lookup_table lives on the env).
- Line 90: ``env_type: stateful`` (no longer registered).
- Line 191: "the builder dispatches ... ``synthetic`` / ``stateful``" framing — the framing is now "deterministic vs synthetic, where synthetic optionally has state".

Two options:

1. Delete the doc entirely — the example YAMLs + dataclass docstrings carry their weight, and a doc that requires "keep in sync with three files" tends to rot.
2. Rewrite as a 30-50 line "two envs" doc that matches the post-refactor design.

I'd lean delete unless the doc has external readers; otherwise we ship the wrong mental model.

(agent)

configs/examples/synthesis/ehr_stateful_synth.yaml
aniruddh-alt, line 1: **Scope / debloat:** 638-line example yaml is heavy for one of two stateful examples.

The 15-patient seed data dominates the file (~250 lines of nested patient records). An example synth config typically needs 3-5 seed rows max to demonstrate the contract — the synthesizer's job is to produce variety, not to consume a realistic dataset. Worth trimming initial_state to ~3 representative patients so reviewers can read the example end-to-end in under a minute.

If the rich 15-patient seed is intentionally exercising scale (e.g. the judge configs we deleted assumed it), it's worth a comment in the yaml header saying why.

(agent)

src/oumi/core/configs/params/grounding_params.py
aniruddh-alt: Done in 909cd9aac. Took Option B — `StateGroundingConfig` now lives on `SyntheticStateParams.grounding`, `ToolGroundingConfig` is back to fields-only for deterministic envs.

src/oumi/environments/synthetic_environment.py
aniruddh-alt: Done in 346c339cf. `expect_state` flag removed; stateless check is now inline in `_step_executor_one`, and the helper renamed to `_validate_executor_output`.

src/oumi/environments/synthetic_environment.py
aniruddh-alt: Resolved by 909cd9aac + 346c339cf. After the StateGroundingConfig split the loops walk a flat `self._state_grounding` list directly — no more shared `state_path is None` filter to duplicate. `sample_grounding` also dropped its defensive guards since `_validate_state_grounding` + state_schema validation establish the invariants.

docs/dev/tool-envs.md
aniruddh-alt: Deleted in 909cd9aac.

configs/examples/synthesis/ehr_stateful_synth.yaml
aniruddh-alt: Trimmed in 346c339cf — 15 patients → 3 representative records (P001 rich, P002 empty, P003 multi-condition). File 633 → 499 lines.

jgreer013 (approved):

PR #2458, opened 2026-05-14
[Synth Env] Phase 2 (examples): EHR stateful synthesis example + executors

aniruddh-alt:
## Description

Part of the **Stateful Tool Execution** PR chain — a worked example of the stateful-tool pathway from #2457, referencing executors by registry name per #2555.

**What changed**: Adds an EHR (electronic health records) synthesis config plus the executors it dispatches to, demonstrating `state_params` + `executor` + `state_path` grounding end-to-end.

**Why**: Reviewers and downstream users need a concrete, runnable reference for the stateful-tool API. Keeping the example out of the core PR lets that PR stay focused on env behavior.

### What's in this PR

The whole example lives in one directory, outside `src/oumi/` so it doesn't ship in the package (@wizeng23's review):

- `configs/examples/synthesis/ehr_stateful_agent/ehr_stateful_synth.yaml` — 6-tool EHR workflow over 15 patient records: list / get / record-vitals / diagnose / prescribe / update-allergies. Wires `state_params`, `state_path` grounding, and per-tool `executor` **registry names** into `SyntheticEnvironment`.
- `configs/examples/synthesis/ehr_stateful_agent/executors.py` — the six executors, registered as `ehr.*` via `@register_tool_executor`. Writes return `updated_state`; read-only tools return outputs only. State updates are non-mutating (`{**state, ...}`), so a sample's state is never edited in place.
- `configs/examples/synthesis/ehr_stateful_agent/extra_deps.txt` — points `OUMI_EXTRA_DEPS_FILE` at the executors module so its decorators run at startup.
- `tests/unit/examples/test_ehr_executors.py` — executor behavior coverage (not-found, duplicate diagnosis, allergy conflict, already-prescribed, read-only invariants, input-state immutability) plus a guard that every `executor:` name in the config resolves to the matching registered function.

### Running it

```bash
OUMI_EXTRA_DEPS_FILE=configs/examples/synthesis/ehr_stateful_agent/extra_deps.txt \
  oumi synth -c configs/examples/synthesis/ehr_stateful_agent/ehr_stateful_synth.yaml
```

## PR Chain Navigation

- **Previous PRs**: #2457 (core: stateful tool execution in `SyntheticEnvironment`), #2555 (registry-name executor resolution)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low


configs/examples/synthesis/ehr_stateful_synth.yaml
```diff
@@ -0,0 +1,663 @@
...
+# reference real names.
+#
+# Requirements:
+#   - Set OPENAI_API_KEY in the environment, or change the inference_config.
+#
+# Usage:
+#   oumi synth -c configs/examples/synthesis/ehr_stateful_synth.yaml
+#
+# See also:
+#   - Spec: docs/superpowers/specs/2026-04-30-stateful-tool-executors-design.md
```
wizeng23: Delete?

src/oumi/examples/ehr/executors/__init__.py
wizeng23, line 1: Can we place these functions outside of our core src/oumi/ library, since it gets imported by all our users? For organization, it might make sense to put them in conigs/examples/synthesis next to the yaml file that references them, to package the whole example in one directory.

configs/examples/synthesis/ehr_stateful_synth.yaml
```diff
@@ -0,0 +1,663 @@
...
+            type: object
+            properties:
+              patient_id: { type: string }
+              timestamp: { type: string }
+              bp: { type: string }
+              hr: { type: integer }
+              temp_f: { type: number }
+            required: [patient_id, timestamp, bp, hr, temp_f]
+          read_only: false
+          executor: oumi.examples.ehr.executors.record_vitals
```
wizeng23: Thoughts on using the registry pattern compared to the module path? It's the common pattern in this repo, and it also makes experimentation easier (ex. allows you to define a function in a colab notebook and then use it)

src/oumi/examples/ehr/executors/add_diagnosis.py
```diff
@@ -0,0 +1,56 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+
+from oumi.core.types.tool_call import ToolResult
+from oumi.examples.ehr._state import find_patient, replace_patient
+
+
+def add_diagnosis(arguments: dict[str, Any], state: dict[str, Any]) -> ToolResult:
```
wizeng23: I might be missing context, but why are we deep copying state across stateless functions? This won't scale well for large states, and is harder to reason about an environment when the implementation is scattered across multiple files. Is there a downside to using an OpenEnv or OpenAI Gym-style environment class?

configs/examples/synthesis/ehr_stateful_synth.yaml
aniruddh-alt: Yes will remove this!

configs/examples/synthesis/ehr_stateful_synth.yaml
aniruddh-alt: Got it, I had gone ahead with the module path for executable environments as well, so this might be a bigger refactor. But I think since that is a common pattern it would make more sense, I shall look into this further.

src/oumi/examples/ehr/executors/add_diagnosis.py
aniruddh-alt: A Gym/OpenEnv-style class that mutates internal state would be preferable for large or live systems. Main now supports that through ExecutableEnvironment, where executors receive a context handle such as a database connection. For this example, the state is a small, config-defined JSON document, so introducing a custom stateful environment class would require additional cloning, serialization, grounding, and rollback hooks without much benefit.

wizeng23 (approved):

PR #2479, opened 2026-05-28
feat(gemma4): regex LoRA targeting + E2B/E4B LoRA recipes

aniruddh-alt:
# Gemma 4 LoRA training support

Adds first-class **LoRA fine-tuning for Gemma 4** (E2B / E4B) in Oumi, plus ready-to-run recipes.

Gemma 4 is multimodal (`Gemma4ForConditionalGeneration`): a text transformer alongside vision and audio towers. Those towers use `Gemma4ClippableLinear` modules that PEFT cannot adapt — and they reuse the same projection names (`q_proj`, `v_proj`, …) as the text model, so a naive LoRA target list crashes on them. Two small core changes make LoRA SFT work end-to-end.

## What's in this PR

- **`lora_exclude_modules`** (`PeftParams`) — a new field that scopes LoRA *away* from named modules. Entries are regex patterns, joined into a single regex and passed to PEFT's `exclude_modules` (matched with `re.fullmatch`); exclusion takes precedence over `lora_target_modules` and over `all-linear`. Recipes target the plain projection names and exclude the towers:

  ```yaml
  peft:
    lora_target_modules: ["q_proj", "v_proj", ...]   # bare names, matched by suffix
    lora_exclude_modules: [".*vision_tower.*", ".*audio_tower.*"]
  ```

  The two fields are intentionally asymmetric: `lora_target_modules` takes bare names (PEFT matches them by suffix), but `lora_exclude_modules` must be regexes — PEFT matches a bare-name exclude list by leaf-suffix, so `"vision_tower"` would exclude only that node and not the `q_proj`/`v_proj` nested under it. A regex like `.*vision_tower.*` reaches the whole subtree, which is what keeps LoRA off the towers. The `lora_target_modules` docstring is also clarified to state names match by suffix.
- **`Message.get()`** (`core/types/conversation.py`) — a dict-style `.get(key, default)` on `Message`. Gemma 4's chat template reads message fields via `.get()`; without it, `apply_chat_template` raised `UndefinedError`. Only declared fields are exposed, so a key that collides with a method name (e.g. `"get"`) returns the default rather than a bound method. (Internal: OPE-1861.)
- **Recipes** (`configs/recipes/gemma4/`) — E2B and E4B LoRA SFT recipes (`sft/{e2b,e4b}_lora/`), the E4B full-fine-tune recipe (`sft/e4b_full/`, originally authored by @wizeng23 and cherry-picked with authorship preserved), GCP launch jobs for each, and a family README documenting the LoRA-scoping approach.

**No dependency changes.** Gemma 4 training needs `transformers >= 5.5.4`, which is already within Oumi's existing pins (installed automatically with `oumi`).

## Validation

Trained both LoRA recipes on **banking77** (77-class banking-intent classification; 3,080-example test set) on H100, then measured accuracy. LoRA lifts the base instruct model substantially:

| Model | Base instruct | **+ LoRA (this recipe)** |
|-------|:-------------:|:------------------------:|
| Gemma 4 E2B | 65.2% | **90.5%** |
| Gemma 4 E4B | 69.5% | **94.4%** |

LoRA attaches only to the language model (~12M / ~17M trainable params), loss decreases to convergence, and there are no `Gemma4ClippableLinear` errors — confirming the recipes train a useful adapter end-to-end on a real downstream task. Unit tests cover `to_lora()` exclude-regex joining (`test_peft_params.py`) and `Message.get()` field scoping (`test_conversation.py`). (Eval/inference tooling for Gemma 4 lives downstream; this PR adds the training path.)

## Related issues

Internal: OPE-1861 (chat-template `.get()` compatibility). Part of the Gemma 4 training onboarding effort.

## Before submitting


src/oumi/core/configs/params/peft_params.py
```diff
@@ -320,6 +334,15 @@ def to_lora(self) -> LoraConfig:
...
             target_modules = "all-linear"
+        elif target_modules and any(
+            re.search(r"[.*+?\\^$|()\[\]{}]", m) for m in target_modules
+        ):
+            # Items contain regex metacharacters — join into a single regex
+            # string so PEFT uses re.fullmatch instead of suffix matching.
+            # This is needed for multimodal models (e.g. Gemma4) where
+            # plain names like "q_proj" would match both language model and
+            # vision/audio tower layers, causing errors on unsupported types.
+            target_modules = "|".join(f"({m})" for m in target_modules)
```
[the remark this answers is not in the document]
aniruddh-alt: Good catch — confirmed against PEFT's `check_target_module_exists`: a `str` `target_modules` goes through `re.fullmatch`, while a list suffix-matches via `key.endswith(f".{name}")`. So mixing a regex entry with a bare name did silently strip the bare name's suffix semantics.

Fixed: when the heuristic fires, bare (metacharacter-free) entries are now auto-promoted to `(.*\.)?<name>`, preserving PEFT's suffix-match meaning (e.g. `gate_proj` still matches `…mlp.gate_proj`) instead of only matching the literal string. Regex entries pass through unchanged, and all-regex lists (like the shipped Gemma 4 recipes) compile to the exact same string as before. Also updated the docstring and replaced the misleading test — it asserted `re.fullmatch(…, "gate_proj")` against the bare string, which masked the bug — with one that checks a realistic dotted path plus a negative case.

Validated end-to-end on an L40S: E2B and E4B LoRA both train with the adapter scoped to `language_model` only (12M / 17M trainable params, no `Gemma4ClippableLinear` error).

wizeng23 (approved):

src/oumi/core/configs/params/peft_params.py
```diff
@@ -130,11 +131,28 @@ class PeftParams(BaseParams):
...
+
+    For multimodal models where you need to target specific model components,
+    use regex patterns to scope targeting. For example, to target only the
+    language model layers in Gemma4 (avoiding vision/audio tower layers):
+
+        lora_target_modules:
+          - ".*language_model.*q_proj"
+          - ".*language_model.*v_proj"
+
+    When any item contains a regex metacharacter, all items are joined into a
```
lefft: Could we instead add a new field `lora_target_modules_regex: bool = False` to PeftParams that determines whether the modules are targeted by name or by a regex pattern? The thinking is that if people are going to specify target modules in this way (which is nice to have IMO!), we want to make sure they are doing so deliberately. Without an explicit field controlling this, a user could just be intending to type a module path but unbeknownst to them, be targeting the wrong set of modules since they did not know that `.` is a metacharacter. 

Adding auto-detection like this has the potential to break external users' existing configs, since PEFT does support patterns like `self_attn.q_proj`, so people could already be using `.` when specifying target modules (not sure if this is likely to happen, but best to not introduce breaking changes if it can be avoided). 

This would also allow us to simplify the `elif target_modules and ...:` branch, and eliminate some of the explainer comments. 

@wizeng23  WDYT?

lefft (commented):
Looks good. See comment on `lora_target_modules_regex` (happy to discuss offline). 

Also, would it be worth adding an inference config for gemma4 here? We have faced some issues specifically with gemma4 LoRA, so would be good to provide users with everything they need to get gemma4 LoRA working e2e. (that could be added in a follow up PR!)

src/oumi/core/configs/params/peft_params.py
aniruddh-alt: I think that makes a lot of sense! I do see the potential breakage that this regex matching could cause. Adding the boolean field should be pretty straightforward. Happy to hear @wizeng23's thoughts!

aniruddh-alt:
> Looks good. See comment on `lora_target_modules_regex` (happy to discuss offline).
> 
> Also, would it be worth adding an inference config for gemma4 here? We have faced some issues specifically with gemma4 LoRA, so would be good to provide users with everything they need to get gemma4 LoRA working e2e. (that could be added in a follow up PR!)

Yes I can create a new PR for the inference config!

src/oumi/core/configs/params/peft_params.py
```diff
@@ -313,19 +334,25 @@ def to_lora(self) -> LoraConfig:
...
         target_modules = self.lora_target_modules
         if target_modules == ["all-linear"]:
             target_modules = "all-linear"
 
+        # Join exclude patterns into a single regex: PEFT matches exclude_modules with
+        # re.fullmatch, so a plain list wouldn't reach modules nested under a tower.
+        exclude_modules = self.lora_exclude_modules
+        if exclude_modules:
+            exclude_modules = "|".join(f"({m})" for m in exclude_modules)
+
```
aniruddh-alt, line 352: I have replaced the regex field with "execlude_modules" field that takes in the modules that we want to exclude. I believe this is more ergonomic than regex boolean.

@lefft @wizeng23

lefft: Thanks, this seems like a good approach. Only issue I see is that peft's exclude_modules (see https://github.com/huggingface/peft/pull/2102) accepts a list of bare module names, whereas here we are requiring them to be regex. 

And for us, `lora_target_modules` accepts bare names but `lora_exclude_modules` requires regex.

Not necessarily a problem but worth calling out explicitly.

lefft (approved):

PR #2492, opened 2026-06-04
feat(gemma4): add Gemma 4 26B (MoE) LoRA recipe

aniruddh-alt:
# Add Gemma 4 26B (MoE) LoRA recipe

Adds a LoRA SFT recipe for **Gemma 4 26B** (`google/gemma-4-26B-A4B-it` — Mixture-of-Experts, ~25.8B total / ~4B active, image+text) under `configs/recipes/gemma4/sft/26b_lora/`. Builds on #2479 (now merged), which added the `lora_exclude_modules` support these recipes rely on.

## What's in this PR

- `configs/recipes/gemma4/sft/26b_lora/train.yaml` — FSDP (FULL_SHARD) LoRA SFT on `alpaca-cleaned`; same text-transformer scoping (`lora_exclude_modules`: `.*vision_tower.*`, `.*multi_modal_projector.*`) and `transformer_layer_cls: Gemma4TextDecoderLayer` as the 31B recipe.
- `configs/recipes/gemma4/sft/26b_lora/gcp_job.yaml` — SkyPilot GCP job (A100:8, FSDP via `oumi distributed torchrun`).
- `configs/recipes/gemma4/README.md` — mark 26B as "LoRA config available" + launch example.

## Validation

Validated end-to-end in oumi's OSS environment on H100s with FSDP `FULL_SHARD`. `google/gemma-4-26B-A4B-it` loads and LoRA trains to completion — **9,292,800 trainable params (~0.035%, attention + dense FFN**, see the MoE note) — loss descending cleanly, adapter saved (`rc=0`).

Pointing the recipe's LoRA setup at a task's training split (in place of the shipped `alpaca-cleaned` default) gives a real downstream gain on the MoE, via native HF generation:

| Task | Base | + LoRA |
|---|---|---|
| banking77 (n=200) | 72.5% | **89.5%** |
| pubmedqa (n=100) | 57.0% | **76.0%** |

(Both measured on `transformers 5.7.0`, within oumi's pin. banking77 re-verified end-to-end on oumi `main` + `transformers 5.7.0`: LoRA trains clean and the long-prompt eval runs at batch 16 with no OOM — see eval note.)

**MoE note:** `q/k/v/o` + `gate/up/down_proj` adapt the attention projections **and** the always-on per-layer dense FFN (`Gemma4TextMLP`). Only the **sparse routed experts** (`Gemma4TextExperts`) are out of reach: they store fused 3D `nn.Parameter`s rather than name-addressable `nn.Linear`, so name-based LoRA can't target them and they stay frozen. Confirmed by inspecting the trained adapter — 205 wrapped modules: 115 attention + 90 dense-MLP (`mlp.gate/up/down_proj`, all 30 layers) + 0 experts. Adapting the experts would need a param-level approach — follow-up. The banking77 gain above reflects attention + dense-FFN LoRA.

**Eval note:** transformers has two MoE expert kernels — `batched_mm` copies an expert's weights per token-expert pair (~25.6 GiB for a ~900-token prompt, which OOMs native HF eval on long prompts) and `grouped_mm` avoids the copy. `grouped_mm` is the **default** for this model across oumi's supported transformers range — verified on **5.7.0, 5.9.0, and 5.10.2** — so long-prompt tasks like banking77's 77-label prompt evaluate cleanly at batch 16 with **no OOM and no config change**. Training is unaffected either way.

## Related issues

N/A — config-only addition. Builds on #2479 (merged).

## Before submitting

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Possible contributions include:
 * Documentation and code readability improvements
 * Code review of pull requests
 * Tutorials, blog posts, talks, and social media posts that promote Oumi
-* Community participation in [issues](https://github.com/oumi-ai/oumi/issues), [Discord](https://discord.gg/oumi), and [X](https://x.com/Oumi_PBC) to share knowledge and help each other.
+* Community participation in [issues](https://github.com/oumi-ai/oumi/issues), [Discord](https://discord.gg/oumi), and [X](https://x.com/Oumi_ai) to share knowledge and help each other.
 
 If you want to contribute but you are short of ideas or have any questions, reach out (<contact@oumi.ai>) and we can help.
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Documentation](https://img.shields.io/badge/Documentation-oumi-blue.svg)](https://oumi.ai/docs/en/latest/index.html)
 [![Blog](https://img.shields.io/badge/Blog-oumi-blue.svg)](https://oumi.ai/blog)
-[![Twitter](https://img.shields.io/twitter/follow/Oumi_PBC)](https://x.com/Oumi_PBC)
+[![Twitter](https://img.shields.io/twitter/follow/Oumi_ai)](https://x.com/Oumi_ai)
 [![Discord](https://img.shields.io/discord/1286348126797430814?label=Discord)](https://discord.gg/oumi)
 [![PyPI version](https://badge.fury.io/py/oumi.svg)](https://badge.fury.io/py/oumi)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@
 
 ## 🔥 News
 
+- [2026/06] Added support for the Gemma 4 model family
 - [2026/05] [Oumi v0.8 released](https://github.com/oumi-ai/oumi/releases/tag/v0.8) with `oumi deploy` CLI for dedicated inference endpoints, an `oumi-mcp` MCP server for Claude/Cursor integration, batch API support across Anthropic/Fireworks/Together, and Transformers v5 / TRL / vLLM dependency upgrades
 - [2026/03] Upgraded to Transformers v5, TRL v0.30, vLLM v0.19, and veRL v0.7 compatibility
 - [2026/03] [MCP Integration Phase 1](https://github.com/oumi-ai/oumi/pull/2234): package scaffold and dependencies for MCP server support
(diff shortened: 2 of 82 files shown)


lefft:
**Note: this comment is AI-generated analysis (Claude, via Tim's review session), not a human reviewer.** Flagging because it touches the accuracy of a specific technical claim in this PR — please verify independently before treating it as authoritative.

## Re: "LoRA currently adapts only the attention projections"

I think this claim (in `train.yaml`'s header comment and in the PR description's MoE note) is likely inaccurate for `google/gemma-4-26B-A4B-it`, based on tracing `transformers` v5.5.4's `modular_gemma4.py`:

- `Gemma4TextDecoderLayer.__init__` sets `self.mlp = Gemma4TextMLP(config, layer_idx)` — a **dense** MLP (subclass of `Gemma3MLP`) with ordinary `nn.Linear` submodules named `gate_proj`/`up_proj`/`down_proj`.
- In `forward()`, this dense MLP runs **unconditionally on every layer**: `hidden_states = self.mlp(hidden_states)`. The sparse MoE block (`self.router` + `self.experts`) is only ever **added on top** when `enable_moe_block` is set: `hidden_states = hidden_states_1 + hidden_states_2` (dense output + expert output). It's additive, not a replacement.
- Only `Gemma4TextExperts` (the routed/sparse experts, via its `MixtralExperts` base class) stores weights as fused `nn.Parameter` 3D tensors (`gate_up_proj`, `down_proj`) rather than `nn.Linear` submodules — genuinely unreachable by name-based LoRA targeting, since PEFT only wraps actual `nn.Module` instances. **This part of the recipe's claim holds.**

But the dense `Gemma4TextMLP.gate_proj/up_proj/down_proj` *are* real `nn.Linear` layers, present on all 30 decoder layers of this checkpoint (`enable_moe_block: true` is a single flag in the model's `config.json`, not per-layer). Since `lora_exclude_modules` only excludes `.*vision_tower.*` / `.*multi_modal_projector.*`, these dense-path projections aren't excluded — so the `gate_proj/up_proj/down_proj` entries already in `lora_target_modules` should be matching and adapting them.

**Rough check against the reported 9,292,800 trainable params:** with `hidden_size=2816`, `intermediate_size=2112`, r=8, over 30 layers:
- Attention only (q/k/v/o, accounting for `v_proj` being `None` on the 5 full-attention layers per `attention_k_eq_v`): ≈5.5M
- + dense MLP (gate/up/down): ≈3.5M
- Sum ≈ 9.0M — closely tracks the reported 9.29M, whereas attention-alone doesn't.

If that's right, the recipe is adapting attention **and** the always-on dense FFN path, and only the sparse routed experts are out of reach — a more capable (and different) picture than "attention projections only." Could be worth confirming directly via `peft_model.print_trainable_parameters()` / inspecting `targeted_module_names` on an actual loaded model, and adjusting the comments/README/PR description if confirmed. Doesn't affect whether the recipe trains correctly — just the accuracy of what it says it's doing.

lefft (approved):
Worth checking the analysis in the main comment, other than that LGTM

aniruddh-alt:
> Worth checking the analysis in the main comment, other than that LGTM

updated docs based on comments!

PR #2497, opened 2026-06-10
deps: cap torch below 2.9 (H100 NCCL SIGSEGV)

min-oumi:


min-oumi:
Closing — `vllm>=0.14` hard-requires `torch>=2.9.1`, so capping torch <2.9 is incompatible with oumi's current vllm pin. Pivoting to a non-version-pin workaround for the underlying NCCL segfault (env-var based, in the consumer).

PR #2498, opened 2026-06-10
feat(inference): partial-inference result types + progress file reporter (1/5)

rlehman221:
1. **#2498 — core types + progress reporter ◀ this PR**
2. #2499 — base-engine `infer_partial()` template
3. #2500 — remote-engine per-row failures
4. #2501 — `judge_partial()`

---

Foundation types for `infer_partial()` (lands in part 2/5):

- `FailureDetail` — per-row failure info: `error_message`, `status_code`, `is_retryable`, `error_type`. A retryable failure has already consumed the engine's internal `max_retries`; `is_retryable=True` means a fresh submission could plausibly succeed.
- `InferenceErrorType` — `str` enum of failure categories (`api_status`, `connection`, `runtime`, `config`, `parse_error`, `engine_failure`, `unknown`) stamped onto `FailureDetail.error_type`.
- `InferenceResult` — `successful: list[(idx, Conversation)]` + `failed_indices` + `failures`, with `error_messages`/`has_failures` derived. Field names mirror `BatchResult` so duck-typed consumers work unchanged; `BatchResult` itself is untouched.
- `ProgressFileReporter` — atomic (temp + `os.replace`), throttled JSON progress snapshots (`{total, completed, failed, updated_at}`) for cross-process pollers (e.g. a SkyPilot job wrapper). Write failures are logged and swallowed — a broken progress path never kills inference.
- `InferenceConfig.progress_path` — YAML-serializable config field for the snapshot location.

### DIFF
--- a/src/oumi/core/configs/inference_config.py
+++ b/src/oumi/core/configs/inference_config.py
@@ -40,14 +40,7 @@ class InferenceConfig(BaseConfig):
     """Path to the output file where the generated text will be saved."""
 
     progress_path: str | None = None
-    """Path to a JSON file where inference progress counters are written.
-
-    When set, partial inference (see `BaseInferenceEngine.infer_partial`)
-    periodically writes an atomic JSON snapshot of the form
-    ``{"total": N, "completed": n, "failed": f, "updated_at": "<iso8601 utc>"}``
-    so an external process can poll row-level progress. The run is complete
-    when ``completed + failed == total``.
-    """
+    """Path to a JSON file where inference progress counters are written."""
 
     engine: InferenceEngineType | None = None
     """The inference engine to use for generation.
--- a/src/oumi/core/configs/inference_config.py
+++ b/src/oumi/core/configs/inference_config.py
@@ -39,6 +39,16 @@ class InferenceConfig(BaseConfig):
     output_path: str | None = None
     """Path to the output file where the generated text will be saved."""
 
+    progress_path: str | None = None
+    """Path to a JSON file where inference progress counters are written.
+
+    When set, partial inference (see `BaseInferenceEngine.infer_partial`)
+    periodically writes an atomic JSON snapshot of the form
+    ``{"total": N, "completed": n, "failed": f, "updated_at": "<iso8601 utc>"}``
+    so an external process can poll row-level progress. The run is complete
+    when ``completed + failed == total``.
+    """
+
     engine: InferenceEngineType | None = None
     """The inference engine to use for generation.
 
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -55,6 +55,61 @@ class BatchResult:
         return len(self.failed_indices) > 0
 
 
+@dataclass(frozen=True)
+class FailureDetail:
+    """Details about a single failed inference request.
+
+    A retryable failure has already consumed the engine's internal retries
+    (``RemoteParams.max_retries``); ``is_retryable=True`` means a fresh
+    submission could plausibly succeed (transient cause), not that the engine
+    will retry it.
+    """
+
+    error_message: str
+    """Human-readable description of the failure."""
+
+    status_code: int | None = None
+    """HTTP status code of the final failed attempt, if applicable."""
+
+    is_retryable: bool = True
+    """Whether resubmitting this request could plausibly succeed."""
+
+    error_type: str = "unknown"
+    """Failure category: "api_status", "connection", "runtime", "config",
+    "parse_error", "engine_failure", or "unexpected"."""
+
+
+@dataclass
+class InferenceResult:
+    """Result of partial online inference, separating successes from failures.
+
+    Returned by :meth:`BaseInferenceEngine.infer_partial`. Unlike
+    :meth:`BaseInferenceEngine.infer`, which raises if any conversation fails,
+    this pairs each successful conversation with its original input index and
+    reports per-row failures separately so callers can keep successes and
+    decide which failures to resubmit.
+    """
+
+    successful: list[tuple[int, Conversation]]
+    """List of (original_index, conversation) for successful requests."""
+
+    failed_indices: list[int]
+    """Sorted indices of requests that failed."""
+
+    failures: dict[int, FailureDetail]
+    """Mapping of failed index to structured failure info."""
+
+    @property
+    def error_messages(self) -> dict[int, str]:
+        """Mapping of failed index to error message."""
+        return {i: detail.error_message for i, detail in self.failures.items()}
+
+    @property
+    def has_failures(self) -> bool:
+        """Return True if any requests failed."""
+        return len(self.failed_indices) > 0
+
+
 class BaseInferenceEngine(ABC):
     """Base class for running model inference."""
 
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -105,11 +105,13 @@ class InferenceResult:
     successful: list[tuple[int, Conversation]]
     """List of (original_index, conversation) for successful requests."""
 
-    failed_indices: list[int]
-    """Sorted indices of requests that failed."""
-
     failures: dict[int, FailureDetail]
     """Mapping of failed index to structured failure info."""
+
+    @property
+    def failed_indices(self) -> list[int]:
+        """Sorted indices of requests that failed."""
+        return sorted(self.failures)
 
     @property
     def error_messages(self) -> dict[int, str]:
@@ -119,7 +121,7 @@ class InferenceResult:
     @property
     def has_failures(self) -> bool:
         """Return True if any requests failed."""
-        return len(self.failed_indices) > 0
+        return len(self.failures) > 0
 
 
 class BaseInferenceEngine(ABC):
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -57,7 +57,7 @@ class BatchResult:
 
 
 class InferenceErrorType(str, Enum):
-    """Failure category for a partial-inference request."""
+    """Failure category for an inference request."""
 
     API_STATUS = "api_status"
     """The API returned a non-success HTTP status code."""
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -20,6 +20,7 @@ import time
 import uuid
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
 
 import jsonlines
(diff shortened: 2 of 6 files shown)


src/oumi/core/configs/inference_config.py
```diff
@@ -39,6 +39,9 @@ class InferenceConfig(BaseConfig):
     output_path: str | None = None
     """Path to the output file where the generated text will be saved."""
 
+    progress_path: str | None = None
```
wizeng23, line 42: Could we not have this arg and instead write the json to a fixed file within output_path?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+class InferenceErrorType(str, Enum):
+    """Failure category for a partial-inference request."""
+
+    API_STATUS = "api_status"
+    """The API returned a non-success HTTP status code."""
+
+    CONNECTION = "connection"
+    """A network-level error prevented reaching the API."""
+
+    RUNTIME = "runtime"
```
wizeng23, line 68: What's the difference between this and API_STATUS?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
         return len(self.failed_indices) > 0
 
 
+class InferenceErrorType(str, Enum):
```
wizeng23, line 59: Can you confirm that later PRs utilize all of these values?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+class FailureDetail:
+    """Details about a single failed inference request."""
+
+    error_message: str
+    """Human-readable description of the failure."""
+
+    status_code: int | None = None
+    """HTTP status code of the final failed attempt, if applicable."""
+
+    is_retryable: bool = True
```
wizeng23, line 94: Can we derive whether a failure is retryable purely from the error_type? If so, what about removing this field and making it a helper method `is_retryable(error_type) -> bool`?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
         return len(self.failed_indices) > 0
 
 
+class InferenceErrorType(str, Enum):
+    """Failure category for a partial-inference request."""
```
wizeng23: nit: What does partial-inference mean here? Are we able to drop it from the description

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+    is_retryable: bool = True
+    """Whether resubmitting this request could plausibly succeed."""
+
+    error_type: InferenceErrorType = InferenceErrorType.UNKNOWN
+    """Failure category for this request."""
+
+
+@dataclass
+class InferenceResult:
+    """Result of partial online inference, separating successes from failures."""
```
wizeng23, line 103: ditto, what does partial mean here ,and do we have to mention it?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+
+    successful: list[tuple[int, Conversation]]
+    """List of (original_index, conversation) for successful requests."""
+
+    failed_indices: list[int]
```
wizeng23: We can derive this from failures.keys(), can we drop this?

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+
+    error_type: InferenceErrorType = InferenceErrorType.UNKNOWN
+    """Failure category for this request."""
+
+
+@dataclass
+class InferenceResult:
+    """Result of partial online inference, separating successes from failures."""
+
+    successful: list[tuple[int, Conversation]]
```
wizeng23, line 105: Could we make this dict[int, Conversation] to be consistent with failures? IMO that's slightly nicer for iterating through as well

wizeng23, line 105: nit: `successes` to be in line with `failures`

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -55,6 +56,72 @@ def has_failures(self) -> bool:
...
+
+    is_retryable: bool = True
+    """Whether resubmitting this request could plausibly succeed."""
+
+    error_type: InferenceErrorType = InferenceErrorType.UNKNOWN
+    """Failure category for this request."""
+
+
+@dataclass
+class InferenceResult:
```
wizeng23, line 102: I see now that the shape of the existing types, names, etc. is for backwards compatibiltiy. I still think it would be useful to have a cleaner base structure that we can move towards later. Based on my research, it seems a unified list is the standard out there, and prevents a lot of potential issues that could arise with our current setup.

```
  @dataclass(frozen=True)
  class BatchItem:
      index: int
      result: Conversation | FailureDetail

      @property
      def is_success(self) -> bool:
          return isinstance(self.result, Conversation)


  @dataclass
  class InferenceResult:
      items: list[BatchItem]

      @property
      def successful(self) -> list[tuple[int, Conversation]]:
          return [
              (item.index, item.result)
              for item in self.items
              if isinstance(item.result, Conversation)
          ]

      @property
      def failures(self) -> dict[int, FailureDetail]:
          return {
              item.index: item.result
              for item in self.items
              if isinstance(item.result, FailureDetail)
          }

      @property
      def failed_indices(self) -> list[int]:
          return sorted(
              item.index
              for item in self.items
              if isinstance(item.result, FailureDetail)
          )

      @property
      def error_messages(self) -> dict[int, str]:
          return {
              item.index: item.result.error_message
              for item in self.items
              if isinstance(item.result, FailureDetail)
          }

      @property
      def has_failures(self) -> bool:
          return any(not item.is_success for item in self.items)
```

This should keep backwards compatibility, while having a clean underlying representation. WDYT?

src/oumi/core/inference/progress_reporter.py
```diff
@@ -0,0 +1,106 @@
...
+            "total": self._total,
+            "completed": self._completed,
+            "failed": self._failed,
+            "updated_at": datetime.now(timezone.utc).isoformat(),
+        }
+        try:
+            path = Path(self._path)
+            path.parent.mkdir(parents=True, exist_ok=True)
+            # Write to a temp file in the same directory so the replace is atomic.
+            tmp_path = path.with_name(path.name + ".tmp")
```
wizeng23: nit: thoughts on moving this to init so we don't have to keep rerunning it?

src/oumi/core/configs/inference_config.py
rlehman221: I would highly advise we use its own path since both judge and synth don't even have the ability to set output path right now, the progress gets reported to s3 at specific cadence compared to the output dump, so we actually want them to be seperate locations

src/oumi/core/inference/base_inference_engine.py
rlehman221: Yup! Most of them are used in PR3 and then parse error is used in PR4 + PR5

src/oumi/core/inference/base_inference_engine.py
rlehman221: essentially API_STATUS means the HTTP layer gave us a something we can reason about by status code." RUNTIME means "the failure is below or after that layer with no actionable status code

the runtime could be JSON parse failures, response to conversation failures, etc

src/oumi/core/inference/base_inference_engine.py
rlehman221: Not purely from error_type unfortunately since API_STATUS splits both ways: a 429/5xx is retryable, a 400/401/403/404/422 isn't

src/oumi/core/inference/base_inference_engine.py
rlehman221: agreed

src/oumi/core/inference/base_inference_engine.py
rlehman221: I think this is important to keep since it's not used just for online inference but only when we need the partial results

src/oumi/core/inference/base_inference_engine.py
rlehman221: So every consumer we have either iterates all successes or looks up a failure by index so having it unified already would reduce that to a slower iteration also we would have to redefine say the judge and synth since there success items can't reuse this since say for the judge we have JudgeOutput which is different than a Conversation

src/oumi/core/inference/base_inference_engine.py
rlehman221: I'm a bit confused on why this would nicer for iterating? If we make this consistent we would break the contract we have that we use with batch

src/oumi/core/inference/base_inference_engine.py
wizeng23: Runtime failures will still result in some kind of status code though right? Is it just 500? IMO would be good to clarify the distinction between the two in the comments

src/oumi/core/inference/base_inference_engine.py
wizeng23: What about making it a helper fn that takes both error_type and status_code and returning the result? If it could be derived from the rest of the fields in this class, IMO it would be better to do that

src/oumi/core/inference/base_inference_engine.py
wizeng23: nit: "partial online inference" isn't clear to me here, could we say something more clear like "Result of online inference, allowing partial failures". What's the difference between online inference and whatever this is?

src/oumi/core/inference/base_inference_engine.py
wizeng23: Fair, it's just slightly easier to lookup by index in a dict, but I think the more important thing is that it's consistent with failures. We can keep it as-is for backwards compatibiliity.

src/oumi/core/inference/base_inference_engine.py
wizeng23: Fair, the new underlying representation isn't needed now so no need to try to change it now, if you're just using the old backwards-compatible format. Feel free to disregard!

For the latter issue of JudgeOutput vs Conversation though, isn't that applicable regardless of what format we have?

src/oumi/core/inference/base_inference_engine.py
wizeng23: Can disregard since this is the backwards-compatible format

src/oumi/core/inference/base_inference_engine.py
wizeng23: Can stick with current format for backwards compatibility

src/oumi/core/inference/base_inference_engine.py
rlehman221: "For the latter issue of JudgeOutput vs Conversation though, isn't that applicable regardless of what format we have?" So if we don't abstract away anything going with just using successful and failures and then just let the judge create its own output instead of having it have to go through all rows in say items and understanding which one is a failure vs not

wizeng23 (approved):

PR #2499, opened 2026-06-10
feat(inference): infer_partial() template in BaseInferenceEngine (2/5)

rlehman221:
1. #2498 — core types + progress reporter
2. **#2499 — base-engine `infer_partial()` template ◀ this PR**
3. #2500 — remote-engine per-row failures
4. #2501 — `judge_partial()`

**Base PR:** #2498

---

Adds `BaseInferenceEngine.infer_partial()`:

- Extracts `infer()`'s preamble (validation, generation-params merge, input loading, deterministic conversation ids, dataset hash, scratch dedup) into `_prepare_inference_run()`, shared by both entry points. `infer()` is byte-equivalent after the refactor — the existing test suites pass unmodified (519 passed).
- `infer_partial()` returns an `InferenceResult` instead of raising on row failures: scratch-resumed rows are credited as successes (and counted as completed in the initial progress snapshot), the scratch file is retained on failure so re-invoking with identical input/config resumes, and `output_path` receives successes only. Row-level failures never raise; invalid call shapes (duplicate ids, both input + input_path) still do.
- Default `_infer_online_partial()` hook wraps `_infer_online()` whole-batch: every engine (vllm/native/llama_cpp) gets a working all-or-nothing implementation today, with scratch-checkpointed rows credited as successes when the batch raises. True per-row support for remote engines lands in part 3/5.
- Progress reporting via the `progress_path` param (takes precedence) or `InferenceConfig.progress_path`.

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -32,6 +32,7 @@ from oumi.core.configs import (
     InferenceConfig,
     ModelParams,
 )
+from oumi.core.inference.progress_reporter import ProgressFileReporter
 from oumi.core.types.conversation import Conversation
 from oumi.utils.logging import logger
 from oumi.utils.math_utils import is_power_of_two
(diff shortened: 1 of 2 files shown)


src/oumi/core/inference/base_inference_engine.py
```diff
@@ -305,6 +350,184 @@ def infer(
...
+        """
+        if not input:
+            return []
+        output_path = inference_config.output_path if inference_config else None
+        try:
+            results = self._infer_online(input, inference_config)
+        except Exception as e:
+            completed_by_id = {
+                conversation.conversation_id: conversation
+                for conversation in self._load_from_scratch(output_path)
```
wizeng23, line 475: Is this scratch file shared by both this inference run and the previous one? Do the new results get appended?

wizeng23 (approved):

src/oumi/core/inference/base_inference_engine.py
rlehman221: yes it gets shared assuming the inputs are the same and they will get appended so it can pick up from where it left off incase of a crash (i've just kept the same functionality we already had in our normal infra)

PR #2500, opened 2026-06-10
feat(inference): per-row partial failures in RemoteInferenceEngine (3/5)

rlehman221:
1. #2498 — core types + progress reporter
2. #2499 — base-engine `infer_partial()` template
3. **#2500 — remote-engine per-row failures ◀ this PR**
4. #2501 — `judge_partial()`

**Base PR:** #2499

---

True per-row `infer_partial()` support for remote engines (OpenRouter and every other `RemoteInferenceEngine` subclass inherit it unchanged):

- `_query_api_guarded` wraps each per-conversation task: a row exhausting its retries returns a `FailureDetail` instead of killing the whole `tqdm.gather`. `CancelledError` still propagates. `_infer` / `_query_api` are untouched.
- `_failure_detail_from_exception` maps the engine's raise points to structured failures: `APIStatusError` → status code + retryability via `is_non_retriable_status_code` (including the known transient 400 patterns), missing api_url/api_key `ValueError` → `config`/non-retryable, retries-exhausted `RuntimeError` → `runtime`/retryable.
- Progress ticks fire at true per-task completion time.
- Docs: new "Partial-Failure Online Inference" + progress-file sections in `inference_engines.md`.

Tests (aioresponses): mixed 200+500 batch with scratch/output assertions, 401 fast-fail (1 request, non-retryable), retriable-400 pattern, connection error, all-fail without raising, missing API key → per-row config failures, progress file counters.

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -237,7 +237,7 @@ def _failure_detail_from_exception(e: Exception) -> FailureDetail:
             error_type=InferenceErrorType.CONNECTION,
         )
     if isinstance(e, ValueError):
-        # Missing api_url/api_key: deterministic, resubmission is futile.
+        # Deterministic config/input error (bad credentials/schema); retry is futile.
         return FailureDetail(
             error_message=str(e),
             is_retryable=False,
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -19,12 +19,12 @@ import os
 import tempfile
 import urllib.parse
 import warnings
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


src/oumi/inference/remote_inference_engine.py
```diff
@@ -210,6 +215,48 @@ class FileListResponse:
...
+        )
+    if isinstance(e, (aiohttp.ClientError, asyncio.TimeoutError)):
+        # Defensive: _query_api wraps these in RuntimeError after retries.
+        return FailureDetail(
+            error_message=str(e),
+            is_retryable=True,
+            error_type=InferenceErrorType.CONNECTION,
+        )
+    if isinstance(e, ValueError):
+        # Missing api_url/api_key: deterministic, resubmission is futile.
```
wizeng23: How do we know for sure this is the cause of ValueError?

wizeng23 (approved):

src/oumi/inference/remote_inference_engine.py
rlehman221: Sorry, that comment is misleading let me fix that

PR #2513, opened 2026-06-15
fix(launcher): populate JobStatus.start_at for Slurm jobs

min-oumi:
## Summary

`SlurmClient` populated `JobStatus.submit_time` but never `start_at`. A Slurm `JobStatus` therefore never carried the time the job began running, so consumers couldn't distinguish a job that **failed before ever starting** from one that **ran and then failed**.

## Changes

- Add `%S` (start time) to the `squeue` format string; parse it into `start_at` in `_parse_squeue_line`.
- Parse `StartTime` into `start_at` in `_parse_scontrol_show_job`.
- `start_at` stays `None` for jobs that never started (`N/A`/`Unknown` are nulled by `_parse_slurm_epoch`), so it cleanly signals whether the job ran.
- Tests: assert `start_at` on the running (squeue) and terminal (scontrol) paths; add a pending-job case asserting `start_at is None`.

### DIFF
--- a/src/oumi/launcher/clients/slurm_client.py
+++ b/src/oumi/launcher/clients/slurm_client.py
@@ -119,11 +119,11 @@ def _parse_slurm_epoch(value: str | None) -> float | None:
 
 
 def _parse_squeue_line(line: str, cluster_name: str) -> JobStatus | None:
-    """Parses one row of ``squeue --noheader --format='%i %j %u %T %V %R'``."""
-    parts = line.strip().split(None, 5)
-    if len(parts) < 5:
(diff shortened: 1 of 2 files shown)


min-oumi:
Closing — not needed. Consumers tolerate a missing `start_at` (the ISO-timestamp helper falls back to now() and duration calcs guard against None), so nothing breaks without it. Populating it would also switch on a consumer-side abort-on-ran-then-failed path we don't want yet: we prefer falling through to the next backend on any failure, since transient runtime crashes (e.g. a flaky NCCL SIGSEGV) recover there. May revisit if we need accurate per-job start times for billing.

PR #2531, opened 2026-07-02
feat(synthesis): expose ConversationSynthesizer building blocks for out-of-process multi-turn synthesis

aniruddh-alt:
## Summary

Exposes the inference-free stages of `ConversationSynthesizer`'s multi-turn pipeline as public methods, so a caller can drive the *inference* (planning, opening-turn, and turn-by-turn generation) on its own infrastructure instead of inside `synthesize()`. `synthesize()`'s behavior is unchanged; this is extract-and-expose plus new unit tests.

## Public building blocks

1. **`build_planner_prompts`** — select `target_turns`, prepare per-sample routers, attach grounding facts, and render one planner prompt per sample. Run the returned conversations under `PLANNER_JSON_SCHEMA` guided decoding, then feed the raw plan strings to the next stage.
2. **`build_opening_turn_prompts`** — parse the inferred plans into per-turn instructions and render the opening (turn 1, USER) generation prompt.
3. **`build_seed_conversations`** — assemble the seed (assistant-persona SYSTEM message + opening USER turn) plus the `generation_state` a turn-by-turn driver needs: `target_turns`, `turn_plans`, `user_persona`, `output_system_prompt`.

New dataclasses: `PlannerPrompt`, `OpeningTurnPrompt`, `SeedConversation`.

## Internal refactors (behavior-preserving)

- `_render_planner_prompts` split out of `build_planner_prompts` so `_plan_samples` reuses it without re-running router/grounding setup that `synthesize()` already did.
- `_build_turn_prompt` extracted from `_synthesize_all_samples` so the in-process loop and the opening-turn builder render turns identically.

## Notes

- `target_turns` is drawn randomly; callers inferring plans out-of-process must persist `augmented_sample["target_turns"]`.

## Testing

**E2E composition smoke** — chained the three builders as an out-of-process caller would (stubbed inference between stages, since these methods are inference-free): `build_planner_prompts` → *[planner infers plan]* → `build_opening_turn_prompts` → *[model infers opening]* → `build_seed_conversations`.

Example — given:

```python
sample  = {"customer_type": "frustrated", "issue": "billing"}
persona = {USER: "You are a {customer_type} customer.", ASSISTANT: "You are a support agent."}
# canned planner output: 3 turns (greet → diagnose → resolve); canned opening: "Hi, my latest bill looks wrong."
```

the chain produces a complete seed + generation_state:

```
seed.conversation:
  [system] You are a support agent.
  [user]   Hi, my latest bill looks wrong.
seed.generation_state:
  target_turns:         3
  turn_plans:           ['Greet and state the billing problem.',
                         'Agent diagnoses the overcharge.',
                         'Agent resolves and confirms refund.']
  user_persona:         You are a frustrated customer.
  output_system_prompt: Support conversation about billing.
```

Unit tests cover each builder in isolation (incl. a regression test for `{current_turn}` persona rendering). Full **integration E2E** against a live model is exercised downstream in the service that consumes these builders via the agentic inference workflow — out of scope here since these methods are inference-free.

### DIFF
--- a/src/oumi/core/synthesis/conversation_synthesizer.py
+++ b/src/oumi/core/synthesis/conversation_synthesizer.py
@@ -456,22 +456,28 @@ class ConversationSynthesizer:
         user_persona = multiturn_attribute.role_instruction_messages[Role.USER]
         seeds: list[SeedConversation] = []
         for sample, opening in zip(samples, opening_turns):
+            # Personas may reference {current_turn}; the seed is turn 1, matching
+            # build_opening_turn_prompts. Rendered once and reused for every turn.
+            sample_with_turn = {**sample, "current_turn": 1}
             seed = Conversation(
                 messages=[
-                    self._format_persona(sample, assistant_persona, Role.ASSISTANT),
+                    self._format_persona(
+                        sample_with_turn, assistant_persona, Role.ASSISTANT
+                    ),
                     Message(role=Role.USER, content=opening),
                 ]
             )
-            output_system_prompt = None
-            if multiturn_attribute.output_system_prompt is not None:
-                output_system_prompt = self._formatter.format(
-                    sample, multiturn_attribute.output_system_prompt
-                )
+            output_message = self._format_output_system_message(
+                sample, multiturn_attribute.output_system_prompt
+            )
+            output_system_prompt = (
+                output_message.content if output_message is not None else None
+            )
             generation_state = {
                 "target_turns": sample["target_turns"],
                 "turn_plans": sample.get("parsed_turn_plans", []),
                 "user_persona": self._formatter.format(
-                    sample, user_persona, missing_values_allowed=False
+                    sample_with_turn, user_persona, missing_values_allowed=False
                 ),
                 "output_system_prompt": output_system_prompt,
             }
--- a/src/oumi/core/synthesis/conversation_synthesizer.py
+++ b/src/oumi/core/synthesis/conversation_synthesizer.py
@@ -65,6 +65,32 @@ class PlannerPrompt:
     conversation: Conversation
 
 
+@dataclasses.dataclass
+class OpeningTurnPrompt:
+    """An augmented sample plus its opening-turn generation prompt.
+
+    ``augmented_sample`` carries ``target_turns`` and the parsed per-turn
+    ``parsed_turn_plans``; ``conversation`` generates the opening user turn.
+    """
+
+    augmented_sample: dict
+    conversation: Conversation
+
+
+@dataclasses.dataclass
+class SeedConversation:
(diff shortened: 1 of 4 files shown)


src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -304,6 +328,158 @@ def synthesize(
...
+                output_system_prompt = self._formatter.format(
+                    sample, multiturn_attribute.output_system_prompt
+                )
+            generation_state = {
+                "target_turns": sample["target_turns"],
+                "turn_plans": sample.get("parsed_turn_plans", []),
+                "user_persona": self._formatter.format(
+                    sample, user_persona, missing_values_allowed=False
+                ),
+                "output_system_prompt": output_system_prompt,
```
wizeng23, line 482: On line 1167, in _format_output_system_message, you have a strip() call but here you don't, is that intentional?

src/oumi/core/synthesis/conversation_synthesizer.py
```diff
@@ -304,6 +328,158 @@ def synthesize(
...
+        self._validate_roles(multiturn_attribute)
+        assistant_persona = multiturn_attribute.role_instruction_messages[
+            Role.ASSISTANT
+        ]
+        user_persona = multiturn_attribute.role_instruction_messages[Role.USER]
+        seeds: list[SeedConversation] = []
+        for sample, opening in zip(samples, opening_turns):
+            seed = Conversation(
+                messages=[
+                    self._format_persona(sample, assistant_persona, Role.ASSISTANT),
```
wizeng23: In the existing code, line 866, you pass in "current_turn" for formatting the persona, but here you don't. Do you intend to pass the same persona template into both this function and the rest of the code? If so, wouldn't a template with `current_turn` crash here?

wizeng23 (commented):
Can you comment in the PR description of what E2E testing you've done for the PR?

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: Yes this was a bug, if the current_turn is not passed in but is referenced in the template, it would crash. Since this is being called for building the seed conversation, I added current_turn=1 to the sample.

src/oumi/core/synthesis/conversation_synthesizer.py
aniruddh-alt: that wasn't intentional, just an inconsistency from formatting inline here. Refactored to reuse the existing _format_output_system_message helper, so the .strip() is applied and both paths share one code path now.

wizeng23 (approved):

PR #2532, opened 2026-07-07
docs: add lm_eval MAX_PATH and cp1252 errors to Windows troubleshooting

ppotash:
# Description

While trying Oumi on native Windows (Git Bash, Python 3.13, oumi 0.8), I ran into two failures that aren't currently covered in the "Installing on Windows" section of the troubleshooting guide:

1. **`lm_eval` install failure due to the Windows `MAX_PATH` limit.** The `lm_eval` dependency ships task configuration files whose paths exceed Windows' default 260-character limit, so `pip install oumi` fails partway through with an `OSError` unless long path support is enabled.
2. **`UnicodeDecodeError` during training.** Python on Windows defaults to the system code page (`cp1252`) rather than UTF-8, which causes a `UnicodeDecodeError` when a dependency reads a bundled UTF-8 file during training.

This PR adds both error signatures to the existing list of native-Windows failures so users can recognize and diagnose them more quickly. It's purely additive to the current section and doesn't change the existing WSL-first recommendation.

## Related issues

N/A — no related issue.

## Before submitting
- [x] Did you read the [[contributor guideline](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md)](https://github.com/oumi-ai/oumi/blob/main/CONTRIBUTING.md) Pull Request guidelines?

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/docs/faq/troubleshooting.md
+++ b/docs/faq/troubleshooting.md
@@ -19,10 +19,70 @@ Installing natively on Windows outside of a WSL environment can lead to installa
 ERROR: Could not find a version that satisfies the requirement ... (from versions: none)
 ```
 
+or, when long path support is disabled, an error while installing the `lm_eval` dependency:
+
+```text
+ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory:
+'...\\lm_eval\\tasks\\arabic_leaderboard_complete\\...'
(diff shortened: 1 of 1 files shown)


wizeng23 (approved):
Thank you for your contribution!

PR #2533, opened 2026-07-08
feat(verl): support multi-turn prompts in GRPO data path

shanghongsim:
# Description

Adds **multi-turn prompt** support to the verl GRPO data path.

Today the verl data conversion collapses a conversation into exactly **one user
message (prompt) + one assistant message (ground truth)**. This PR generalizes
the conversion so a conversation can carry history:

- **Prompt** = all messages up to the final assistant message (system +
  user/assistant turns), emitted in verl's native chat-list format
  (`[{"role", "content"}, ...]`).
- **Ground truth** = the final assistant message's text.
- **Single-turn is the two-message special case** — its emitted row is
  unchanged (one-element prompt list, QwenVL `<image>` prepend intact).

This works on verl's side with no rollout/reward changes: verl's `RLHFDataset`
already applies the chat template to `prompt` when it's a list of role/content
dicts. The change is purely in how Oumi builds the Parquet row.

### What changed
- `grpo_utils.py`: new pure function
  `extract_prompt_images_completion_from_conversation` (conversation dict →
  `(prompt_messages, images, completion)`), turn-count-agnostic and testable
  without verl installed.
- `verl_grpo_trainer.py`: `_create_verl_data_entry_from_single_turn_conversation`
  → `_create_verl_data_entry_from_conversation`, now emits the multi-message
  prompt list; the VLM image policy moved into
  `_extract_prompt_images_answer_from_conversation`.
- The existing single-turn TRL helper
  (`extract_prompt_images_completion_from_single_turn_conversation`) is left
  untouched — the TRL GRPO path depends on its string-prompt shape.
- Unit tests for single-turn, multi-turn, error cases, and the image policy.

### Behavior change worth noting
A **single-turn** conversation that included a `system` message previously had
that system message **silently dropped**; it is now included in the prompt.
This is a correctness improvement but is an observable change.

### Open questions / caveats (please weigh in)
1. **`"ability": "math"` is hardcoded** in the emitted row. It does nothing
   functionally but is misleading (not every dataset is math). I left it as-is
   to keep this PR focused, but we should consider **removing it** (or making it
   configurable).
2. **Image handling is fast-and-dirty for multi-turn — effectively ignored.**
   Images are only supported for single-turn conversations (existing QwenVL 2.5
   `<image>` prepend). A multi-turn conversation containing images raises a clear
   `ValueError` rather than attempting per-turn `<image>` placement (which is
   model-specific). Multi-turn is text-only for now.
3. **Ground-truth source.** This PR extends current behavior by extracting the
   GT from the **final assistant turn**. An alternative is to carry GT in row
   **metadata** instead (decoupling it from the conversation, allowing prompts
   that end on a user turn). Worth deciding which we want long-term.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/trainers/verl_grpo_trainer.py
+++ b/src/oumi/core/trainers/verl_grpo_trainer.py
@@ -27,7 +27,7 @@ from omegaconf import DictConfig, OmegaConf
 
 from oumi.core.types.conversation import Conversation
 from oumi.utils.grpo_utils import (
-    extract_prompt_images_completion_from_single_turn_conversation,
+    extract_prompt_images_completion_from_conversation,
 )
 
 try:
@@ -186,7 +186,7 @@ class VerlGrpoTrainer(BaseTrainer):
                 raise ValueError(
                     "Invalid conversation_json in training or validation dataset."
                 ) from e
-            return VerlGrpoTrainer._create_verl_data_entry_from_single_turn_conversation
+            return VerlGrpoTrainer._create_verl_data_entry_from_conversation
         return None
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


shanghongsim:
Handed off to Ani before my medical OOO, addressed in: https://github.com/oumi-ai/oumi/pull/2542

PR #2536, opened 2026-07-10
feat(det-env): allow any JSON value as deterministic lookup output

aniruddh-alt:
## What

Widen the deterministic environment's lookup-table `output` from "must be a dict" to **any JSON value** (string, number, boolean, array, object, or null), and adapt grounding so it still works.

Aligns OSS with the enterprise API ([oumi-ai/api#5314](https://github.com/oumi-ai/api/pull/5314)), whose `LookupEntry.output` is typed `JsonValue`. Many real tools return a scalar or list (e.g. `get_stock_price -> 191.23`, `list_tables -> ["users","orders"]`); forcing a `{value: ...}` wrapper is unnatural. OSS only forced dict because grounding needed named fields to project.

## Changes

- **`core/types/tool_call.py`**: add a `JsonValue` alias; retype `ToolResult.output` to `JsonValue`.
- **`environments/deterministic_environment.py`**: retype `ToolLookupEntry.output` to `JsonValue` (default `None`).
- **Grounding**: `sample_grounding` builds each fact row as a named-field projection. A non-dict output has no field names to project, so it now grounds on its **input fields only**; dict outputs still merge input+output (output wins on collision, unchanged).

No change needed in `conversation_synthesizer._tool_message` — it already `json.dumps`es non-str output, so lists/scalars/None serialize correctly. The synthetic env's `_parse_and_validate` still (correctly) requires dict for LLM-simulated output — separate constraint, left alone.

### Why grounding projects input-only for non-dict outputs

Grounding is a **named-field projection**: `grounding.tools[tool_id].fields` is a whitelist of field *names*, and each `GroundingFact` renders as `name=value` in the planner prompt. Only a dict exposes named fields; a bare scalar/list has no name to match against the whitelist. The tool *input* is always a dict, so it always projects — but a non-dict *output* contributes nothing.

Example — a `get_stock_price` tool grounded on `fields=["ticker", "price"]`:

| lookup entry | projected grounding fact |
| --- | --- |
| `input={"ticker":"AAPL"}, output={"price":191.23}` (dict) | `ticker=AAPL`, `price=191.23` |
| `input={"ticker":"AAPL"}, output=191.23` (scalar) | `ticker=AAPL` only |

The scalar `191.23` isn't rejected — it just can't answer "what is the `price` field?", so it contributes only its input fields.

## Tests

- Non-dict outputs (`float`, `int`, `list`, `str`, `bool`, `None`) round-trip through `step()`.
- Non-dict `ToolResult` output is json-encoded at the TOOL message boundary (`None`→`"null"`, `191.23`→`"191.23"`, list, `bool`).
- Scalar-output grounded tool projects input fields only (no output field, no crash).
- All existing grounding tests (dict-output merge path, output-wins-on-collision) still pass.

`tests/unit/environments/test_deterministic_environment.py`, `test_conversation_synthesizer.py`, and `tests/unit/core/types/` all green (192 passed).

> **CI note:** the red `static-checks` (pyright) check is **pre-existing on `main`** — 3 `reportOptional*` errors in `builders/training.py`, `core/datasets/vision_language_dpo_dataset.py`, and `core/tuners/optuna_tuner.py`, none of which this PR touches. Not introduced by this change.

Suggested reviewer: @wizeng23

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/types/tool_call.py
+++ b/src/oumi/core/types/tool_call.py
@@ -25,7 +25,6 @@ keys at validation time would silently lose information that
 downstream code relies on.
 """
 
-from dataclasses import dataclass
 from enum import Enum
 from typing import Any, Literal
 
@@ -177,8 +176,7 @@ class ToolCall(pydantic.BaseModel):
     """The function the model called."""
 
 
-@dataclass
-class ToolResult:
+class ToolResult(pydantic.BaseModel):
     """Result returned by an environment ``step()``.
 
     Runtime value (not an OpenAI wire-format type) — projected by the
--- a/src/oumi/core/types/tool_call.py
+++ b/src/oumi/core/types/tool_call.py
@@ -36,9 +36,6 @@ JSONSchemaType = Literal[
     "object", "string", "number", "integer", "boolean", "array", "null"
 ]
 
-# Any JSON-serializable value.
-JsonValue = str | int | float | bool | list[Any] | dict[str, Any] | None
-
 
 class JSONSchema(pydantic.BaseModel):
     """A JSON Schema object describing the shape of a value.
@@ -190,5 +187,5 @@ class ToolResult:
     else is json-encoded at the message boundary.
     """
 
-    output: JsonValue
+    output: pydantic.JsonValue
     updated_state: dict[str, Any] | None = None
--- a/src/oumi/core/types/tool_call.py
+++ b/src/oumi/core/types/tool_call.py
@@ -36,6 +36,9 @@ JSONSchemaType = Literal[
     "object", "string", "number", "integer", "boolean", "array", "null"
 ]
 
+# Any JSON-serializable value.
+JsonValue = str | int | float | bool | list[Any] | dict[str, Any] | None
+
 
 class JSONSchema(pydantic.BaseModel):
     """A JSON Schema object describing the shape of a value.
@@ -183,9 +186,9 @@ class ToolResult:
 
     Runtime value (not an OpenAI wire-format type) — projected by the
     synthesizer into ``Message(role=TOOL, content=...)`` before output.
-    ``output`` may be a string or a JSON-serializable dict; the
-    synthesizer json-encodes dicts at the message boundary.
+    ``output`` may be any JSON value — a string is used as-is, everything
+    else is json-encoded at the message boundary.
     """
 
-    output: str | dict[str, Any]
+    output: JsonValue
     updated_state: dict[str, Any] | None = None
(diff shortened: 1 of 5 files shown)


src/oumi/core/types/tool_call.py
```diff
@@ -183,9 +186,9 @@ class ToolResult:
...
     Runtime value (not an OpenAI wire-format type) — projected by the
     synthesizer into ``Message(role=TOOL, content=...)`` before output.
-    ``output`` may be a string or a JSON-serializable dict; the
-    synthesizer json-encodes dicts at the message boundary.
+    ``output`` may be any JSON value — a string is used as-is, everything
+    else is json-encoded at the message boundary.
     """
 
-    output: str | dict[str, Any]
+    output: JsonValue
```
wizeng23, line 188: There is a pydantic JsonValue type, are you able to use that? https://pydantic.dev/docs/validation/latest/api/pydantic/types/#pydantic.types.JsonValue

Is there a reason this class uses dataclass when the rest use pydantic?

src/oumi/environments/deterministic_environment.py
```diff
@@ -138,7 +141,12 @@ def sample_grounding(
                 continue
             whitelist = set(tool_grounding.fields)
             for entry in self._kwargs.lookup_table.get(tool.id, []):
-                row = {**entry.input, **entry.output}
+                # Non-dict outputs (scalars/lists) have no named fields to
+                # project, so they ground on their input fields only; dict
+                # outputs merge both.
+                row = dict(entry.input)
+                if isinstance(entry.output, dict):
+                    row.update(entry.output)
```
wizeng23, line 152: Do we want to warn if there's a key conflict between input and output?

src/oumi/core/types/tool_call.py
aniruddh-alt: got it, shall switch to the pydantic type.

aniruddh-alt: This was initially a dataclass because it was a runtime class and was not involved in round trips. However, i think it is just better to follow repo convention and stick with Pydantic! Thanks for catching this.

src/oumi/environments/deterministic_environment.py
aniruddh-alt: Yes!

src/oumi/core/types/tool_call.py
wizeng23: nit: for consistency with other files, consider doing `from pydantic import BaseModel, JsonValue, ...`

wizeng23 (approved):

src/oumi/core/types/tool_call.py
aniruddh-alt: Done — switched to `from pydantic import BaseModel, ConfigDict, JsonValue`.

PR #2540, opened 2026-07-11
feat(collators): add pad_to_multiple_of to TextCompletionsCollatorWithPadding

lefft:
## What

Add an optional `pad_to_multiple_of` to `TextCompletionsCollatorWithPadding` — pad each batch up to a multiple (e.g. `128`) instead of just the longest row. Set via `collator_kwargs`; default `None` keeps the existing pad-to-longest behavior.

## Why

torch `flex_attention` cannot compile a batch shorter than its 128-token block (`NoValidChoicesError`), so a long-context dataset that also contains very short rows crashes at the first short batch. Padding batches up to a block multiple keeps short rows trainable.

## How

Under right padding the collator drops the `attention_mask` and emits sequential `position_ids`. A mask — all-ones, or the `[1..|0..|1..]` of a mixed-length batch — forces transformers to build per-batch mask closures that defeat `torch.compile` caching and knock flex_attention off its fused fast path (silent fallback to the eager kernel → OOM at long context); dropping it keeps every batch on the compiled path. `train.py` also raises torch dynamo's recompile limit (default 8 → 64, overridable via `OUMI_DYNAMO_RECOMPILE_LIMIT`) when `pad_to_multiple_of` is set, so flex_attention's per-shape bucket compiles don't exhaust the limit.

Loss-neutral for right-padded causal LMs: real tokens form a prefix so they never attend padding, and padding labels are `ignore_index` (no loss/gradient). Left padding keeps the mask.

## Test

Unit tests cover single-row, mixed-length, and `pad_to_multiple_of=None` batches. Validated end-to-end in a gemma-4-E4B 64k FFT job (`attn_implementation=flex_attention`): an 8-step run over a dataset mixing ~60-token and ~62k-token rows completed 8/8 with zero dynamo recompile warnings and a clean save; without the option, short rows crash flex compile at the first step.

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/collators/text_completions_collator_with_padding.py
+++ b/src/oumi/core/collators/text_completions_collator_with_padding.py
@@ -113,10 +113,19 @@ class TextCompletionsCollatorWithPadding:
         # An all-ones mask carries no information, but its presence makes
         # transformers wrap attention in per-batch mask closures that compiled
         # attention backends cannot cache (torch.compile guards on closure
-        # identity). Dropping it keeps the pure-causal fast path.
+        # identity). Replace it with explicit sequential position_ids: models
+        # derive identical positions themselves, attention falls back to the
+        # pure-causal fast path, and TRL's token accounting accepts
+        # position_ids in lieu of a mask (its padding-free convention).
         mask = result.get("attention_mask")
         if mask is not None and bool(mask.all()):
             del result["attention_mask"]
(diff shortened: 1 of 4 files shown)


jgreer013 (approved):

lefft:
Flipping back to draft while I continue tweaking to get 64k training to work with E4B...

PR #2551, opened 2026-07-21
fix(inference): use os.replace for atomic progress-file swap (Python 3.10)

wizeng23:
# Description

`ProgressFileReporter._write_snapshot` (`src/oumi/core/inference/progress_reporter.py`) performed its atomic temp→final swap with `pathlib.Path.replace`. On **Python 3.10**, `Path.replace` does not route through `os.replace` (it uses a bound accessor captured at import), so the tests that simulate a failing swap via `patch("os.replace", side_effect=OSError)` were silently bypassed — the replace succeeded, the file got written, and no warning fired. This failed 2 tests on the `install-test (3.10, 2.8.0)` CI job while passing on 3.11+ (where `Path.replace` does call `os.replace`):

```
FAILED inference/test_progress_reporter.py::test_write_failure_never_raises - AssertionError: assert not True
FAILED inference/test_progress_reporter.py::test_write_failure_warns_once - AssertionError: assert 0 == 1
```

**Fix:** call `os.replace(self._tmp_path, self._path)` directly. This makes the atomic-swap intent explicit and exercises the failure path consistently across Python 3.10–3.14. No behavior change on the success path; tests unchanged.

## Related issues

N/A — pre-existing latent test bug on `main` (test coupled to a CPython 3.11 pathlib implementation detail), surfaced by the `pyproject.toml`-triggered install-test on #2548.

## Before submitting

- [x] Did you add / update tests where needed? (Existing tests already cover this; they now pass on Python 3.10.)

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/inference/progress_reporter.py
+++ b/src/oumi/core/inference/progress_reporter.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import os
 import threading
 import time
 from datetime import datetime, timezone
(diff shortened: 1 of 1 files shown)


wizeng23:
Folded into #2550 so both Python 3.10 install-test fixes land together and the 3.10 matrix cell goes green in a single run. The os.replace fix is included there verbatim.

PR #2557, opened 2026-07-23
feat(models): add text_only opt-in for dual-mode checkpoints (e.g. Qwen3.5)

wizeng23:
# Description

Some checkpoints are dual-mode: transformers exposes both a text-only class (e.g. `Qwen3_5ForCausalLM`) and a multimodal class (e.g. `Qwen3_5ForConditionalGeneration`) for the same weights. Oumi's registry maps one `model_type` to one model class, so it could not represent both modes — a dual-mode checkpoint was silently treated as text-only (falling back to `AutoModelForCausalLM`), which meant vision fine-tuning ran without a processor, VL chat template, or image features and produced no vision learning, with no error. This PR adds an opt-in so the same checkpoint can be loaded either way per run.

- **New config field**: `ModelParams.text_only: bool = False`. Default loads the vision-language model (unchanged behavior); `text_only=True` loads only the language backbone, skipping the vision tower.
- **Dual-mode capability check** (`is_dual_mode_model_type` in `supported_models.py`): computed from the transformers auto-mappings rather than a hardcoded list — a model is dual-mode iff `AutoModelForCausalLM` resolves it to a class distinct from its `AutoModelForImageTextToText` class. This correctly admits `qwen3_5`/`qwen3_5_moe`, and excludes both vision-only models (`qwen3_vl`, no causal class) and models whose causal class equals their VLM class (`gemma3`).
- **Validation**: `text_only=True` on a non-dual-mode model raises a clear `OumiConfigError` at config-finalization time, instead of surfacing later as a raw transformers `Unrecognized configuration class` error.
- **Threaded through both VLM-ness derivations**: `is_image_text_llm` (gates processor/collator/`remove_unused_columns`) and `_get_transformers_model_class` (selects the `Auto*` class) both honor the flag, so the model class and the data pipeline stay consistent.
- **Registry**: `qwen3_5` and `qwen3_5_moe` are registered as vision-language models (processor resolves to `Qwen3VLProcessor`; reuses the existing Qwen3-VL internal config).
- **Evaluation**: the LM Harness backend picks the text harness model (`hf`/`vllm`) over the multimodal one (`hf-multimodal`/`vllm-vlm`) when `text_only` is set.

Scope: the flag takes effect for training, native inference, LM Harness (native/hf) evaluation, and the SGLang processor path. It is a documented no-op for the vLLM engine, which manages its own model loading and currently exposes only the multimodal class for these checkpoints; the field docstring states this.

## Related issues

Towardsa OPE-2158

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/builders/models.py
+++ b/src/oumi/builders/models.py
@@ -365,9 +365,19 @@ def _get_transformers_model_class(config):
 
 
 def is_image_text_llm_using_model_name(
-    model_name: str, trust_remote_code: bool, revision: str | None = None
+    model_name: str,
+    trust_remote_code: bool,
+    revision: str | None = None,
+    *,
+    text_only: bool = False,
 ) -> bool:
-    """Determines whether the model is a basic image+text LLM."""
+    """Determines whether the model is a basic image+text LLM.
+
+    When ``text_only`` is True the model is treated as a language model regardless
+    of its registry entry (used to load a dual-mode checkpoint's text backbone).
+    """
+    if text_only:
+        return False
     model_config = find_internal_model_config_using_model_name(
         model_name, trust_remote_code=trust_remote_code, revision=revision
     )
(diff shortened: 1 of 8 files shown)


lefft (approved):
Awesome, thanks @wizeng23! Is the plan to ultimately set `text_only` downstream for dual-mode models supported in ent?

PR #2560, opened 2026-07-24
feat(verl): OumiVerlTool — verl tool RL over any Oumi executable environment

aniruddh-alt:
## What

`OumiVerlTool` connects verl's tool-RL loop to any Oumi executable environment. During a verl GRPO rollout the policy model makes a tool call, that call runs for real against a live environment, and the resulting trajectory gets scored into a training signal. Nothing in the adapter is tied to a particular tool or domain.

The two systems model tools differently. verl expects one tool backed by one executor. Oumi has one environment that exposes many tools through a `ToolRouter`. `OumiVerlTool` sits in between:

```
Policy model → verl ToolAgentLoop → OumiVerlTool.execute
  → env_provider.get_or_build_router (keyed by request_id)
  → ToolRouter.route_batch → ExecutableEnvironment.step → executor
```

- One `ToolRouter` per rollout, keyed by `request_id` in a module-level registry. The router holds live environment connections, which aren't picklable. It stays off `agent_data` because the trajectory gets serialized and shipped to a separate reward worker. Teardown runs on `weakref.finalize` at the rollout boundary.
- The parent router is built once per env-config path and every rollout is just a `for_sample()` clone of it, carrying its own `create_kwargs` (#2569). So each rollout builds one env per isolating environment — nothing is built and discarded, shared envs are never rebuilt, and the config is never copied.
- A `tool_agent` branch in the data path. Rows tagged `agent_name: tool_agent` skip the usual conversation-to-verl converter: the branch packs the prompt inline, forwards `tools_kwargs`, and reads `ground_truth` from metadata. Any tool-agent dataset flows through it, so this PR carries no `grpo_utils` changes.

## Test plan

`pytest` on the touched units; CI green (4787 passed, 41 skipped).

- `test_env_provider.py`: the same `request_id` shares a router; different `request_id`s get independently-seeded isolated envs; each rollout builds its envs exactly once (counts `build_environment` calls). Runs in CPU CI — it doesn't import verl.
- `test_oumi_verl_tool.py`: the adapter routes a call through the rollout's shared router; an invalid env config is rejected at tool construction. Gated on `importorskip("verl")`, and verl sits in the `gpu` extra — so this skips in CPU CI and only runs where verl is installed (scheduled GPU job, or the cluster). Verified against a stubbed verl locally, and end to end on the cluster.
- `test_verl_grpo_trainer.py::test_tool_agent_row_carries_agent_name_and_tools_kwargs`.

## Validation

Validated end to end on real Spider (Qwen2.5-3B-Instruct, 8×H100, verl 0.7.1, 100 GRPO steps over 6,997 train rows):

- **Training curve**, `sql_execution_match` on the **full 1,034-row dev set**: 0.394 → 0.696 → 0.681 → **0.716** (evals every 25 steps).
- **Independent evaluation of the saved checkpoint**: reloaded `global_step_100` in a fresh process across all 8 ranks and re-scored the full dev set → **0.735**. This is the load-bearing check: the improvement is in the weights, not in the training loop's own reporting.
- **Audited 200 validation generations**: every one produced a real query, and none of the scoring answers were literal value echoes, so the reward is not being gamed. 143 of 144 scoring trajectories used exactly one `run_sql` call.

Metrics are in the internal W&B project `oumi-nl2sql-verl` (runs `6zk6dwhp` for training, `hk65jihf` for the checkpoint evaluation and the generation table).

Caveats, stated up front:

- This is our execution-match reward, **not** the official Spider `evaluation.py` metric, so it is not comparable to published Spider numbers. It compares result sets on a single database instance, and 39.5% of dev golds return a single scalar, so a semantically wrong query can still score. Test-suite accuracy is the follow-up.
- **Single seed**, no confidence intervals. The last three evals (0.696 / 0.681 / 0.716) sit within the late-phase step-to-step reward noise (σ ≈ 0.05), so read this as a plateau around 0.68–0.73 rather than a converged figure.
- `actor/entropy` collapses 0.408 → 0.040 (`entropy_coeff: 0`, `use_kl_loss: false`), which is consistent with that plateau — more steps will not help without a KL or entropy term.
- No no-tool ablation, so this does **not** establish that the tool caused the gain.

It demonstrates that the integration trains end to end and the reward improves — not a benchmark result.

Tested stack (required): verl **0.7.1** (`<0.8`), vllm 0.14.0, ray 2.56.1, torch 2.9.1+cu128, **trl 0.19.1** — trl must be `<0.20` because verl 0.7.1 imports `trl.AutoModelForCausalLMWithValueHead`. Multi-GPU additionally requires `NCCL_NVLS_ENABLE=0` on hosts where NVLS multicast cannot bind.

## Related — split of #2544

#2544 was split into three reviewable PRs (merge order):
1. #2569 — `for_sample(env_kwargs_by_env)` in `ToolRouter`. Split out of this PR at review request; base main. **This PR bases on it.**
2. **#2560 (this)** — core adapter.
3. #2561 — NL2SQL `run_sql` executor + `sql_execution_match` reward. Independent, base main.
4. #2562 — Spider demo configs + row builder. Merge last (its configs reference #2560 and #2561).

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -17,8 +17,8 @@
 from __future__ import annotations
 
 import json
-from collections.abc import Callable
-from dataclasses import dataclass
+from collections.abc import Callable, Mapping
+from dataclasses import dataclass, replace
 from typing import Any
 
 from oumi.builders.environments import build_environment
@@ -95,7 +95,9 @@ class ToolRouter:
             on_env_built=on_env_built,
         )
 
-    def for_sample(self) -> ToolRouter:
+    def for_sample(
+        self, env_kwargs_by_env: Mapping[str, Mapping[str, Any]] | None = None
+    ) -> ToolRouter:
         """Return a router safe to use for one sample.
 
         Envs whose ``requires_isolation()`` returns ``True`` are rebuilt via
@@ -104,13 +106,33 @@ class ToolRouter:
         don't require isolation (e.g. ``DeterministicEnvironment`` and
         stateless ``SyntheticEnvironment``) are shared with the parent
         router to avoid the per-sample build + inference-engine attach cost.
+
+        ``env_kwargs_by_env`` replaces ``env_kwargs`` on the rebuilt envs with
+        per-sample data (e.g. the database a single rollout should query). Shared
+        envs are rejected because the override would silently apply to every sample.
         """
+        overrides = env_kwargs_by_env or {}
+        unknown_env_ids = overrides.keys() - self.env_by_id.keys()
+        if unknown_env_ids:
+            raise ValueError(
+                f"Per-sample env_kwargs target unknown environments: "
+                f"{sorted(unknown_env_ids)}. Known: {sorted(self.env_by_id)}"
+            )
+
         env_by_id_new: dict[str, BaseEnvironment] = {}
         for env_id, parent_env in self.env_by_id.items():
             if not parent_env.requires_isolation():
+                if env_id in overrides:
+                    raise ValueError(
+                        f"Environment {env_id!r} is shared across samples, so "
+                        "per-sample env_kwargs cannot be applied to it."
+                    )
                 env_by_id_new[env_id] = parent_env
                 continue
-            fresh = build_environment(self.env_params_by_id[env_id])
+            env_params = self.env_params_by_id[env_id]
+            if env_id in overrides:
+                env_params = replace(env_params, env_kwargs=dict(overrides[env_id]))
+            fresh = build_environment(env_params)
             if self.on_env_built is not None:
                 self.on_env_built(fresh)
             env_by_id_new[env_id] = fresh
@@ -128,20 +150,16 @@ class ToolRouter:
             on_env_built=self.on_env_built,
         )
 
-    def close(self, *, include_shared: bool = False) -> None:
+    def close(self) -> None:
         """Release per-sample envs this router owns.
 
         Only isolation-requiring envs are rebuilt (and thus owned) per router;
         shared envs belong to the parent router and are left open. Each close is
         guarded so one env's failure can't leak the rest; the first error re-raises.
-
-        Pass ``include_shared=True`` when the caller owns the whole router graph
-        (one parent per clone, as in a verl rollout) so parent-shared envs are
-        released too rather than left to garbage collection.
         """
         first_error: BaseException | None = None
         for env in self.env_by_id.values():
-            if not include_shared and not env.requires_isolation():
+            if not env.requires_isolation():
                 continue
             try:
                 env.close()
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -111,6 +111,8 @@ class ToolRouter:
         per-sample data (e.g. the database a single rollout should query). Shared
         envs are rejected because the override would silently apply to every sample.
         """
+        # Both checks run before the first build, so a rejected override can't
+        # strand envs already rebuilt in the loop (their teardown needs a router).
         overrides = env_kwargs_by_env or {}
         unknown_env_ids = overrides.keys() - self.env_by_id.keys()
         if unknown_env_ids:
@@ -118,15 +120,20 @@ class ToolRouter:
                 f"Per-sample env_kwargs target unknown environments: "
                 f"{sorted(unknown_env_ids)}. Known: {sorted(self.env_by_id)}"
             )
+        shared_env_ids = [
+            env_id
+            for env_id in overrides
+            if not self.env_by_id[env_id].requires_isolation()
+        ]
+        if shared_env_ids:
+            raise ValueError(
+                f"Environments {sorted(shared_env_ids)} are shared across samples, "
+                "so per-sample env_kwargs cannot be applied to them."
+            )
 
         env_by_id_new: dict[str, BaseEnvironment] = {}
         for env_id, parent_env in self.env_by_id.items():
             if not parent_env.requires_isolation():
-                if env_id in overrides:
-                    raise ValueError(
-                        f"Environment {env_id!r} is shared across samples, so "
-                        "per-sample env_kwargs cannot be applied to it."
-                    )
                 env_by_id_new[env_id] = parent_env
                 continue
             env_params = self.env_params_by_id[env_id]
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Callable, Mapping
+from contextlib import suppress
 from dataclasses import dataclass, replace
 from typing import Any
 
@@ -132,17 +133,27 @@ class ToolRouter:
             )
 
         env_by_id_new: dict[str, BaseEnvironment] = {}
-        for env_id, parent_env in self.env_by_id.items():
-            if not parent_env.requires_isolation():
-                env_by_id_new[env_id] = parent_env
-                continue
-            env_params = self.env_params_by_id[env_id]
-            if env_id in overrides:
-                env_params = replace(env_params, env_kwargs=dict(overrides[env_id]))
-            fresh = build_environment(env_params)
-            if self.on_env_built is not None:
-                self.on_env_built(fresh)
-            env_by_id_new[env_id] = fresh
+        built_here: list[BaseEnvironment] = []
+        try:
+            for env_id, parent_env in self.env_by_id.items():
+                if not parent_env.requires_isolation():
+                    env_by_id_new[env_id] = parent_env
+                    continue
+                env_params = self.env_params_by_id[env_id]
+                if env_id in overrides:
+                    env_params = replace(env_params, env_kwargs=dict(overrides[env_id]))
+                fresh = build_environment(env_params)
+                built_here.append(fresh)
+                if self.on_env_built is not None:
+                    self.on_env_built(fresh)
+                env_by_id_new[env_id] = fresh
+        except BaseException:
+            # No router escapes to close these, so release them here. Shared envs
+            # aren't in `built_here` — they belong to the parent.
+            for env in built_here:
+                with suppress(Exception):
+                    env.close()
+            raise
 
         return ToolRouter(
             tool_specs=self.tool_specs,
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -128,16 +128,20 @@ class ToolRouter:
             on_env_built=self.on_env_built,
         )
 
-    def close(self) -> None:
+    def close(self, *, include_shared: bool = False) -> None:
         """Release per-sample envs this router owns.
 
         Only isolation-requiring envs are rebuilt (and thus owned) per router;
         shared envs belong to the parent router and are left open. Each close is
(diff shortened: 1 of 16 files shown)


src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,96 @@
...
+_ROUTERS: dict[str, ToolRouter] = {}
+
+
+class _ToolKwargs(TypedDict, total=False):
+    create_kwargs: Mapping[str, Any]
+
+
+class _RolloutData(Protocol):
+    @property
+    def request_id(self) -> str: ...
```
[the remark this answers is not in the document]
aniruddh-alt: False positive: this is the `...` body of a `typing.Protocol` property stub (`_RolloutData.request_id`), which is the standard way to declare a structural-typing member. There is no runtime statement to have an effect.

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def tools_kwargs(self) -> Mapping[str, _ToolKwargs | None] | None: ...
```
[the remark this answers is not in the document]
aniruddh-alt: Same as above — `...` body of the `_RolloutData.tools_kwargs` Protocol property stub, not a stray expression.

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,96 @@
...
+        if env_id in env_kwargs_by_env and env_kwargs_by_env[env_id] != create_kwargs:
+            raise ValueError(
+                f"Tools provide conflicting create_kwargs for environment {env_id!r}."
+            )
+        env_kwargs_by_env[env_id] = create_kwargs
+    for env in cfg.environments:
+        if env.id in env_kwargs_by_env:
+            env.env_kwargs = env_kwargs_by_env[env.id]
+    return ToolRouter.from_environment_config(cfg).for_sample()
+
```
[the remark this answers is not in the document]
aniruddh-alt: fixed in 277ee14 — though not the way suggested. Keeping and closing the parent would not have helped: `close()` skips non-isolating envs on the parent too ("shared envs belong to the parent router and are left open"), so those envs are unreachable from either `close()`.

Took the second option instead — `ToolRouter.close()` now accepts a keyword-only `include_shared` flag, and `_teardown` passes `include_shared=True` because each rollout builds its own parent and therefore owns the whole graph.

Worth noting this is slightly worse than described: `BaseEnvironment.requires_isolation()` **defaults to `False`**, so it is not only a hypothetical second env — any stateless `SyntheticEnvironment` (or a future env holding a remote session) takes that path.

Tests: `test_close_include_shared_releases_parent_shared_envs` (plain `close()` leaves a non-isolating env open, the flag closes it) and `test_router_closed_and_evicted_when_agent_data_is_collected` now asserts the flag is actually passed. The synthesizer is unaffected — keyword-only, default `False`, so `conversation_synthesizer`'s `router.close()` keeps its current behavior.

src/oumi/core/trainers/verl_agentic/oumi_verl_tool.py
```diff
@@ -0,0 +1,78 @@
...
+        instance_id: str,
+        parameters: dict[str, Any],
+        agent_data: Any = None,
+        **kwargs: Any,
+    ) -> tuple[ToolResponse, float, dict]:
+        """Routes one tool call through this rollout's shared Oumi router."""
+        router = get_or_build_router(agent_data, self._env_config)
+        args = {k: v for k, v in (parameters or {}).items() if v is not None}
+        result = router.route_batch([(self._tool_id, args)])[0]
+        out = result.output
```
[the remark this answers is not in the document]
aniruddh-alt: Agreed and fixed in 9bea609. `execute` now wraps `route_batch` and returns the error text as the tool response (`Tool error: <msg>`, reward `0.0`) plus a `logger.warning` so a systematic failure is still visible rather than silently degrading training.

Catching broadly is deliberate here: `route_batch` raises `ToolLookupError`/`ToolArgumentError` and also propagates whatever the executor raises, and for tool RL any of those should become an observation.

Note the NL2SQL executor already caught `sqlite3.Error` itself, so malformed *SQL* was surfaced before this change; the gap was malformed *tool calls*. Verified both paths against a stubbed verl: bad SQL returns `Tool error: near "NOT": syntax error` and an unknown tool id returns `Tool error: Unknown tool 'nope'...`, neither raising. Covered by `test_failed_tool_call_becomes_an_observation`.

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -245,10 +247,85 @@ def _extract_prompt_images_answer_from_conversation(
...
+                )
+            tools_kwargs = metadata.get("tools_kwargs")
+            if not isinstance(tools_kwargs, dict) or any(
+                not isinstance(tool_name, str) or not isinstance(tool_kwargs, dict)
+                for tool_name, tool_kwargs in tools_kwargs.items()
+            ):
+                raise ValueError(
+                    "Tool-agent conversation metadata 'tools_kwargs' must be a "
+                    "mapping of tool names to dictionaries."
+                )
```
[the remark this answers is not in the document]
aniruddh-alt: Split in 277ee14. One note on the premise: the original was not correct "only by accident" — `not isinstance(None, dict)` is `True`, so a `None` short-circuits and raises deterministically.

Readability was a fair point though, so it is now an `isinstance` check followed by a loop, with the shared message in a local so it is not duplicated, and each branch appending what it actually saw (`Got NoneType.` / `Got 3: list.`).

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -245,10 +247,85 @@ def _extract_prompt_images_answer_from_conversation(
...
+                "need_tools_kwargs": True,
+                "tools_kwargs": tools_kwargs,
+            },
+        }
+
     @staticmethod
     def _create_verl_data_entry_from_conversation(
         example: dict, idx: int, data_source: str, split: str
     ) -> dict:
+        conversation = Conversation.from_json(example["conversation_json"])
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed in 277ee14 using the second suggestion — branch on metadata before parsing. It now reads `metadata` off the raw JSON with `json.loads` and only builds a `Conversation` inside the tool-agent branch, so the non-tool-agent path does one pydantic validation instead of two. Threading the parsed object through was the alternative, but that would have changed `grpo_utils.extract_prompt_images_completion_from_conversation`'s signature and its 11 test call sites.

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,96 @@
...
+    if router is not None:
+        router.close()
+
+
+def get_or_build_router(
+    agent_data: _RolloutData, base_env_config: EnvironmentConfig
+) -> ToolRouter:
+    """Return the rollout-scoped router."""
+    request_id = agent_data.request_id
+    router = _ROUTERS.get(request_id)
```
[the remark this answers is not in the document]
aniruddh-alt: Documented in 9bea609 — took option (a). `get_or_build_router`'s docstring now states that `request_id` must be unique per live rollout, because teardown is tied to that `agent_data` and a reused id would close the router still held by the other owner.

Skipped refcounting deliberately: verl mints `request_id` per rollout in `ToolAgentLoop.run()`, so the invariant holds upstream, and refcounting would add shared mutable state to the hot path to defend against a condition that would itself be a verl bug.

src/oumi/core/synthesis/tool_router.py
```diff
@@ -128,16 +128,20 @@ def for_sample(self) -> ToolRouter:
             on_env_built=self.on_env_built,
         )
 
-    def close(self) -> None:
+    def close(self, *, include_shared: bool = False) -> None:
```
shanghongsim: What is the motivation of adding this field?

src/oumi/core/trainers/verl_agentic/oumi_verl_tool.py
```diff
@@ -0,0 +1,87 @@
...
+
+from oumi.core.configs.environment_config import EnvironmentConfig
+from oumi.core.trainers.verl_agentic.env_provider import get_or_build_router
+from oumi.utils.logging import logger
+from oumi.utils.packaging import is_verl_v0_7_or_later
+
+
+@cache
+def _load_env_config(path: str) -> EnvironmentConfig:
+    return EnvironmentConfig.from_yaml(path)
```
shanghongsim: Should we call `finalize_and_validate()` to check that the config is well formed?

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,101 @@
...
+        env_id = tool_env[tool_id]
+        if env_id in env_kwargs_by_env and env_kwargs_by_env[env_id] != create_kwargs:
+            raise ValueError(
+                f"Tools provide conflicting create_kwargs for environment {env_id!r}."
+            )
+        env_kwargs_by_env[env_id] = create_kwargs
+    for env in cfg.environments:
+        if env.id in env_kwargs_by_env:
+            env.env_kwargs = env_kwargs_by_env[env.id]
+    return ToolRouter.from_environment_config(cfg).for_sample()
```
shanghongsim: I read the code for this. From my understanding, if `env.requires_isolation()`, the env is closed and rebuilt in `for_sample()`. This seems inefficient for cases where have a large number of trajectories and corresponding isolated routers. Any way to prevent this build and teardown?

shanghongsim: I guess for spider, since its connecting with db_path. Its just two connects and should not be too expensive. But if there is a dataset that uses `schema_sql`, every rollout runs the full DDL + seed script twice.

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,101 @@
...
+        if entry and (create_kwargs := entry.get("create_kwargs")):
+            create_kwargs_by_tool[tool_id] = dict(create_kwargs)
+    return create_kwargs_by_tool
+
+
+def _build_router(
+    base_env_config: EnvironmentConfig,
+    create_kwargs_by_tool: dict[str, dict[str, Any]],
+) -> ToolRouter:
+    cfg = copy.deepcopy(base_env_config)
```
shanghongsim: The config may be huge (esp for lookup_table), since tool call definitions can be many and long. Is there any way to make more efficient?

src/oumi/core/synthesis/tool_router.py
shanghongsim: Can you place this in a separate PR?

src/oumi/core/trainers/verl_agentic/env_provider.py
aniruddh-alt: I agree! this is pretty wasteful. It would be very ineffecient to build and close, and this calls for lazy loading of the environment. I shall fix this!

src/oumi/core/trainers/verl_agentic/oumi_verl_tool.py
aniruddh-alt: Yes! that would be helpful here!

shanghongsim (approved):
LGTM!! 🚢

PR #2562, opened 2026-07-24
feat(examples): NL2SQL Spider tool-agent GRPO demo (configs + row builder)

aniruddh-alt:
## What

The runnable Spider demo that wires the adapter and the NL2SQL env/reward together: the example configs and the script that builds tool-agent training rows.

- `scripts/datasets/build_spider_tool_agent.py` turns a Spider release into train/val JSONL in Oumi Conversation format. Each row carries the schema-primed system prompt, the question, and metadata (`agent_name: tool_agent`, `ground_truth`, and per-tool `create_kwargs` with the DB path).
- `configs/examples/grpo_verl_nl2sql/`: the GRPO `train.yaml`, the Oumi `environment_config.yaml` (the `run_sql` executor plus `read_only`), and the verl `verl_tool_config.yaml` (model-facing schema plus dispatch to `OumiVerlTool`).

Merge last: the configs reference `OumiVerlTool` (adapter PR) and the `run_sql` executor plus `sql_execution_match` reward (env/reward PR). The diff itself is self-contained and its parse test passes on `main` alone.

## Test plan

`pytest tests/unit/scripts/test_build_spider_tool_agent.py tests/unit/core/configs/test_parse_configs.py` — 1033 pass on `main`. `test_parse_configs` excludes the two non-`TrainingConfig` yamls (env config and verl tool config) and validates the rest.

## Reviewer notes (carried from #2544)

- **Colocated vs server:** colocated (hybrid engine). `mode: async` selects the rollout execution path (verl's in-process agent-loop `AsyncLLM` server, required for multi-turn tool calling), not GPU placement. Placement is still the default hybrid engine, so actor and vLLM share GPUs and `gpu_memory_utilization` carves out vLLM's slice.
- **`run_sql` declared twice** (verl schema here, Oumi executor in `environment_config.yaml`): the two can drift. Since `ToolParams` already carries name/description/parameters, this manifest is derivable from the env config. Tracked as a follow-up, eventually generated at trainer startup so this file disappears.

## Related — split of #2544

#2544 was split into three reviewable PRs (merge order):
1. #2560 — core `OumiVerlTool` adapter. Independent, base main.
2. #2561 — NL2SQL env + reward. Independent, base main.
3. **#2562 (this)** — Spider demo. Merge last (its configs reference #2560 and #2561).

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/configs/examples/grpo_verl_nl2sql/README.md
+++ b/configs/examples/grpo_verl_nl2sql/README.md
@@ -1,0 +1,17 @@
+# NL2SQL tool-agent GRPO
+
+This example trains a verl tool agent on Spider execution accuracy. Prepare the
+Spider JSON files and SQLite databases before launching training:
+
+```bash
+python scripts/datasets/build_spider_tool_agent.py \
+  --spider-root /path/to/spider \
+  --db-root /path/to/spider/database
+
+oumi train -c configs/examples/grpo_verl_nl2sql/train.yaml
+```
+
+The Spider root must contain `train_spider.json` and `dev.json`. The database
+root must contain `<db_id>/<db_id>.sqlite`. The builder writes the configured
+`data/grpo_verl_nl2sql/train.jsonl` and `val.jsonl` files without copying the
+database contents into each row.
--- a/configs/examples/grpo_verl_nl2sql/README.md
+++ b/configs/examples/grpo_verl_nl2sql/README.md
@@ -1,7 +1,20 @@
 # NL2SQL tool-agent GRPO
 
-This example trains a verl tool agent on Spider execution accuracy. Prepare the
-Spider JSON files and SQLite databases before launching training:
+This example trains a verl tool agent on [Spider](https://github.com/taoyds/spider),
+the cross-domain text-to-SQL benchmark of
+[Yu et al. (2018)](https://arxiv.org/abs/1809.08887). Spider scores predictions
+with exact-set match and with execution accuracy; this example optimizes
+execution accuracy — run the predicted and the gold SQL against the same
+database and compare the result sets (see the *Evaluation Metrics* discussion in
+the paper and the reference implementation in
+[`evaluation.py`](https://github.com/taoyds/spider/blob/master/evaluation.py)).
+The `sql_execution_match` reward implements that comparison, treating row order
+as significant only when the gold query has a top-level `ORDER BY`.
+
+Download the JSON files and SQLite databases from the
+[Spider site](https://yale-lily.github.io/spider) or the
+[`xlangai/spider`](https://huggingface.co/datasets/xlangai/spider) mirror, then
+prepare the rows and launch training:
 
 ```bash
 python scripts/datasets/build_spider_tool_agent.py \
@@ -11,7 +24,43 @@ python scripts/datasets/build_spider_tool_agent.py \
 oumi train -c configs/examples/grpo_verl_nl2sql/train.yaml
 ```
 
-The Spider root must contain `train_spider.json` and `dev.json`. The database
-root must contain `<db_id>/<db_id>.sqlite`. The builder writes the configured
-`data/grpo_verl_nl2sql/train.jsonl` and `val.jsonl` files without copying the
-database contents into each row.
+The Spider root must contain `train_spider.json` (7,000 examples) and `dev.json`
+(1,034). The database root must contain `<db_id>/<db_id>.sqlite`.
+
+## Prepared rows
+
+The builder writes `data/grpo_verl_nl2sql/train.jsonl` and `val.jsonl` in Oumi
+`Conversation` format — one row per question, referencing the database by path
+instead of copying its contents into the row:
+
+- `messages[0]` — system prompt carrying the DDL of every table in that row's DB.
+- `messages[1]` — the natural-language question.
+- `metadata.agent_name` — the verl agent loop to run (`tool_agent`).
+- `metadata.ground_truth` — the gold SQL, scored by `sql_execution_match`.
+- `metadata.tools_kwargs.run_sql.create_kwargs.db_path` — absolute path to the
+  SQLite file the `run_sql` tool opens for this rollout.
+
+An example row, with the schema abridged to two tables:
+
+````json
+{
+  "messages": [
+    {
+      "role": "system",
+      "content": "You are a SQL assistant for a SQLite database with this schema:\nCREATE TABLE singer (\nSinger_ID int,\nName text,\nCountry text,\nSong_Name text,\nSong_release_year text,\nAge int,\nIs_male bool,\nPRIMARY KEY (Singer_ID)\n)\n\nCREATE TABLE stadium (\nStadium_ID int,\nLocation text,\nName text,\nCapacity int,\nHighest int,\nLowest int,\nAverage int,\nPRIMARY KEY (Stadium_ID)\n)\nYou may call the tool run_sql(query) to execute a SQL query and inspect results. When you are done, give your final answer as a single SQL query inside a ```sql code block."
+    },
+    { "role": "user", "content": "How many singers do we have?" }
+  ],
+  "metadata": {
+    "agent_name": "tool_agent",
+    "ground_truth": "SELECT count(*) FROM singer",
+    "tools_kwargs": {
+      "run_sql": {
+        "create_kwargs": {
+          "db_path": "/path/to/spider/database/concert_singer/concert_singer.sqlite"
+        }
+      }
+    }
+  }
+}
+````
--- a/configs/examples/grpo_verl_nl2sql/README.md
+++ b/configs/examples/grpo_verl_nl2sql/README.md
@@ -27,11 +27,18 @@ oumi train -c configs/examples/grpo_verl_nl2sql/train.yaml
 The Spider root must contain `train_spider.json` (7,000 examples) and `dev.json`
 (1,034). The database root must contain `<db_id>/<db_id>.sqlite`.
 
+A few Spider rows carry gold SQL that does not run against its own database. The
+builder drops them from both splits — an unscoreable gold aborts the reward
+mid-run — logging each one and a per-split kept/dropped count. Both splits
+therefore come out smaller than the counts above, so validation accuracy here is
+not directly comparable to a published Spider dev number without accounting for
+the dropped rows.
+
 ## Prepared rows
 
 The builder writes `data/grpo_verl_nl2sql/train.jsonl` and `val.jsonl` in Oumi
-`Conversation` format — one row per question, referencing the database by path
-instead of copying its contents into the row:
+`Conversation` format — one row per surviving question, referencing the database
+by path instead of copying its contents into the row:
 
 - `messages[0]` — system prompt carrying the DDL of every table in that row's DB.
 - `messages[1]` — the natural-language question.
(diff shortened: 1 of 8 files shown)


configs/examples/grpo_verl_nl2sql/README.md
```diff
@@ -0,0 +1,17 @@
+# NL2SQL tool-agent GRPO
+
+This example trains a verl tool agent on Spider execution accuracy. Prepare the
```
shanghongsim: could we link the spider repo and paper? specifically portion on the metric

configs/examples/grpo_verl_nl2sql/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+```bash
+python scripts/datasets/build_spider_tool_agent.py \
+  --spider-root /path/to/spider \
+  --db-root /path/to/spider/database
+
+oumi train -c configs/examples/grpo_verl_nl2sql/train.yaml
+```
+
+The Spider root must contain `train_spider.json` and `dev.json`. The database
```
shanghongsim: also add a link to where to get these files

configs/examples/grpo_verl_nl2sql/train.yaml
```diff
@@ -0,0 +1,54 @@
...
+  eval_strategy: "steps"
+  eval_steps: 20
+  learning_rate: 1.0e-6
+  reward_functions: ["sql_execution_match"]
+  grpo:
+    max_completion_length: 1024
+    use_vllm: True
+    temperature: 1.0
+    vllm_gpu_memory_utilization: 0.5
+  verl_config_overrides:
```
shanghongsim, line 49: https://github.com/verl-project/verl/blob/main/verl/workers/config/rollout.py
https://github.com/verl-project/verl/blob/main/verl/trainer/config/rollout/rollout.yaml

cite these so user can know what options there are

configs/examples/grpo_verl_nl2sql/train.yaml
```diff
@@ -0,0 +1,54 @@
...
+    data:
+      train_batch_size: 32
+      max_prompt_length: 4096
+      filter_overlong_prompts: True
+    actor_rollout_ref:
+      rollout:
+        name: vllm
+        mode: async
+        multi_turn:
```
shanghongsim, line 58: we should probably set a `max_assistant_turns`

configs/examples/grpo_verl_nl2sql/README.md
shanghongsim: Can you also paste some examples of the data and/or give an overview of the data?

configs/examples/grpo_verl_nl2sql/README.md
aniruddh-alt: Done — the README now links the Spider repo, the Yu et al. (2018) paper, and the reference `evaluation.py`, and spells out which metric this trains on: execution accuracy (run the predicted and the gold SQL against the same DB, compare result sets). Also noted there that `sql_execution_match` treats row order as significant only when the gold query has a top-level `ORDER BY`.

configs/examples/grpo_verl_nl2sql/README.md
aniruddh-alt: Done — added download links for the [Spider site](https://yale-lily.github.io/spider) and the [`xlangai/spider`](https://huggingface.co/datasets/xlangai/spider) mirror, plus the expected contents (`train_spider.json`, 7,000 examples; `dev.json`, 1,034).

configs/examples/grpo_verl_nl2sql/train.yaml
aniruddh-alt: Done — both links are now a comment above `verl_config_overrides`.

configs/examples/grpo_verl_nl2sql/train.yaml
aniruddh-alt: Set `max_assistant_turns: 5`.

Worth flagging what leaving it unset actually meant here: verl's config comment says `null` → `max_length // 3`, but that default is only applied on the sglang rollout path. `tool_agent_loop` just skips the check when the value is falsy (`tool_agent_loop.py:85`), so with `default_agent_loop: tool_agent` we had no turn cap at all — the only backstop was `max_completion_length: 1024`.

5 leaves room for a few exploratory `run_sql` calls plus the final answer. The token budget still binds first in practice, so this is an explicit guard rather than a behavior change against the run we validated.

configs/examples/grpo_verl_nl2sql/README.md
aniruddh-alt: Added a **Prepared rows** section: a field-by-field description of what the builder emits (`messages[0]` schema-primed system prompt, `messages[1]` question, `metadata.agent_name` / `ground_truth` / `tools_kwargs.run_sql.create_kwargs.db_path`), plus an example row — real builder output, with the schema abridged to two tables so it stays readable.

shanghongsim:
Thank you Ani for this. Could you walk me through how the training was done and how the model was tested? Also have the corresponding training metrics as that will be super helpful in understanding model training dynamics.

In PR https://github.com/oumi-ai/oumi/pull/2560 you mentioned you validated train results: 

> Exercised end to end through the full stack (this adapter, the NL2SQL env/reward, and the Spider configs): real Spider, job 126, 1×H100, Qwen2.5-3B, val EX went from 0.367 to 0.684.

aniruddh-alt:
# How this example was validated

## What the run does, in plain terms

We take an off-the-shelf 3B chat model and give it two things: a database schema in
its system prompt, and one tool called `run_sql`. It gets a question in English
("How many singers do we have?"). It can call `run_sql` as many times as it likes to
poke at the database, then it writes a final SQL query in a ```sql block.

To score that answer we run **its** SQL and the **gold** SQL against the same SQLite
file and compare the rows that come back. Same rows means reward 1.0, anything else
means 0.0. There is no partial credit and no string comparison of the SQL text — two
completely different-looking queries both score 1.0 if they return the same rows.

GRPO then does the learning. For each question it samples 8 different attempts,
compares their rewards against each other, and pushes the model toward the attempts
that scored better. That is the whole loop.

## The five steps to reproduce

1. **Get Spider.** You need `train_spider.json`, `dev.json`, and the databases laid
   out as `database/<db_id>/<db_id>.sqlite`.

2. **Build the rows.**
   ```bash
   python scripts/datasets/build_spider_tool_agent.py \
     --spider-root /path/to/spider --db-root /path/to/spider/database
   ```
   This writes `data/grpo_verl_nl2sql/{train,val}.jsonl`. One row per question. The
   row does not contain the database, only the absolute path to it, under
   `metadata.tools_kwargs.run_sql.create_kwargs.db_path`. The builder also runs each
   gold query once and drops the row if the gold itself errors — we got
   **6,997 of 7,000 train rows** (3 dropped) and **all 1,034 dev rows**.

3. **Train.**
   ```bash
   oumi train -c configs/examples/grpo_verl_nl2sql/train.yaml
   ```

4. **What verl does with that.** `default_agent_loop: tool_agent` plus
   `multi_turn.enable: true` means each rollout is a real multi-turn conversation, not
   one shot. The `run_sql` tool comes from `verl_tool_config.yaml`, which points at
   `OumiVerlTool`, which points at `environment_config.yaml`, which points at the
   executor `oumi.environments.examples.nl2sql.run_sql`. Each rollout opens its own
   read-only SQLite connection on that row's `db_path`.

5. **Scoring.** `reward_functions: ["sql_execution_match"]`. Same function for training
   reward and for validation.

## Exactly what was run

| | |
|---|---|
| model | `Qwen/Qwen2.5-3B-Instruct` |
| hardware | 8×H100, single node |
| verl | 0.7.1 |
| steps | 100 GRPO steps, 2h30m wall clock, `EXIT=0` |
| batch | 64 prompts × `n=8` samples = **512 trajectories per step** |
| completion budget | 2048 tokens, shared across all turns of a rollout |
| sampling | temperature 1.0 for training, greedy (`t=0, n=1`) for validation |
| validation set | the **full** 1,034-row Spider dev set, every 25 steps |
| turns actually used | `num_turns/mean` ranged 2.2 to 5.4 over training |

## The numbers

Execution match on the full 1,034-row dev set:

| step | exec-match |
|---|---|
| 0 (untrained base) | 0.3936 |
| 25 | 0.6963 |
| 50 | 0.6809 |
| 75 | 0.7157 |
| 100 | **0.7273** |

So 0.39 → 0.73 over 100 steps. Most of the gain lands in the first 25 steps.

## The config that ran is not byte-identical to the `train.yaml` in this PR

Being upfront about this, since it is the thing I would want to know as a reviewer.
The validated run used a full-node config tuned for our cluster. The `train.yaml`
shipping here is the portable version. Differences:

| knob | validated run | `train.yaml` in this PR |
|---|---|---|
| `max_completion_length` | 2048 | 1024 |
| `train_batch_size` | 64 | 32 |
| `tensor_model_parallel_size` | 1 (8-way data parallel) | 2 |
| `max_assistant_turns` | unset (uncapped) | 5 |
| `eval_steps` | 25 | 20 |
| `vllm_gpu_memory_utilization` | 0.6 | 0.5 |
| `agent.num_workers` | 32 | verl default |
| attention | forced `sdpa`, `use_remove_padding: false` | not overridden |
| steps | capped at 100 | 1 full epoch |

The attention override was a workaround for a flash-attn/cutlass mismatch in our
container image and is deliberately not in the shipped config. The rest are real
differences: shorter completion budget and a 5-turn cap in particular could change
results, because the validated run averaged up to 5.4 turns and some rollouts used
more.

## What the metric is, and what it is not

`sql_execution_match` does this:

1. Pull the model's final SQL: last ```sql fence, else the last `SELECT` onward, else
   the last non-empty line.
2. Open the row's SQLite file read-only (a sqlite authorizer denies anything that is
   not a read).
3. Run gold and prediction on the **same connection**, compare result sets.
4. Row order only matters when the gold query has a top-level `ORDER BY`. Otherwise
   both sides are sorted before comparison.
5. If the prediction raises a SQL error, that is 0.0, not a crash.
6. Non-UTF-8 bytes in the DB are decoded with `errors="replace"` for both sides, so
   the comparison stays fair and a bad byte cannot kill the training run.

**It is not the official Spider `evaluation.py`.** So 0.7273 is not directly
comparable to a published Spider leaderboard number.

## Known gaps

- `log_val_generations: 0`, so no validation generations were saved. The headline
  number cannot be audited by reading model outputs.
- Single seed. No error bars.
- The training reward is also the validation metric.
- The baseline is step 0 of the same run (untrained base model), not a separately
  prompted or tuned baseline.
- A checkpoint exists at `global_step_100` but was never evaluated independently.
- `response_length/mean` is non-monotonic: 299 → 44 → 128 tokens across the run.

## Cheapest ways to check me

Roughly in order of value per minute:

1. **Gold-vs-gold sanity check.** Score every dev row's gold SQL as if it were the
   prediction. It should return 1.0 on all 1,034. Anything less is a reward bug.
   Minutes to run, and it independently validates the metric.
2. **Score the `global_step_100` checkpoint with the official Spider
   `evaluation.py`** to get a number comparable to published work.
3. **Re-run 25 steps with `log_val_generations` above 0** and read ~20 rollouts, to
   confirm the model really calls `run_sql` and emits SQL rather than gaming the
   reward.
4. **Second seed using the `train.yaml` in this PR**, so the config being merged is
   the config being measured.

PR #2563, opened 2026-07-24
feat(verl): UserSimToolAgentLoop for tool + simulated-user GRPO rollouts

aniruddh-alt:
## What

verl 0.8 deletes `verl.interactions.base.BaseInteraction`, which the conversational
GRPO rollout was built on. This replaces it with a registered agent-loop subclass:

```python
class UserSimToolAgentLoop(ToolAgentLoop):
    """ToolAgentLoop plus a simulated user that speaks when the assistant stops."""
```

One class covers tools, simulated-user conversations, and both together. Rows select it
through verl's `agent_name` field.

## Why now

verl deprecated `verl/interactions/` together with the rollout chat-completion API in
favour of the token-in-token-out agent loop (`AgentLoopBase`, present since 0.4.2).
There is no built-in user simulator upstream — `SingleTurnAgentLoop` and `ToolAgentLoop`
are the only two — so subclassing is the sanctioned path rather than a workaround.

This is **not** a bug fix. `OumiVerlInteraction` worked: it ran end to end on the cluster
(job 273, `num_turns` mean 7.08, judge reward moving, GRPO updating the policy). We are
replacing validated code whose upstream API is being removed.

- <https://github.com/verl-project/verl/discussions/2744>
- <https://verl.readthedocs.io/en/latest/advance/agent_loop.html>

## Architecture

```
core/rollout/user_sim.py              RolloutState, next_user_turn    <- no verl
verl_agentic/user_sim_provider.py     cached inference engine         <- no verl
verl_agentic/user_sim_agent_loop.py   tokens, masks, state machine    <- only verl import
```

Core speaks messages and rewards; the adapter speaks token ids and masks. A backend swap
rewrites one file. This is the anti-corruption-layer shape agreed in the design thread,
and it mirrors `oumi_verl_tool.py` + `env_provider.py`, which already do it for tools.

## How tools and conversation compose

Nothing in the loop dispatches between modes. `_handle_generating_state` delegates to
`super()`, which returns `PROCESSING_TOOLS` when the assistant emitted a tool call and
`TERMINATED` when it did not. We intercept only the `TERMINATED` branch.

| Row | Behaviour |
|---|---|
| `tools_kwargs`, no `interaction_kwargs` | `_sim` is `None`; identical to stock `ToolAgentLoop` |
| `interaction_kwargs`, no tools | `tool_calls` is always empty, so every stop hands to the simulator |
| both | tools run while the model calls them; the first stop without a call hands over, and tools stay available afterwards |

Expected transcript for the combined case:
`user -> assistant(tool_call) -> tool -> assistant(text) -> sim_user -> assistant`

## The training invariant

`response_mask` is the RL contract: `1` means the policy produced this token and it earns
gradient, `0` means environment observation. Simulated-user turns and tool results are both
environment text, so both are masked `0`. Mark them `1` and you train the model to write the
user's side — training completes, loss descends, and the model is silently wrong. There is no
runtime signal, so this is guarded by tests and by a mask probe over a real rollout.

`_append_environment_turn` is the only place tokens are mutated, and it extends
`prompt_ids` and `response_mask` together so the prompt/response boundary cannot drift.

## Config

Defaults are unchanged — they are shared with the single-turn recipes. `verl_trainer_config.yaml`
documents what an agent-loop run needs, and `VerlGrpoTrainer._validate_agent_loop_config`
raises at config time (before any GPU work) if rows route to the agent loop without it:

```yaml
actor_rollout_ref:
  rollout:
    mode: async
    multi_turn:
      enable: true
    agent:
      agent_loop_config_path: <yaml registering oumi_user_sim_tool_agent>
```

## Verification

Cluster, verl 0.7.1:

- tools + simulator: `EXIT=0`, 11 tool calls, `num_turns` mean 12.3
- conversation only: `EXIT=0`, `num_turns` mean 9.5, reward moving, policy updating
- mask probe over a real rollout: policy tokens at `1` (assistant text and its own
  tool-call emission), environment tokens at `0` (simulated-user turns, tool responses)

Local: 85 passed / 5 skipped without verl; the adapter suite passes against **verl 0.7.1 and
0.8.0** alike; 1034 configs parse; ruff and pyright clean.

## Notes for reviewers

- **No `@register` decorator.** verl's decorator does
  `_agent_loop_registry[name] = {"_target_": fqdn}` — an unconditional overwrite that drops
  every other key. Importing this module is what resolves `_target_`, so decorating it would
  wipe the `user_sim_inference` path the agent-loop YAML supplies. Registration comes from
  `agent_loop_config_path`. A regression test covers it.
- **Every row carries `agent_name`,** including plain ones (`single_turn_agent`).
  `Dataset.map` fixes the output schema from the first row, so a mix of rows with and without
  the key either raises or silently drops the column, routing agent-loop rows to single-turn
  generation with no error.
- **`verl>=0.7`.** The loop needs `AgentLoopBase.apply_chat_template`, which 0.6 lacks.
  The pin cannot move to 0.8 yet for an unrelated reason: 0.8 removed
  `verl.workers.fsdp_workers`, and `AsyncActorRolloutRefWorker` / `CriticWorker` have no
  counterpart in `engine_workers.py`, so `VerlGrpoTrainer` will not import. That migration is
  its own PR; the loop itself is already verified on 0.8.
- **The simulated user sees user and assistant turns only.** Passing verl's `messages`
  through would hand it tool turns with no assistant `tool_calls` preceding them, which chat
  APIs reject outright.

## Out of scope

The LLM-judge reward (#2564) and the demo recipe (#2565). The example configs here register
the loop and the user-sim engine; they do not constitute a runnable training recipe on their own.


src/oumi/core/rollout/user_sim.py
```diff
@@ -0,0 +1,79 @@
...
+
+def next_user_turn(
+    state: RolloutState,
+    messages: list[dict],
+    infer_fn: Callable[[Conversation], str],
+    done_sentinel: str = DEFAULT_DONE_SENTINEL,
+) -> tuple[bool, str, float]:
+    """Produces the next simulated-user turn."""
+    state.turn_idx += 1
+    if state.turn_idx >= state.max_turns:
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed — the cap check now runs before the increment, so `max_turns` means exactly "simulated user turns produced," and the last generated turn's prompt reads "turn N of at most N". `RolloutState`'s docstring now states that the dataset's opening user message isn't counted. `test_next_user_turn_hits_cap` is replaced by `test_next_user_turn_produces_exactly_max_turns`, which walks the boundary at `max_turns=1` (first call generates, second ends).

Related bug this surfaced: `build_user_turn_prompt` hardcoded `DEFAULT_DONE_SENTINEL` in the instruction text while `next_user_turn` checked the configurable `done_sentinel`, so setting `done_sentinel` in config meant early termination never fired. The sentinel is threaded through now, with a test.

src/oumi/core/trainers/verl_conversational_interaction.py
```diff
@@ -0,0 +1,87 @@
...
+    ) -> tuple[bool, str, float, dict]:
+        """Produce the next simulated-user turn."""
+        state = self._state[instance_id]
+        done, text, score = await asyncio.to_thread(
+            next_user_turn, state, messages, self._infer_one, self._done_sentinel
+        )
+        return done, text, score, {}
+
+    def _infer_one(self, conversation: Conversation) -> str:
+        results = self._engine.infer(
```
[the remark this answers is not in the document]
aniruddh-alt: Good catch. Serialized with a `threading.Lock`, but only for local engines — an engine that is a `RemoteInferenceEngine` gets a `contextlib.nullcontext()` instead, since remote `infer()` goes through `safe_asyncio_run` (fresh event loop per thread) and is safe to call concurrently. Locking those too would bottleneck the intended path at rollout scale. Keying off the class hierarchy rather than an engine-type list means it can't drift as new remote providers are added.

src/oumi/core/trainers/verl_conversational_interaction.py
```diff
@@ -0,0 +1,87 @@
...
+            [conversation], inference_config=self._infer_config
+        )
+        if not results or not results[0].messages:
+            return ""
+        content = results[0].messages[-1].content
+        return content if isinstance(content, str) else ""
```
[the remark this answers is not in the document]
aniruddh-alt: Agreed. `_infer_one` now raises on non-text content instead of returning `""`, and the `not results or not results[0].messages` guard is gone — it only fired on an engine bug, and swallowing it made a misconfigured user-sim indistinguishable from a legitimately empty completion. A config error should fail loudly at rollout start rather than quietly burning turns to the cap.

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -249,6 +250,13 @@ def _extract_prompt_images_answer_from_conversation(
     def _create_verl_data_entry_from_conversation(
         example: dict, idx: int, data_source: str, split: str
     ) -> dict:
+        # SFT datasets preserve conversation metadata in conversation_json.
+        conversation = Conversation.from_json(example["conversation_json"])
```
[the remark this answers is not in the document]
aniruddh-alt: Fixed both. The metadata check is now a plain `json.loads` + `.get("metadata", {}).get("interaction_kwargs")`, so pydantic validation only happens on the interaction branch; `Conversation.from_json` moved inside that branch.

For the missing-key case I used `example.get("conversation_json") or "{}"` rather than duplicating the error text: a missing key yields no `interaction_kwargs`, falls through to `_extract_prompt_images_answer_from_conversation`, and surfaces the existing descriptive `ValueError` from `grpo_utils` — one message, one place. Added `test_missing_conversation_json_raises_with_context`.

src/oumi/core/trainers/verl_conversational_interaction.py
```diff
@@ -0,0 +1,87 @@
...
+
+    async def calculate_score(self, instance_id, **kwargs) -> float:
+        """Turn-level score is unused; reward is trajectory-level."""
+        return 0.0
+
+    async def finalize_interaction(self, instance_id, **kwargs) -> None:
+        """Drop the rollout's user-sim state."""
+        self._state.pop(instance_id, None)
```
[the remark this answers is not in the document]
aniruddh-alt: Leaving this one. Each entry is a persona string plus two ints, and it only leaks on rollouts that abort before `finalize_interaction`, so the bound is negligible next to the training run's own footprint. The clean fix is a try/finally at the caller, which is verl's rollout loop rather than our code. Happy to revisit if a long run shows the dict actually growing.

src/oumi/core/trainers/verl_conversational_interaction.py
```diff
@@ -0,0 +1,101 @@
...
+        self._state[iid] = RolloutState(
+            persona=user_persona,
+            goal=goal,
+            max_turns=(max_turns if max_turns is not None else self._default_max_turns),
+        )
+        return iid
+
+    async def generate_response(
+        self, instance_id, messages, **kwargs
+    ) -> tuple[bool, str, float, dict]:
```
shanghongsim: please add comment on what the output stands for. An example from verl's docs on `generate_response` from `BaseInteraction`: 

```
Generate response, return (should_terminate, response, score, metadata)
```

src/oumi/core/rollout/user_sim.py
```diff
@@ -0,0 +1,99 @@
...
+
+def next_user_turn(
+    state: RolloutState,
+    messages: list[dict],
+    infer_fn: Callable[[Conversation], str],
+    done_sentinel: str = DEFAULT_DONE_SENTINEL,
+) -> tuple[bool, str, float]:
+    """Produces the next simulated-user turn, or ends the rollout at the cap."""
+    if state.turn_idx >= state.max_turns:
+        return True, "", 0.0
```
shanghongsim, line 100: What are these outputs? Add comment to indicate what they represent

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -266,6 +279,46 @@ def _create_verl_data_entry_from_conversation(
...
+                "customer message)."
+            )
+        prompt_messages = [
+            {"role": m.role.value, "content": m.compute_flattened_text_content()}
+            for m in messages
+        ]
+        goal = interaction_kwargs.get("goal", "")
+        resolved_kwargs = {
+            "name": interaction_kwargs.get("name", "oumi_conversation"),
+            "user_persona": interaction_kwargs["user_persona"],
```
shanghongsim: Why is user_persona mandatory and goal optional? I see here that if goal is not specified, we just feed empty string

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -266,6 +279,46 @@ def _create_verl_data_entry_from_conversation(
...
+            "max_turns": interaction_kwargs.get("max_turns", 6),
```
shanghongsim: How did you decide on the default value of 6 turns? We should also put the 6 in a default constant

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -249,6 +251,17 @@ def _extract_prompt_images_answer_from_conversation(
     def _create_verl_data_entry_from_conversation(
         example: dict, idx: int, data_source: str, split: str
     ) -> dict:
+        # Cheap metadata peek; a missing key falls through to the extractor's error.
```
shanghongsim: > a missing key falls through to the extractor's error.

What do you mean? Remove if not necessary or make more clear. I don't understand what its trying to say

shanghongsim (changes requested):

src/oumi/core/trainers/verl_conversational_interaction.py
aniruddh-alt: Fixed

src/oumi/core/rollout/user_sim.py
aniruddh-alt: Fixed

src/oumi/core/trainers/verl_grpo_trainer.py
aniruddh-alt: Fixed

src/oumi/core/trainers/verl_grpo_trainer.py
aniruddh-alt: `user_persona` is required because it defines the simulated user’s behavior. `goal` is optional because some persona-driven conversations are intentionally open-ended; when omitted, the goal section is not added to the simulator prompt.

src/oumi/core/trainers/verl_grpo_trainer.py
aniruddh-alt: Six is a conservative fallback rather than a benchmark-derived value: it permits several exchanges while bounding rollout cost if the termination sentinel is never emitted. It remains configurable globally or per example. I moved it into the shared `DEFAULT_MAX_TURNS` constant.

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -353,6 +363,46 @@ def _create_verl_data_entry_from_conversation(
...
+            "name": interaction_kwargs.get("name", "oumi_conversation"),
+            "user_persona": interaction_kwargs["user_persona"],
+            "max_turns": interaction_kwargs.get("max_turns", DEFAULT_MAX_TURNS),
+            "goal": goal,
+        }
+        return {
+            "data_source": data_source,
+            "prompt": prompt_messages,
+            "images": [],
+            "ability": "conversation",
```
shanghongsim: Need to add ` "agent_name": ...,` to invoke the `ToolAgentLoop`

aniruddh-alt: This PR is scoped to conversational rollouts, hoping to integrate with tool agent loops in subsequent PRs!

PR #2556, opened 2026-07-23
feat(verl): conversational GRPO rollout (persona simulated-user + LLM-judge reward)

aniruddh-alt:
## Summary

Adds a **conversational GRPO** track for the verl integration: train a policy to hold a
multi-turn conversation with a **simulated user**, where the user side is a
persona-conditioned LLM turn (reusing how Oumi already does multi-turn conversation
synthesis). This is the sibling to the tool track — it uses verl's **Interaction** system
(`BaseInteraction`) instead of the tool agent loop. Everything rides through
`verl_config_overrides`; **no trainer architecture change.**

First task is **customer support** (user = customer with an issue, policy = support agent),
scored by an **LLM-judge reward** over the finished conversation.

> ⚠️ **Stacks on #2533** (multi-turn GRPO data path). This branch forked from that work, so
> the diff against `main` includes its 4 data-path commits as well as the 13 conversational
> commits here. Review the `feat(verl)/feat(grpo)` conversational commits; this should land
> **after** #2533. Happy to re-base onto the #2533 branch if that's preferred for review.

## What's included

1. **`verl_conversational_turn.py`** — verl-free core: per-rollout state + build-user-turn-prompt
   + "produce next user turn / decide to end." Pure logic, unit-tested without verl.
2. **`verl_conversational_interaction.py`** — `OumiVerlInteraction(BaseInteraction)`, the thin
   adapter verl drives. Holds a remote user-sim inference engine; delegates all turn logic to
   the core. Lifecycle: `start_interaction` → (policy turn → `generate_response` for the user
   turn)\* → `finalize_interaction`.
3. **`conversation_judge_reward.py`** (+ `rewards/__init__.py`) — a `conversation_llm_judge`
   reward function wrapping the existing `SimpleJudge` to score the transcript against the
   customer's goal.
4. **`verl_grpo_trainer.py`** — data-path branch: emits `extra_info.interaction_kwargs`
   (persona, goal, max_turns) for conversations carrying interaction metadata (read from
   `conversation.metadata`, which survives the SFT loader). The existing final-turn path is
   unchanged.
5. **`train.py`** — `ray.init` respects `OUMI_RAY_NUM_CPUS` (some images misreport CPU count,
   which otherwise hangs `ray.init`).
6. **`configs/examples/verl_conversational/`** — interaction registry, remote user-sim config,
   judge config, a small customer-support seed set, the training YAML, and the launch sbatch.

## How one rollout flows

[...]

PR #2564, opened 2026-07-24
feat(grpo): conversation LLM-judge reward

aniruddh-alt:
## What

`conversation_llm_judge_reward`: an LLM-as-judge reward for conversational GRPO. It scores a finished conversation with a judge model and maps the verdict to a scalar, registered under the GRPO reward registry.

Independent of the interaction adapter — the reward scores a trajectory and doesn't import the rollout machinery. Base: main.

Heads-up: `rewards/__init__.py` here and in #2561 (`sql_execution_match`) both append to the same registry list, so whichever lands second hits a one-line merge conflict.

## Test plan

`pytest tests/unit/datasets/grpo/rewards/test_conversation_judge_reward.py` — 3 pass on main (judge mocked).

## Related — split of #2556

#2556 was split into four PRs (also rebased onto main — it was 13 commits behind):
1. #2563 — `OumiVerlInteraction` core. Independent, base main.
2. **#2564 (this)** — conversation LLM-judge reward. Independent, base main.
3. #2565 — customer-support demo configs. Merge last (references #2563 and #2564).
4. #2566 — unrelated Ray-CPU fix, split out. Independent, base main.

### DIFF
--- a/src/oumi/datasets/grpo/rewards/conversation_judge_reward.py
+++ b/src/oumi/datasets/grpo/rewards/conversation_judge_reward.py
@@ -1,0 +1,51 @@
+# Copyright 2025 - Oumi
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""LLM-judge reward for conversational GRPO rollouts (wraps SimpleJudge)."""
+
+import functools
+import os
+
+from oumi.core.configs.judge_config import JudgeConfig
+from oumi.core.registry import RegistryType, register
+from oumi.judges.simple_judge import SimpleJudge
+
+
+@functools.lru_cache(maxsize=8)
+def _get_judge(judge_config_path: str) -> SimpleJudge:
+    return SimpleJudge(JudgeConfig.from_path(judge_config_path))
+
+
+@register("conversation_llm_judge", RegistryType.REWARD_FUNCTION)
+def conversation_llm_judge_reward(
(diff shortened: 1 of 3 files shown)


src/oumi/datasets/grpo/rewards/conversation_judge_reward.py
```diff
@@ -0,0 +1,51 @@
...
+            "conversation_llm_judge needs 'judge_config_path' (arg, extra_info, "
+            "or the OUMI_JUDGE_CONFIG_PATH env var)."
+        )
+    judge = _get_judge(path)
+    result = judge.judge([{"conversation": solution_str, "goal": ground_truth or ""}])[
+        0
+    ]
+    scores = result.field_scores or {}
+    value = scores.get("judgment")
+    return float(value) if value is not None else 0.0
```
shanghongsim, line 51: Might be good to add logging when value is None, that should indicate an issue

tests/unit/datasets/grpo/rewards/test_conversation_judge_reward.py
```diff
@@ -0,0 +1,61 @@
...
+    ]
+    assert stub.seen == expected_seen
+
+
+def test_reward_defaults_to_zero_when_no_score(monkeypatch):
+    monkeypatch.setattr(rew, "_get_judge", lambda path: _StubJudge({"judgment": None}))
+    score = rew.conversation_llm_judge_reward(
+        "src", "x", "g", {"judge_config_path": "j.yaml"}
+    )
+    assert score == 0.0
```
shanghongsim, line 56: This may be problematic... say the model did output a correct response but because of judge failure, we score the trajectory 0, it pollutes the training signal. I think its better to raise an error or signal (or some way to indicate that the judge failed here) and trigger a retry.

shanghongsim (changes requested):

tests/unit/datasets/grpo/rewards/test_conversation_judge_reward.py
aniruddh-alt: that makes sense! will raise an error here

src/oumi/datasets/grpo/rewards/conversation_judge_reward.py
aniruddh-alt: Will do!

PR #2565, opened 2026-07-24
feat(examples): conversational GRPO customer-support demo

aniruddh-alt:
## What

The runnable customer-support demo that wires the interaction adapter and the judge reward together: example configs, a seed conversation set, and a cluster smoke sbatch.

- `configs/examples/verl_conversational/`: the GRPO `train.yaml`, the verl `verl_interaction_config.yaml` (dispatch to `OumiVerlInteraction`), the persona `user_sim_inference.yaml`, `judge.yaml`, a `seed_support.jsonl` seed set, and `smoke.sbatch`.
- Excludes `verl_interaction_config.yaml` (a verl-format config, not an Oumi schema) from the generic config-parse test, mirroring how the nl2sql verl tool config is handled.

Merge last: the configs reference `OumiVerlInteraction` (interaction PR) and `conversation_llm_judge_reward` (reward PR). The diff is self-contained and its parse test passes on main.

## Test plan

`pytest tests/unit/core/configs/test_parse_configs.py -k verl_conversational` — 6 pass on main after the exclusion.

## Related — split of #2556

#2556 was split into four PRs (also rebased onto main — it was 13 commits behind):
1. #2563 — `OumiVerlInteraction` core. Independent, base main.
2. #2564 — conversation LLM-judge reward. Independent, base main.
3. **#2565 (this)** — customer-support demo configs. Merge last (references #2563 and #2564).
4. #2566 — unrelated Ray-CPU fix, split out. Independent, base main.

PR #2566, opened 2026-07-24
fix(train): cap Ray CPUs via OUMI_RAY_NUM_CPUS

aniruddh-alt:
## What

Cap Ray's CPU count with a new `OUMI_RAY_NUM_CPUS` env var. On shared or containerized nodes `ray.init()` reads the host's CPU count rather than the container's and oversubscribes; setting `OUMI_RAY_NUM_CPUS` passes an explicit `num_cpus` to `ray.init()`.

Unrelated to conversational rollout — split out of #2556 so it can land on its own. Base: main.

## Test plan

No new tests: a single `ray.init()` kwarg gated on an env var. Behavior is unchanged when `OUMI_RAY_NUM_CPUS` is unset.

## Related — split of #2556

Split out of #2556 (this is the unrelated infra piece). The conversational-rollout work is in #2563 (core), #2564 (reward), and #2565 (demo).

### DIFF
--- a/src/oumi/train.py
+++ b/src/oumi/train.py
@@ -254,15 +254,20 @@ def _verl_train(partial_trainer: Callable[[], BaseTrainer]):
         )
     if not ray.is_initialized():
         logger.info("Initializing Ray cluster...")
-        ray.init(
-            runtime_env={
+        ray_init_kwargs: dict = {
+            "runtime_env": {
                 "env_vars": {
                     "TOKENIZERS_PARALLELISM": "true",
                     "NCCL_DEBUG": "WARN",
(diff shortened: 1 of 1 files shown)


shanghongsim:
To help me understand this PR, what motivated this change? Did you see something crash? What are the downsides of not doing this?

aniruddh-alt:
> To help me understand this PR, what motivated this change? Did you see something crash? What are the downsides of not doing this?



> To help me understand this PR, what motivated this change? Did you see something crash? What are the downsides of not doing this?

Yes! i did have a crash, I have this in draft as I am investigating what really happened and what is the best fix for this.

PR #2569, opened 2026-07-26
feat(verl): OumiVerlTool — verl tool RL over any Oumi executable environment

aniruddh-alt:
## Review scope

@shanghongsim #2560 landed on this branch instead of main, so the two halves are folded back together here.

| Part | Files | Status |
| --- | --- | --- |
| verl adapter | `core/trainers/verl_agentic/*`, `verl_grpo_trainer.py` + tests | **approved** in #2560, squash-merged here as `c8378cd` |
| `for_sample(env_kwargs_by_env_id)` | `core/synthesis/tool_router.py` + `test_tool_router.py` | **needs review** — the piece you asked to split out |

Two files: commits `3c4503f`, `3d76881`, `35628a5`, plus round 2 below.

## Round 2 — your review

| Feedback | Commit |
| --- | --- |
| `env_kwargs_by_env` → `env_kwargs_by_env_id`, keyed by id not object | `9a6f8b2` |
| "What do you mean by *no router escapes to close these*?" — teardown comment reworded | `21fdbe9` |
| Enterprise dependency ⇒ higher testing bar | `b9d2873`, `d80da8a` |

On the enterprise question: the api worker's `synth_generate_dataset_activity` drives `ConversationSynthesizer`, which is the only existing `for_sample()` caller, and it calls the zero-arg form. So the path is reached but unchanged. Rather than rest on "additive default", the cross-sample invariants are now pinned (see test plan). Two of them were genuine holes: swapping the teardown loop to `env_by_id_new.values()`, or replacing `built_here` with a `requires_isolation()` filter, each passed the full suite before these tests.

## The part under review

A caller with per-sample environment data, the database one RL rollout should query, had no way in. `env_kwargs` lives on the config, so it had to deep-copy the whole `EnvironmentConfig` (every tool schema and inline lookup table), rewrite one env's kwargs, and build a fresh router per sample, discarding a freshly-built template env each time.

`for_sample()` now takes that data directly:

```python
router.for_sample({"nl2sql_db": {"db_path": row_db_path}})
```

Two guards, both raising rather than silently no-opping: unknown environment ids, and **shared** (non-isolating) envs, where an override would apply to every sample instead of one. Both run before the first build, so a rejected override can't strand an env this pass already rebuilt. On that path `for_sample` raises without returning, so no router would ever close it and a database env would leave its snapshot on disk. A mid-loop `build_environment` failure is handled the same way: whatever the pass built is closed on the way out.

Default is `None`, so `for_sample()` behaves exactly as before and the synthesizer is untouched.

## The approved part, for context

`OumiVerlTool` connects verl's tool-RL loop to any Oumi executable environment: the policy model makes a tool call, the call runs for real against a live environment, and the trajectory is scored into a training signal. Nothing in the adapter is tied to a particular tool or domain.

```
Policy model → verl ToolAgentLoop → OumiVerlTool.execute
  → env_provider.get_or_build_router (keyed by request_id)
  → ToolRouter.route_batch → ExecutableEnvironment.step → executor
```

- One `ToolRouter` per rollout, keyed by `request_id` in a module-level registry. The router holds live environment connections, which aren't picklable, so it stays off `agent_data` (the trajectory is serialized and shipped to a separate reward worker). Teardown runs on `weakref.finalize` at the rollout boundary.
- The parent router is built once per env-config path; every rollout is a `for_sample()` clone carrying its own `create_kwargs`. One env per isolating environment per rollout, so nothing is built and discarded, shared envs are never rebuilt, and the config is never copied.
- Rows tagged `agent_name: tool_agent` skip the usual conversation-to-verl converter: the branch packs the prompt inline, forwards `tools_kwargs`, and reads `ground_truth` from metadata. Any tool-agent dataset flows through it, so this PR carries no `grpo_utils` changes.

## Test plan

Full CPU suite green in CI (`static-checks` runs pre-commit plus `pytest`).

- `test_tool_router.py`: an override rebuilds the isolated env with the sample's data and leaves the parent's params intact; shared env and unknown env id are rejected; neither rejection builds anything; a failed build closes what the pass built. Round 2 adds the cross-sample invariants: an override doesn't leak into the next sample off the same parent, overriding one env doesn't retarget its sibling isolating envs, a failed build leaves the parent's shared envs open, and a failed `on_env_built` closes the env it fired on.
- `test_env_provider.py`: the same `request_id` shares a router; different `request_id`s get independently-seeded isolated envs; each rollout builds its envs exactly once (counts `build_environment` calls); two env configs in one rollout are rejected. Runs in CPU CI, since it doesn't import verl.
- `test_oumi_verl_tool.py`: the adapter routes a call through the rollout's shared router; an invalid env config and an unknown tool name are rejected at construction. Gated on `importorskip("verl")`, and verl sits in the `gpu` extra, so this skips in CPU CI and runs only where verl is installed (scheduled GPU job, or the cluster). Verified against a stubbed verl locally and end to end on the cluster.
- `test_verl_grpo_trainer.py::test_tool_agent_row_carries_agent_name_and_tools_kwargs`.

## Validation

Validated end to end on real Spider (Qwen2.5-3B-Instruct, 8×H100, verl 0.7.1, 100 GRPO steps over 6,997 train rows):

- **Training curve**, `sql_execution_match` on the **full 1,034-row dev set**: 0.394 → 0.696 → 0.681 → **0.716** (evals every 25 steps).
- **Independent evaluation of the saved checkpoint**: reloaded `global_step_100` in a fresh process across all 8 ranks and re-scored the full dev set → **0.735**. This is the load-bearing check: the improvement is in the weights, not in the training loop's own reporting.
- **Audited 200 validation generations**: every one produced a real query, and none of the scoring answers were literal value echoes, so the reward is not being gamed. 143 of 144 scoring trajectories used exactly one `run_sql` call.

Metrics are in the internal W&B project `oumi-nl2sql-verl` (runs `6zk6dwhp` for training, `hk65jihf` for the checkpoint evaluation and the generation table).

Caveats, stated up front:

- This is our execution-match reward, **not** the official Spider `evaluation.py` metric, so it is not comparable to published Spider numbers. It compares result sets on a single database instance, and 39.5% of dev golds return a single scalar, so a semantically wrong query can still score. Test-suite accuracy is the follow-up.
- **Single seed**, no confidence intervals. The last three evals (0.696 / 0.681 / 0.716) sit within the late-phase step-to-step reward noise (σ ≈ 0.05), so read this as a plateau around 0.68–0.73 rather than a converged figure.
- `actor/entropy` collapses 0.408 → 0.040 (`entropy_coeff: 0`, `use_kl_loss: false`), which is consistent with that plateau — more steps will not help without a KL or entropy term.
- No no-tool ablation, so this does **not** establish that the tool caused the gain.

It demonstrates that the integration trains end to end and the reward improves — not a benchmark result.

Tested stack (required): verl **0.7.1** (`<0.8`), vllm 0.14.0, ray 2.56.1, torch 2.9.1+cu128, **trl 0.19.1** — trl must be `<0.20` because verl 0.7.1 imports `trl.AutoModelForCausalLMWithValueHead`. Multi-GPU additionally requires `NCCL_NVLS_ENABLE=0` on hosts where NVLS multicast cannot bind.

## Follow-up

#2571 — `route_batch` runs synchronously inside `async execute`, so a slow environment blocks verl's agent-loop event loop. Deferred: `asyncio.to_thread` isn't a valid fix (the SQLite connection is thread-pinned), so it needs a per-router worker thread and its own cluster validation.

## Related — split of #2544

1. **#2569 (this)** — `for_sample(env_kwargs_by_env_id)`, plus the approved #2560 adapter merged in here.
2. #2561 — NL2SQL `run_sql` executor + `sql_execution_match` reward. Independent, base main.
3. #2562 — Spider demo configs + row builder. Merge last (its configs reference this PR and #2561).

### DIFF
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -111,6 +111,8 @@ class ToolRouter:
         per-sample data (e.g. the database a single rollout should query). Shared
         envs are rejected because the override would silently apply to every sample.
         """
+        # Both checks run before the first build, so a rejected override can't
+        # strand envs already rebuilt in the loop (their teardown needs a router).
         overrides = env_kwargs_by_env or {}
         unknown_env_ids = overrides.keys() - self.env_by_id.keys()
         if unknown_env_ids:
@@ -118,15 +120,20 @@ class ToolRouter:
                 f"Per-sample env_kwargs target unknown environments: "
                 f"{sorted(unknown_env_ids)}. Known: {sorted(self.env_by_id)}"
             )
+        shared_env_ids = [
+            env_id
+            for env_id in overrides
+            if not self.env_by_id[env_id].requires_isolation()
+        ]
+        if shared_env_ids:
+            raise ValueError(
+                f"Environments {sorted(shared_env_ids)} are shared across samples, "
+                "so per-sample env_kwargs cannot be applied to them."
+            )
 
         env_by_id_new: dict[str, BaseEnvironment] = {}
         for env_id, parent_env in self.env_by_id.items():
             if not parent_env.requires_isolation():
-                if env_id in overrides:
-                    raise ValueError(
-                        f"Environment {env_id!r} is shared across samples, so "
-                        "per-sample env_kwargs cannot be applied to it."
-                    )
                 env_by_id_new[env_id] = parent_env
                 continue
             env_params = self.env_params_by_id[env_id]
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -97,7 +97,7 @@ class ToolRouter:
         )
 
     def for_sample(
-        self, env_kwargs_by_env: Mapping[str, Mapping[str, Any]] | None = None
+        self, env_kwargs_by_env_id: Mapping[str, Mapping[str, Any]] | None = None
     ) -> ToolRouter:
         """Return a router safe to use for one sample.
 
@@ -108,13 +108,13 @@ class ToolRouter:
         stateless ``SyntheticEnvironment``) are shared with the parent
         router to avoid the per-sample build + inference-engine attach cost.
 
-        ``env_kwargs_by_env`` replaces ``env_kwargs`` on the rebuilt envs with
+        ``env_kwargs_by_env_id`` replaces ``env_kwargs`` on the rebuilt envs with
         per-sample data (e.g. the database a single rollout should query). Shared
         envs are rejected because the override would silently apply to every sample.
         """
         # Both checks run before the first build, so a rejected override can't
         # strand envs already rebuilt in the loop (their teardown needs a router).
-        overrides = env_kwargs_by_env or {}
+        overrides = env_kwargs_by_env_id or {}
         unknown_env_ids = overrides.keys() - self.env_by_id.keys()
         if unknown_env_ids:
             raise ValueError(
@@ -148,8 +148,9 @@ class ToolRouter:
                     self.on_env_built(fresh)
                 env_by_id_new[env_id] = fresh
         except BaseException:
-            # No router escapes to close these, so release them here. Shared envs
-            # aren't in `built_here` — they belong to the parent.
+            # Per-sample envs are normally closed through the returned router; on this
+            # path there is no return value, so nothing else can reach them. Shared
+            # envs aren't in `built_here` — they belong to the parent.
             for env in built_here:
                 with suppress(Exception):
                     env.close()
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -148,9 +148,8 @@ class ToolRouter:
                     self.on_env_built(fresh)
                 env_by_id_new[env_id] = fresh
         except BaseException:
-            # Per-sample envs are normally closed through the returned router; on this
-            # path there is no return value, so nothing else can reach them. Shared
-            # envs aren't in `built_here` — they belong to the parent.
+            # Raising without returning means no router will ever close these.
+            # `built_here` omits the parent's shared envs, which must stay open.
             for env in built_here:
                 with suppress(Exception):
                     env.close()
--- a/src/oumi/core/synthesis/tool_router.py
+++ b/src/oumi/core/synthesis/tool_router.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Callable, Mapping
+from contextlib import suppress
 from dataclasses import dataclass, replace
 from typing import Any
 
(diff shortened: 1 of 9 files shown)


src/oumi/core/synthesis/tool_router.py
```diff
@@ -95,7 +96,9 @@ def from_environment_config(
             on_env_built=on_env_built,
         )
 
-    def for_sample(self) -> ToolRouter:
+    def for_sample(
```
shanghongsim, line 99: does any enterprise feature depend on this at the moment? if it does, this is critical and will need more robust testing before merge

src/oumi/core/synthesis/tool_router.py
```diff
@@ -104,16 +107,53 @@ def for_sample(self) -> ToolRouter:
...
+                env_params = self.env_params_by_id[env_id]
+                if env_id in overrides:
+                    env_params = replace(env_params, env_kwargs=dict(overrides[env_id]))
+                fresh = build_environment(env_params)
+                built_here.append(fresh)
+                if self.on_env_built is not None:
+                    self.on_env_built(fresh)
+                env_by_id_new[env_id] = fresh
+        except BaseException:
+            # No router escapes to close these, so release them here. Shared envs
```
shanghongsim: > No router escapes to close these, so release them here

What do you mean by this?

src/oumi/core/synthesis/tool_router.py
```diff
@@ -104,16 +107,53 @@ def for_sample(self) -> ToolRouter:
         don't require isolation (e.g. ``DeterministicEnvironment`` and
         stateless ``SyntheticEnvironment``) are shared with the parent
         router to avoid the per-sample build + inference-engine attach cost.
+
+        ``env_kwargs_by_env`` replaces ``env_kwargs`` on the rebuilt envs with
```
shanghongsim: nit: use `env_kwargs_by_env_id` since it is keyed by env_id and not the env object

src/oumi/core/trainers/verl_agentic/env_provider.py
```diff
@@ -0,0 +1,102 @@
...
+        if entry and (create_kwargs := entry.get("create_kwargs")):
+            create_kwargs_by_tool[tool_id] = dict(create_kwargs)
+
+    unknown_tool_ids = create_kwargs_by_tool.keys() - tool_env_map.keys()
+    if unknown_tool_ids:
+        raise ValueError(
+            "Rollout create_kwargs contain unknown tool IDs: "
+            f"{sorted(unknown_tool_ids)}"
+        )
+    env_kwargs_by_env: dict[str, dict[str, Any]] = {}
```
shanghongsim: same nit here

shanghongsim (changes requested):

src/oumi/core/synthesis/tool_router.py
aniruddh-alt: Fixed in 9a6f8b27f.

src/oumi/core/trainers/verl_agentic/env_provider.py
aniruddh-alt: Fixed in 9a6f8b27f.

src/oumi/core/synthesis/tool_router.py
aniruddh-alt: Not yet, but it could in the future! Added some more tests to harden this.

src/oumi/core/synthesis/tool_router.py
aniruddh-alt: A router will not close the environment with just raising an exception, so we also need to close it here.

shanghongsim (approved):
LGTM!

PR #2573, opened 2026-07-29
Apply argument defaults in deterministic environments

aniruddh-alt:
## Summary
- apply JSON Schema defaults to deterministic lookup rows and incoming tool calls
- recursively fill defaults in existing or explicitly defaulted nested objects
- reject lookup rows that become duplicates after normalization

Follows the `_fill` shape agreed in Slack: fill a missing property only when it declares its own `default`, then descend into any object value. So with an `opts` object that has no `default` of its own, `{location: sf}` stays distinct from `{location: sf, opts: {}}` and `{location: sf, opts: {unit: fahrenheit}}`, and those latter two are equivalent.

## Notes for review
- `_validate_lookup_table` normalizes `entry.input` in place, so `sample_grounding` now projects filled defaults too — a row configured as `{location: sf}` emits the fact `{location: sf, unit: fahrenheit, ...}` when `unit` is whitelisted. Intended, and pinned by `test_sample_grounding_includes_filled_defaults`.
- An explicit `null` counts as present and is not replaced by the default.

## Unreachable defaults (addresses review comment 1)

Filling walks `properties` chains only, so a `default` anywhere else would be silently dropped. Two changes close that:

- **Local `$ref`s are resolved**, so a pydantic `model_json_schema()` tool schema fills exactly like the equivalent inline one. Sibling keys win over the target's (pydantic emits `{"$ref": ..., "default": ...}`). Non-local and dangling refs are left alone — `jsonschema` still validates them, they just contribute no defaults.
- **Everything still unreachable is a hard error at construction**, naming the tool and the JSON path — a `default` under `items`, `anyOf`/`oneOf`/`allOf`, `additionalProperties`, and so on. So `Optional[NestedModel]`, `list[NestedModel]`, and `dict[str, NestedModel]` schemas whose nested model has a defaulted field are now rejected rather than silently mismatching later. No config under `configs/` is affected.

The scan walks only subschema positions, so instance data is never read as a declaration: a property literally named `default`, or a `json_schema_extra` annotation holding `{"default": ...}`, is not flagged. Its cycle guard is keyed on reachability, so a `$ref` reached both fillably and not is checked both ways. Filling never inserts a default through a `$ref` already on the path, so a self-referential default terminates instead of recursing forever.

Known gap: a hand-written `$ref` with sibling `properties` merges shallowly rather than applying both schemas per JSON Schema 2020-12. Pydantic never emits that shape.

## Follow-up (not in this PR)
`SyntheticEnvironment._cache_key` keys `cache_by_input` on the raw `json.dumps(arguments, sort_keys=True)`, so it has the same equivalence gap this PR closes for deterministic envs: `{location: sf}` and `{location: sf, unit: fahrenheit}` are two cache entries and can produce two different simulated outputs for the same semantic call. Left out to keep this PR scoped to its title.

## Test Plan
- `PYTHONPATH=src pytest tests/unit/environments tests/unit/core/synthesis/test_tool_router.py -q`
- `uvx ruff@0.15.0 check src/oumi/environments/deterministic_environment.py tests/unit/environments/test_deterministic_environment.py`
- `uvx ruff@0.15.0 format --check src/oumi/environments/deterministic_environment.py tests/unit/environments/test_deterministic_environment.py`
- `pyright src/oumi/environments/deterministic_environment.py tests/unit/environments/test_deterministic_environment.py --pythonpath "$(command -v python)"`

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/environments/deterministic_environment.py
+++ b/src/oumi/environments/deterministic_environment.py
@@ -37,7 +37,10 @@ def _fill_argument_defaults(
     arguments: dict[str, Any],
     schema: dict[str, Any],
 ) -> dict[str, Any]:
-    """Return a copy of arguments with JSON Schema property defaults applied."""
+    """Return a copy of arguments with JSON Schema property defaults applied.
+
+    An absent object property is only created if it declares its own ``default``.
+    """
     result = dict(arguments)
     properties = schema.get("properties", {})
     if not isinstance(properties, dict):
@@ -108,8 +111,9 @@ class DeterministicEnvironment(BaseEnvironment):
         """Initialize a DeterministicEnvironment."""
         self._params = params
         self._kwargs = kwargs
-        self._tools_by_id = {tool.id: tool for tool in params.tools}
-        self._tool_ids = set(self._tools_by_id)
+        self._tools_by_id: dict[str, ToolParams] = {
+            tool.id: tool for tool in params.tools
+        }
         self._validate_lookup_table()
 
     def step(self, calls: list[tuple[str, dict[str, Any]]]) -> list[ToolResult]:
@@ -117,15 +121,13 @@ class DeterministicEnvironment(BaseEnvironment):
         return [self._resolve_one(tool_id, args) for tool_id, args in calls]
 
     def _resolve_one(self, tool_id: str, arguments: dict[str, Any]) -> ToolResult:
-        if tool_id not in self._tool_ids:
+        tool = self._tools_by_id.get(tool_id)
+        if tool is None:
             raise ValueError(
                 f"Tool '{tool_id}' not found in environment '{self._params.id}'. "
-                f"Available tools: {sorted(self._tool_ids)}"
+                f"Available tools: {sorted(self._tools_by_id)}"
             )
-        arguments = _fill_argument_defaults(
-            arguments,
-            self._tools_by_id[tool_id].parameters,
-        )
+        arguments = _fill_argument_defaults(arguments, tool.parameters)
         entries = self._kwargs.lookup_table.get(tool_id, [])
         for entry in entries:
             if entry.matches(arguments):
@@ -192,10 +194,11 @@ class DeterministicEnvironment(BaseEnvironment):
         - Stale ``lookup_table`` keys (no matching tool): log a warning;
           entries are dormant.
         - Tools without entries: hard error.
+        - Entry inputs are normalized in place with schema defaults.
         - Duplicate inputs within a tool's entries: hard error.
         """
         for tool_id in self._kwargs.lookup_table:
-            if tool_id not in self._tool_ids:
+            if tool_id not in self._tools_by_id:
                 logger.warning(
                     "Environment '%s': lookup_table.'%s' references unknown "
                     "tool. Entries will be ignored.",
--- a/src/oumi/environments/deterministic_environment.py
+++ b/src/oumi/environments/deterministic_environment.py
@@ -34,28 +34,133 @@ from oumi.environments.base_environment import BaseEnvironment
 from oumi.environments.utils import parse_env_kwargs
 from oumi.utils.logging import logger
 
+# Keywords whose value is a subschema, a list of them, or a map of them. Only
+# these are walked, so instance data never gets read as a declaration.
+# ``$defs``/``definitions`` are absent: inert until reached through a ``$ref``.
+_SUBSCHEMA = frozenset(
+    {
+        "additionalItems",
+        "additionalProperties",
+        "contains",
+        "else",
+        "if",
+        "items",
+        "not",
+        "propertyNames",
+        "then",
+        "unevaluatedItems",
+        "unevaluatedProperties",
+    }
+)
+_SUBSCHEMA_LIST = frozenset({"allOf", "anyOf", "oneOf", "prefixItems"})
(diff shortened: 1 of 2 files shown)


wizeng23:
### Code review

Algorithm looks right — I checked `_fill_argument_defaults` against absent-parent, empty-parent, explicit-form, parent-`default` opt-in, 3-level chain, and explicit-`null` cases and all match the intended semantics. Fill-before-recurse ordering and the `deepcopy` are both correct, and it's applied to both sides of the match. Two comments.

**1. Consider rejecting `$ref` / `oneOf` / `anyOf` / `allOf` at construction**

The PR notes these aren't walked, but the skip is silent: a `$ref`-typed parent leaves its nested defaults unfilled, and a `default` inside `anyOf` is ignored. Neither produces a signal — the author finds out later via a `ToolLookupError` printing two inputs that look equivalent.

Suggest erroring in `_validate_lookup_table` when `parameters` contains any of these keywords at any depth, naming the tool and JSON path. Ajv is precedent: with `useDefaults`, a `default` inside `anyOf` throws at compile time (`strict mode: default is ignored for: ...`) rather than silently misbehaving.

https://github.com/oumi-ai/oumi/blob/a3a0da52fa94a8632a95b2143ca4885d512ea63f/src/oumi/environments/deterministic_environment.py#L45-L56

**2. Follow-up: arguments are schema-validated before defaults are filled**

Not blocking and arguably pre-existing, but this PR makes it reachable. `ToolRouter` validates raw arguments, then `env.step()` fills defaults afterward:

https://github.com/oumi-ai/oumi/blob/5cfe5d5dbd04efca03175e760749d52672f5093e/src/oumi/core/synthesis/tool_router.py#L169-L171

So for a schema where `opts` has `default: {}` and `opts.required: ["unit"]` with `unit` defaulted, a call sending `opts: {}` fails validation with "`unit` is a required property" — even though the env would fill `unit` and resolve the row. Narrow (needs nested `required` plus a nested `default` on the same field), but the two components now disagree on what "complete arguments" means.

Fix would be hoisting the fill into a `ToolParams` method so the router can normalize before validating; that also covers `SyntheticEnvironment` and `ExecutableEnvironment`, which validate raw args too.

aniruddh-alt:
> ### Code review
> Algorithm looks right — I checked `_fill_argument_defaults` against absent-parent, empty-parent, explicit-form, parent-`default` opt-in, 3-level chain, and explicit-`null` cases and all match the intended semantics. Fill-before-recurse ordering and the `deepcopy` are both correct, and it's applied to both sides of the match. Two comments.
> 
> **1. Consider rejecting `$ref` / `oneOf` / `anyOf` / `allOf` at construction**
> 
> The PR notes these aren't walked, but the skip is silent: a `$ref`-typed parent leaves its nested defaults unfilled, and a `default` inside `anyOf` is ignored. Neither produces a signal — the author finds out later via a `ToolLookupError` printing two inputs that look equivalent.
> 
> Suggest erroring in `_validate_lookup_table` when `parameters` contains any of these keywords at any depth, naming the tool and JSON path. Ajv is precedent: with `useDefaults`, a `default` inside `anyOf` throws at compile time (`strict mode: default is ignored for: ...`) rather than silently misbehaving.
> 
> https://github.com/oumi-ai/oumi/blob/a3a0da52fa94a8632a95b2143ca4885d512ea63f/src/oumi/environments/deterministic_environment.py#L45-L56
> 
> **2. Follow-up: arguments are schema-validated before defaults are filled**
> 
> Not blocking and arguably pre-existing, but this PR makes it reachable. `ToolRouter` validates raw arguments, then `env.step()` fills defaults afterward:
> 
> https://github.com/oumi-ai/oumi/blob/5cfe5d5dbd04efca03175e760749d52672f5093e/src/oumi/core/synthesis/tool_router.py#L169-L171
> 
> So for a schema where `opts` has `default: {}` and `opts.required: ["unit"]` with `unit` defaulted, a call sending `opts: {}` fails validation with "`unit` is a required property" — even though the env would fill `unit` and resolve the row. Narrow (needs nested `required` plus a nested `default` on the same field), but the two components now disagree on what "complete arguments" means.
> 
> Fix would be hoisting the fill into a `ToolParams` method so the router can normalize before validating; that also covers `SyntheticEnvironment` and `ExecutableEnvironment`, which validate raw args too.

agreed on both, added changes to address comment 1 in this PR, and opened a follow up PR for comment 2. Thanks!

src/oumi/environments/deterministic_environment.py
```diff
@@ -34,6 +35,29 @@
...
 
+def _fill_argument_defaults(
+    arguments: dict[str, Any],
+    schema: dict[str, Any],
+) -> dict[str, Any]:
+    """Return a copy of arguments with JSON Schema property defaults applied.
+
+    An absent object property is only created if it declares its own ``default``.
+    """
+    result = dict(arguments)
```
wizeng23: nit: might be safer to deepcopy as well here so the returned result doesn't reference any values from the input.

src/oumi/environments/deterministic_environment.py
```diff
@@ -172,10 +200,11 @@ def _validate_lookup_table(self) -> None:
         - Stale ``lookup_table`` keys (no matching tool): log a warning;
           entries are dormant.
```
wizeng23: nit: can you mention here the stale keys don't get normalized?

wizeng23 (approved):
Approving assuming outstanding comment gets addressed

PR #2574, opened 2026-07-30
Validate deterministic env lookup_table rows against tool schemas

aniruddh-alt:
Stacked on #2573 — review that one first; this diff is only the last commit.

## Problem

`DeterministicEnvironment` never schema-validated its `lookup_table` rows.
`_validate_lookup_table` caught duplicate inputs and tools with no entries, but
nothing called `tool.validate_arguments(entry.input)`, and `output_schema` was
declared on `ToolParams` and never enforced anywhere.

So this env constructed fine:

```yaml
tools:
  - id: get_weather
    parameters:
      properties:
        location: {type: string}
      required: [location]
      additionalProperties: false
lookup_table:
  get_weather:
    - input: {locaton: "sf"}      # typo
      output: {temperature: 65}
```

The row is unreachable: `location` is required, so any valid call carries it and
never matches this input. You only found out mid-run via `ToolLookupError`. With
`additionalProperties: false` it's worse — those arguments also fail
`validate_arguments` in `ToolRouter`, so the model can never produce a matching
call at all. A guaranteed-dead config, silently accepted.

Credit to @wizeng23 for catching this while reviewing #2573.

## Change

Validate every row at construction, inside the existing per-entry loop in
`_validate_lookup_table`:

- **Inputs** via `tool.validate_arguments(entry.input)`.
- **Outputs** via `output_schema` when the tool declares one — a bad output is
  worse than a bad input, since it flows into generated data rather than erroring.
- `ToolArgumentError` is re-raised as `ValueError`, matching the other
  construction-time errors in that method. `ToolError` subclasses are for the
  LLM self-correction loop, not config validation.

Ordering matters: validation runs **after** `_fill_argument_defaults`, so a row
that is invalid raw but valid once filled (a nested `required` field satisfied by
a nested `default`) is accepted. `test_entry_input_is_validated_after_defaults_are_filled`
pins that ordering — it only fails if validation moves before the fill.

## Breaking change

Any existing config with a sloppy unreachable row now fails at construction
instead of at runtime. That is the check working as intended, but it is a
behavior change for configs that "worked" only because the dead row was never
hit.

`configs/examples/synthesis/library_tool_use_synth.yaml` is the only in-repo
config with a `lookup_table` (9 rows); verified it still constructs.

## Tests

- Invalid input (typo'd key) → `ValueError`
- Validation runs after defaults are filled (ordering pin)
- Invalid output vs `output_schema` → `ValueError`
- Valid output constructs and resolves through `step`

One `test_tool_router.py` fixture supplied a schema-invalid `{}` row and is
updated to a conforming one.

417 passed in `tests/unit/environments/` + `tests/unit/core/synthesis/`.

## Update: non-JSON outputs

`jsonschema` accepts Python values that `ToolResult.output` (`JsonValue`) rejects
— a dict with int keys passes `{"type": "object"}`, and YAML preserves numeric
mapping keys through OmegaConf. Such a row passed validation and then raised at
the first `step()` that resolved it, which is the mid-run failure this PR exists
to prevent. Each output is now checked against `ToolResult` — the actual consumer
— before schema validation. Verified end-to-end from a YAML config.

### DIFF
--- a/src/oumi/environments/deterministic_environment.py
+++ b/src/oumi/environments/deterministic_environment.py
@@ -37,7 +37,10 @@ def _fill_argument_defaults(
     arguments: dict[str, Any],
     schema: dict[str, Any],
 ) -> dict[str, Any]:
-    """Return a copy of arguments with JSON Schema property defaults applied."""
+    """Return a copy of arguments with JSON Schema property defaults applied.
+
+    An absent object property is only created if it declares its own ``default``.
+    """
     result = dict(arguments)
     properties = schema.get("properties", {})
     if not isinstance(properties, dict):
@@ -108,8 +111,9 @@ class DeterministicEnvironment(BaseEnvironment):
         """Initialize a DeterministicEnvironment."""
         self._params = params
         self._kwargs = kwargs
-        self._tools_by_id = {tool.id: tool for tool in params.tools}
-        self._tool_ids = set(self._tools_by_id)
+        self._tools_by_id: dict[str, ToolParams] = {
+            tool.id: tool for tool in params.tools
+        }
         self._validate_lookup_table()
 
     def step(self, calls: list[tuple[str, dict[str, Any]]]) -> list[ToolResult]:
@@ -117,15 +121,13 @@ class DeterministicEnvironment(BaseEnvironment):
         return [self._resolve_one(tool_id, args) for tool_id, args in calls]
 
     def _resolve_one(self, tool_id: str, arguments: dict[str, Any]) -> ToolResult:
-        if tool_id not in self._tool_ids:
+        tool = self._tools_by_id.get(tool_id)
+        if tool is None:
             raise ValueError(
                 f"Tool '{tool_id}' not found in environment '{self._params.id}'. "
-                f"Available tools: {sorted(self._tool_ids)}"
+                f"Available tools: {sorted(self._tools_by_id)}"
             )
-        arguments = _fill_argument_defaults(
-            arguments,
-            self._tools_by_id[tool_id].parameters,
-        )
+        arguments = _fill_argument_defaults(arguments, tool.parameters)
         entries = self._kwargs.lookup_table.get(tool_id, [])
         for entry in entries:
             if entry.matches(arguments):
@@ -192,10 +194,11 @@ class DeterministicEnvironment(BaseEnvironment):
         - Stale ``lookup_table`` keys (no matching tool): log a warning;
           entries are dormant.
         - Tools without entries: hard error.
+        - Entry inputs are normalized in place with schema defaults.
         - Duplicate inputs within a tool's entries: hard error.
         """
         for tool_id in self._kwargs.lookup_table:
-            if tool_id not in self._tool_ids:
+            if tool_id not in self._tools_by_id:
                 logger.warning(
                     "Environment '%s': lookup_table.'%s' references unknown "
                     "tool. Entries will be ignored.",
--- a/src/oumi/environments/deterministic_environment.py
+++ b/src/oumi/environments/deterministic_environment.py
@@ -207,7 +207,9 @@ class DeterministicEnvironment(BaseEnvironment):
           entries are dormant.
         - Tools without entries: hard error.
         - Entry inputs are normalized in place with schema defaults.
-        - Entry inputs and outputs must conform to the tool schemas.
+        - Entry inputs must conform to the tool's ``parameters``; outputs must
+          be JSON values, and conform to ``output_schema`` when the tool
+          declares one (it is optional).
         - Duplicate inputs within a tool's entries: hard error.
         """
         for tool_id in self._kwargs.lookup_table:
--- a/src/oumi/environments/deterministic_environment.py
+++ b/src/oumi/environments/deterministic_environment.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+import copy
 import dataclasses
 import json
 import random
@@ -30,6 +31,26 @@ from oumi.core.registry import register_environment
 from oumi.core.types.tool_call import ToolResult
 from oumi.environments.base_environment import BaseEnvironment
 from oumi.utils.logging import logger
+
+
+def _fill_argument_defaults(
+    arguments: dict[str, Any],
+    schema: dict[str, Any],
+) -> dict[str, Any]:
+    """Return a copy of arguments with JSON Schema property defaults applied."""
+    result = dict(arguments)
+    properties = schema.get("properties", {})
+    if not isinstance(properties, dict):
+        return result
+
+    for name, property_schema in properties.items():
+        if not isinstance(property_schema, dict):
+            continue
+        if name not in result and "default" in property_schema:
+            result[name] = copy.deepcopy(property_schema["default"])
+        if isinstance(result.get(name), dict):
+            result[name] = _fill_argument_defaults(result[name], property_schema)
+    return result
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/oumi/environments/deterministic_environment.py
```diff
@@ -201,6 +207,7 @@ def _validate_lookup_table(self) -> None:
           entries are dormant.
         - Tools without entries: hard error.
         - Entry inputs are normalized in place with schema defaults.
+        - Entry inputs and outputs must conform to the tool schemas.
```
wizeng23: nit: can you mention that the output schema is optional and validation is only done against it if provided?

wizeng23 (approved):

wizeng23:
### Test coverage gaps in `_validate_lookup_table`

Coverage is solid — these are the input/output cases the new validation handles correctly but nothing asserts. One test per gap:

| Test | Previously missing |
|---|---|
| `test_valid_entry_input_constructs` | Input happy path (only output had one) |
| `test_entry_input_missing_required_field_raises` | Missing-required, isolated from extra-key |
| `test_entry_input_unrecognized_field_raises` | `additionalProperties: False` — never reached, see below |
| `test_entry_input_wrong_type_raises` | Wrong scalar type |
| `test_nested_entry_input_violations_raise` | Nested extra-key + nested wrong-type (2 params) |
| `test_nested_default_satisfies_required_field` | Defaults-then-validate, success direction |
| `test_nested_entry_output_violations_raise` | Nested-object + array-item outputs (2 params) |
| `test_output_is_unvalidated_when_tool_declares_no_schema` | `output_schema=None` accepts arbitrary JSON |
| `test_non_json_output_raises_without_output_schema` | `ToolResult` check is independent of `output_schema` |

Two worth calling out:

**The unrecognized-key case never executes today.** [`test_invalid_entry_input_raises`](https://github.com/oumi-ai/oumi/blob/57f0a090ce17e646eacaefe5aed3757104e0b88f/tests/unit/environments/test_deterministic_environment.py#L189-L213) sets `required: ["location"]` *and* `additionalProperties: False`, then passes `{"locaton": "sf"}` — that typo trips both, and jsonschema reports `'location' is a required property` first, so `additionalProperties: False` is dead. Splitting them isolates each rule.

**Your ordering test is load-bearing** — I tried to break it and couldn't: reordering yields `'unit' is a required property`, which misses the `.*fahrenheit.*not one of` regex. It only covers the direction where a default makes validation *fail*, so the new test covers the other half — a nested `required` field satisfied *only* by its schema default.

Mutation-tested rather than assumed: removing `validate_arguments` fails 7, removing the output `jsonschema.validate` fails 7, removing the `ToolResult` check fails 2, reordering defaults/validation fails 2. 47 → 58 in this file, 91 passing with `test_tool_router.py`. `ruff@0.15.0` clean.

<details>
<summary>Diff (208 lines, tests only)</summary>

```diff
_STRICT_PARAMS = {
    "type": "object",
    "properties": {"location": {"type": "string"}},
    "required": ["location"],
    "additionalProperties": False,
}


def test_valid_entry_input_constructs():
    tool = _make_tool(parameters=_STRICT_PARAMS)

    env = DeterministicEnvironment.from_params(
        _make_params(
            tools=[tool],
            lookup_table={
                "tool1": [ToolLookupEntry(input={"location": "sf"}, output={"t": 65})]
            },
        )
    )

    assert env.step([("tool1", {"location": "sf"})]) == [ToolResult(output={"t": 65})]


def test_entry_input_missing_required_field_raises():
    tool = _make_tool(parameters=_STRICT_PARAMS)

    with pytest.raises(ValueError, match="invalid input.*'location' is a required"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={"tool1": [ToolLookupEntry(input={}, output={})]},
            )
        )


def test_entry_input_unrecognized_field_raises():
    # Distinct from the missing-required case: `location` is present and valid,
    # so `additionalProperties: False` is the only rule the entry can trip.
    tool = _make_tool(parameters=_STRICT_PARAMS)

    with pytest.raises(ValueError, match="invalid input.*'extra' was unexpected"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={
                    "tool1": [
                        ToolLookupEntry(input={"location": "sf", "extra": 1}, output={})
                    ]
                },
            )
        )


def test_entry_input_wrong_type_raises():
    tool = _make_tool(parameters=_STRICT_PARAMS)

    with pytest.raises(ValueError, match="invalid input.*not of type 'string'"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={
                    "tool1": [ToolLookupEntry(input={"location": 123}, output={})]
                },
            )
        )


@pytest.mark.parametrize(
    "bad_input,match",
    [
        ({"opts": {"unit": "c", "bogus": 1}}, "'bogus' was unexpected"),
        ({"opts": {"unit": 99}}, "not of type 'string'"),
    ],
    ids=["nested_unrecognized_field", "nested_wrong_type"],
)
def test_nested_entry_input_violations_raise(bad_input, match):
    tool = _make_tool(
        parameters={
            "type": "object",
            "properties": {
                "opts": {
                    "type": "object",
                    "properties": {"unit": {"type": "string"}},
                    "additionalProperties": False,
                }
            },
        }
    )

    with pytest.raises(ValueError, match=f"invalid input.*{match}"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={"tool1": [ToolLookupEntry(input=bad_input, output={})]},
            )
        )


def test_nested_default_satisfies_required_field():
    # The success half of the defaults-then-validate ordering: `unit` is required
    # but omitted, and only the schema default makes the entry valid. Validating
    # before filling would reject this.
    tool = _make_tool(
        parameters={
            "type": "object",
            "properties": {
                "opts": {
                    "type": "object",
                    "properties": {"unit": {"type": "string", "default": "celsius"}},
                    "required": ["unit"],
                }
            },
            "required": ["opts"],
        }
    )

    env = DeterministicEnvironment.from_params(
        _make_params(
            tools=[tool],
            lookup_table={
                "tool1": [ToolLookupEntry(input={"opts": {}}, output={"ok": True})]
            },
        )
    )

    entry = env._kwargs.lookup_table["tool1"][0]
    assert entry.input == {"opts": {"unit": "celsius"}}


@pytest.mark.parametrize(
    "output_schema,bad_output,match",
    [
        (
            {
                "type": "object",
                "properties": {
                    "d": {
                        "type": "object",
                        "properties": {"t": {"type": "number"}},
                        "required": ["t"],
                    }
                },
                "required": ["d"],
            },
            {"d": {"t": "hot"}},
            "not of type 'number'",
        ),
        (
            {
                "type": "array",
                "items": {
                    "type": "object",
                    "properties": {"id": {"type": "number"}},
                    "required": ["id"],
                },
            },
            [{"id": 1}, {"id": "x"}],
            "not of type 'number'",
        ),
    ],
    ids=["nested_object", "array_items"],
)
def test_nested_entry_output_violations_raise(output_schema, bad_output, match):
    tool = _make_tool(output_schema=output_schema)

    with pytest.raises(ValueError, match=f"invalid output.*{match}"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={"tool1": [ToolLookupEntry(input={}, output=bad_output)]},
            )
        )


def test_output_is_unvalidated_when_tool_declares_no_schema():
    # Without `output_schema` the only output check is JSON-validity, so an
    # output that would violate any reasonable schema still constructs.
    tool = _make_tool(output_schema=None)
    output = {"junk": "nonsense", "n": [1, 2]}

    env = DeterministicEnvironment.from_params(
        _make_params(
            tools=[tool],
            lookup_table={"tool1": [ToolLookupEntry(input={}, output=output)]},
        )
    )

    assert env.step([("tool1", {})]) == [ToolResult(output=output)]


def test_non_json_output_raises_without_output_schema():
    # The ToolResult JSON check is independent of `output_schema`, so an
    # int-keyed dict is rejected even when no schema is declared.
    tool = _make_tool(output_schema=None)

    with pytest.raises(ValueError, match="non-JSON output"):
        DeterministicEnvironment.from_params(
            _make_params(
                tools=[tool],
                lookup_table={
                    "tool1": [
                        ToolLookupEntry(input={}, output={1: "x"})  # type: ignore[arg-type]
                    ]
                },
            )
        )


```

</details>

PR #2576, opened 2026-07-31
Optional env name/description; emit tools on synthesized conversations

wizeng23:
# Description

Two small, independent improvements to the agentic-synthesis path, each in its own commit:

1. **`EnvironmentParams.name`/`description` are now optional.** Only `id` (the reference key used by `MultiTurnAttribute.available_environments`) and `env_type` (the registry key) are required to build an environment. `name` and `description` are human-readable labels; requiring them non-empty forced programmatic callers to fabricate throwaway strings. Hand-authored configs that set them are unaffected.

2. **The multi-turn synthesizer now emits tool definitions on the output conversation.** It already attached the available tools to each assistant *inference prompt*, but not to the returned `Conversation` — so the tool schemas were absent from the synthesized dataset. Downstream tool-use SFT reads a conversation's top-level `tools`, so without this the model never sees the tool schemas it was trained to call. The emitted `Conversation` now carries `tools`.

## Before submitting

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/configs/params/environment_params.py
+++ b/src/oumi/core/configs/params/environment_params.py
@@ -31,12 +31,25 @@ class EnvironmentParams(BaseParams):
     """Pure-data description of an environment."""
 
     id: str = ""
-    name: str = ""
-    description: str = ""
+    """Reference key for the environment, unique within an ``EnvironmentConfig``."""
+
+    name: str | None = None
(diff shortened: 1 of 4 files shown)


aniruddh-alt (approved):
LGTM, thanks for catching the conversation synthesizer bug for not passing the tools used.

PR #2590, opened 2026-08-07
Tool response in assistant span detection and tests

shanghongsim:
# Description

Some chat templates render tool results inside the assistant turn (like Gemma 4), which leaves them unmasked. This results in model being trained on environment output leading to model degradation. 

This seems to be a common problems faced by the community trying to use Gemma 4 for agentic cases: 
https://git.sethpc.xyz/Seth/gemma4-research/src/commit/91aaaa48d7275333e92f76a6b4be6c1a5cdbeb89/tooling/fine-tuning
there has not been a canonical way or fix suggested by the community for this issue

Our fix here is two fold:
1. In `collators.py`: We have a mechanism that autodetects tool_response start and end if the user does not supply it and we detect that tool response is indeed nested in assistant span. 
2. In `trl_data_collator_for_completion_only_lm.py`: we remask the regions between tool_response start and end, if they are supplied.

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: medium

### DIFF
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -48,6 +48,8 @@ _TOOL_FIX_HINT = (
     "collator_kwargs, or use a chat template that renders tool results in their "
     "own turn."
 )
+_SPAN_TRAIN_TARGETS = ("all_assistant_turns", "final_assistant_turn")
+_TOOL_TEMPLATE_KEYS = ("tool_response_template", "end_of_tool_response_template")
 _PROBE_RESULT = "PROBE_TOOL_RESULT_"
 _PROBE_CALL_ID = "probecall"
 _PROBE_TOOLS: list[Any] = [
@@ -284,12 +286,12 @@ def resolve_tool_response_template(
     """Detects the bracket around tool results nested inside assistant spans.
 
     Span masking unmasks everything between ``response_template`` and the next
-    ``end_of_turn_template``. Some chat templates (gemma-4) render tool results
-    *inside* the assistant turn, which puts environment output inside that span.
-    This finds the marker pair bracketing those results so masking can exclude them.
+    ``end_of_turn_template``. Some chat templates (e.g. gemma-4) render tool results
+    inside the assistant turn resulting in environment output being unmasked.
+    The returned bracket lets the collator remask it.
 
-    Returns token ids rather than strings: ``_tokenize_template`` accepts either, and
-    ids avoid a decode/re-encode round trip that isn't guaranteed to be lossless.
+    Returns token ids rather than strings to avoid a decode/re-encode round trip
+    that isn't guaranteed to be lossless.
 
     Returns:
         (opener_ids, closer_ids) when tool results are nested and a specific bracket
@@ -383,6 +385,62 @@ def resolve_tool_response_template(
         closer,
     )
     return [open_id], [close_id]
+
+
+def _resolve_tool_bracket(
+    tokenizer: "BaseTokenizer",
+    collator_kwargs: dict,
+    config_collator_kwargs: dict,
+) -> tuple[list[int], list[int]] | None:
+    """The bracket around tool results this template nests in the assistant turn.
+
+    Some templates render tool results inside the assistant turn, which leaves them
+    unmasked. The returned bracket lets the collator mask them again.
+
+    Returns:
+        (opener_ids, closer_ids), or None when the bracket is already configured, the
+        train_target isn't span-based, or tool results aren't nested.
+
+    Raises:
+        ValueError: Only one of the two markers was supplied, or the template nests
+            tool results but no bracket could be recovered.
+    """
+    supplied = [key for key in _TOOL_TEMPLATE_KEYS if config_collator_kwargs.get(key)]
+    if len(supplied) == 1:
+        # One marker alone masks nothing, and it would also suppress the detection
+        # that reports a nesting template.
+        missing = next(key for key in _TOOL_TEMPLATE_KEYS if key not in supplied)
+        raise ValueError(
+            f"collator_kwargs sets '{supplied[0]}' without '{missing}'. Tool results "
+            "are only excluded from the training loss when both markers are set.\n"
+            f"Fix: also set '{missing}' in collator_kwargs."
+        )
+    if supplied:
+        return None  # Hand-supplied bracket wins; nothing to detect.
+
+    if collator_kwargs.get("train_target") not in _SPAN_TRAIN_TARGETS:
+        return None  # Only span-based masking can unmask a tool result.
+
+    response_template = config_collator_kwargs.get(
+        "response_template", collator_kwargs.get("response_template")
+    )
+    if not response_template:
+        return None
+
+    # Needed only to tell a nested tool result from one in its own turn.
+    turn_boundary = config_collator_kwargs.get(
+        "end_of_turn_template"
+    ) or collator_kwargs.get("end_of_turn_template")
+    if not turn_boundary:
+        # train_target='final_assistant_turn' carries no end_of_turn_template. Detection
+        # can fail, and a hand-supplied response_template is the usual reason it was
+        # bypassed, so give up on the bracket rather than failing the build.
+        try:
+            turn_boundary = resolve_collator_templates(tokenizer)[1]
+        except ValueError:
+            return None
+
+    return resolve_tool_response_template(tokenizer, response_template, turn_boundary)
 
 
 def build_data_collator(
@@ -619,39 +677,12 @@ def build_collator_from_config(
                 "or provide response_template in collator_kwargs."
             )
 
-    # Templates that nest tool results inside the assistant turn put environment output
-    # inside the unmasked span.
-    _tool_keys = ("tool_response_template", "end_of_tool_response_template")
-    supplied_tool_keys = [key for key in _tool_keys if config_collator_kwargs.get(key)]
-    if len(supplied_tool_keys) == 1:
-        # Masking needs both markers, so one alone would silently do nothing — and it
-        # would also suppress the auto-detection that reports a nesting template.
-        missing = next(key for key in _tool_keys if key not in supplied_tool_keys)
-        raise ValueError(
-            f"collator_kwargs sets '{supplied_tool_keys[0]}' without '{missing}'. "
-            "Tool results are only excluded from the training loss when both markers "
-            f"are set.\nFix: also set '{missing}' in collator_kwargs."
-        )
-
-    effective_response = config_collator_kwargs.get(
-        "response_template", collator_kwargs.get("response_template")
+    tool_bracket = _resolve_tool_bracket(
+        tokenizer, collator_kwargs, config_collator_kwargs
     )
-    effective_eot = config_collator_kwargs.get(
-        "end_of_turn_template", collator_kwargs.get("end_of_turn_template")
-    )
-    if (
-        collator_kwargs.get("train_target")
-        in ("all_assistant_turns", "final_assistant_turn")
-        and effective_response
-        and effective_eot
-        and not supplied_tool_keys
-    ):
-        tool_bracket = resolve_tool_response_template(
-            tokenizer, effective_response, effective_eot
-        )
-        if tool_bracket is not None:
-            collator_kwargs["tool_response_template"] = tool_bracket[0]
-            collator_kwargs["end_of_tool_response_template"] = tool_bracket[1]
+    if tool_bracket is not None:
+        opener_key, closer_key = _TOOL_TEMPLATE_KEYS
+        collator_kwargs[opener_key], collator_kwargs[closer_key] = tool_bracket
 
     # User-provided collator_kwargs override auto-resolved values
     collator_kwargs.update(config_collator_kwargs)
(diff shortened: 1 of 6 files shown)


src/oumi/core/collators/trl_data_collator_for_completion_only_lm.py
```diff
@@ -92,6 +94,10 @@ def __init__(
         self.response_token_ids: list[int] = self._tokenize_template(response_template)  # type: ignore[assignment]
         self.end_of_turn_template = end_of_turn_template
         self.end_of_turn_token_ids = self._tokenize_template(end_of_turn_template)
+        self.tool_response_token_ids = self._tokenize_template(tool_response_template)
+        self.end_of_tool_response_token_ids = self._tokenize_template(
+            end_of_tool_response_template
+        )
```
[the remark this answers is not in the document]
shanghongsim: This should not be a problem. L408 in collator.py disallows users to only supply one (either tool_response or end_of_tool_response)

src/oumi/builders/collators.py
```diff
@@ -430,6 +622,31 @@ def build_collator_from_config(
...
+    effective_eot = config_collator_kwargs.get(
+        "end_of_turn_template", collator_kwargs.get("end_of_turn_template")
+    )
+    if (
+        collator_kwargs.get("train_target")
+        in ("all_assistant_turns", "final_assistant_turn")
+        and effective_response
+        and effective_eot
+        and not any(config_collator_kwargs.get(key) for key in _tool_keys)
+    ):
```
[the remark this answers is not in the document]
shanghongsim: Fixed

jgreer013 (approved):

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+            "name": "probe_fn",
+            "content": f"{_PROBE_RESULT}{i}",
+        }
+        for i in range(num_results)
+    ]
+    messages.append({"role": "assistant", "content": "PROBE_ANSWER"})
+    return messages
+
+
+def _diff_ids(shorter: list[int], longer: list[int]) -> list[int]:
```
wizeng23, line 262: Can this get directly unit-tested?

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+    """Token ids `longer` gained over `shorter`, stripping common prefix and suffix."""
```
wizeng23, line 263: Is prefix+suffix==len(shorter) the expectation in this function at the end? The wording of this seems to imply that. If that is the expectation, might be good to have a logger warning if it's not the case. If not , perhaps a more accurate description is "Token ids left in `longer` after stripping common prefix and suffix shared with `shorter`."

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+        )
+        return None
+
+    if not isinstance(one, str) or not isinstance(two, str):
+        return None
+
+    first_result = f"{_PROBE_RESULT}0"
+    # Template drops tool messages entirely; nothing to mask.
+    if first_result not in one:
+        return None
```
wizeng23, line 331: Can we add debug messages to the other early-exit points?

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+        )
+    except Exception as e:
+        # A template that rejects the probe tells us nothing about nesting.
+        logger.debug(
+            "Tool-response probe failed to render: %s: %s", type(e).__name__, e
+        )
+        return None
+
+    if not isinstance(one, str) or not isinstance(two, str):
+        return None
```
wizeng23, line 326: Should this ever trigger? If it's not ever expected, might be worth a ValueError

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+            "Chat template renders tool results inside assistant turns, but adding a "
+            "second tool result did not change the rendered output, so the tool-result "
+            "block could not be isolated and cannot be excluded from the training "
+            f"loss.\n{_TOOL_FIX_HINT}"
+        )
+
+    # Filter for marker tokens
+    # get_added_vocab/convert_ids_to_tokens are declared on the concrete tokenizer
+    # classes, not on the PreTrainedTokenizerBase alias; the cast is types-only.
+    vocab_tokenizer = cast(PreTrainedTokenizerFast, tokenizer)
```
wizeng23, line 353: Is there a reason to cast mid-function instead of declaring the tokenizer to be this type in the arguments? The type expectation for tokenizer gets a bit buried

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+        return None
+
+    first_result = f"{_PROBE_RESULT}0"
+    # Template drops tool messages entirely; nothing to mask.
+    if first_result not in one:
+        return None
+
+    # Nested when the closest preceding turn boundary is an assistant header.
+    head = one[: one.index(first_result)]
+    not_nested = head.rfind(response_template) <= head.rfind(end_of_turn_template)
```
wizeng23, line 335: Do we need to first do a check that neither rfind resulted in -1?

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+    while (
+        suffix < len(shorter) - prefix
+        and suffix < len(longer) - prefix
+        and shorter[len(shorter) - suffix - 1] == longer[len(longer) - suffix - 1]
+    ):
+        suffix += 1
+    return longer[prefix : len(longer) - suffix]
+
+
+def resolve_tool_response_template(
```
wizeng23, line 281: template generally has a different meaning in our codebase, and the function name signals nothing about its return type (tool bracket token ids). Could it be renamed to something more clear, ex. "detect_tool_response_bracket_token_ids"?

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+    Span masking unmasks everything between ``response_template`` and the next
+    ``end_of_turn_template``. Some chat templates (e.g. gemma-4) render tool results
+    inside the assistant turn resulting in environment output being unmasked.
+    The returned bracket lets the collator remask it.
+
+    Returns token ids rather than strings to avoid a decode/re-encode round trip
+    that isn't guaranteed to be lossless.
+
+    Returns:
+        (opener_ids, closer_ids) when tool results are nested and a specific bracket
```
wizeng23, line 297:
```suggestion
        ([opener_id], [closer_id]) when tool results are nested and a specific bracket
```

Could you make it more clear that only one of each is returned, so readers don't have to read the whole function to find out? I know the function return type is convenient for downstream ingestions, but as a reader it was confusing why it was returning a list when the function description makes it sound like it should only be returning two ints

tests/unit/core/collators/test_text_completions_collator_with_padding.py
```diff
@@ -637,3 +637,152 @@ def test_pad_to_multiple_of_none_keeps_longest_in_batch():
...
     batch = collator([row])
     assert batch["input_ids"].shape[1] == len(row["input_ids"])
+
+
+# ---------------------------------------------------------------------------
+# Tool results nested inside an assistant turn
+# ---------------------------------------------------------------------------
+
+# Markers standing in for a template that renders tool results inside the assistant
+# turn (gemma-4 renders <|tool_response|>...<tool_response|> there).
```
wizeng23, line 647:
```suggestion
# turn (gemma-4 renders <|tool_response>...<tool_response|> there).
```

src/oumi/core/collators/trl_data_collator_for_completion_only_lm.py
wizeng23, line 45: This is missing the new fields

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+def _resolve_tool_bracket(
+    tokenizer: "BaseTokenizer",
+    collator_kwargs: dict,
+    config_collator_kwargs: dict,
+) -> tuple[list[int], list[int]] | None:
+    """The bracket around tool results this template nests in the assistant turn.
+
+    Some templates render tool results inside the assistant turn, which leaves them
+    unmasked. The returned bracket lets the collator mask them again.
+
```
wizeng23, line 399: Could you have an Args: block in your docstring? It's not immediately clear how the kwargs would be used, and they're not mentioned in the docstring at all. Similar comment applies elsewhere

src/oumi/builders/collators.py
```diff
@@ -430,6 +677,13 @@ def build_collator_from_config(
                 "or provide response_template in collator_kwargs."
             )
 
+    tool_bracket = _resolve_tool_bracket(
```
wizeng23, line 680: Does this only apply to text_completions_only_with_padding? If so it should be moved in that if block. Otherwise, please confirm that this doesn't break jobs with other collators in E2E tests

tests/integration/builders/test_collator_template_detection.py
```diff
@@ -229,3 +244,231 @@ def test_template_detection_newer_transformers(
...
+
+@pytest.mark.parametrize(
+    "model_name,trust_remote_code",
+    [
+        pytest.param("google/gemma-4-E2B-it", False, id="gemma-4-nested"),
+        pytest.param("zai-org/GLM-4.5", False, id="glm-4.5-nested"),
+        pytest.param("Qwen/Qwen3-0.6B", False, id="qwen3-separate-turn"),
+    ],
+)
+@requires_hf_token()
```
wizeng23, line 442: What's the token needed for? Are any of these models gated

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+        )
+    if supplied:
+        return None  # Hand-supplied bracket wins; nothing to detect.
+
+    if collator_kwargs.get("train_target") not in _SPAN_TRAIN_TARGETS:
+        return None  # Only span-based masking can unmask a tool result.
+
+    response_template = config_collator_kwargs.get(
+        "response_template", collator_kwargs.get("response_template")
+    )
```
wizeng23, line 426:
```suggestion
    response_template = config_collator_kwargs.get(
        "response_template") or collator_kwargs.get("response_template")
```

To have consistent fallback behavior with turn_boundary

src/oumi/builders/collators.py
```diff
@@ -196,6 +222,227 @@ def resolve_collator_templates(
...
+            tokenize=False,
+            add_generation_prompt=False,
+        )
+        two = tokenizer.apply_chat_template(
+            _tool_probe_messages(2),
+            tools=_PROBE_TOOLS,
+            tokenize=False,
+            add_generation_prompt=False,
+        )
+    except Exception as e:
```
wizeng23, line 318: I agree that we should return None in a lot of these exit cases to not disrupt the majority of trainings, but in several cases, it's probably worth log.warning the user that something was wrong, instead of silently skipping. For example, the cases at lines 318, 335 (if both rfinds return -1), and 441  are probably worth a warning

wizeng23 (commented):
This PR is solving a pretty tough problem, and generalizing the approach to make it work across all models makes it even tougher.

1. However, we know which models are the offenders that need this PR, and don't need to probe them every time. WDYT about directly specifying the took bracket tokens in these cases so we can skip the probe? ex:

```
_TOOL_RESPONSE_MARKERS = (
    ("<|tool_response>", "<tool_response|>"),   # gemma-4
    ("<tool_response>", "</tool_response>"),    # GLM-4.5, Qwen3, Hermes-3
)
```

2. The blast radius for this change is quite large if it runs a probe for most jobs running collators. Could you make sure to confirm that other jobs aren't affected?

src/oumi/core/collators/text_completions_collator_with_padding.py
```diff
@@ -48,6 +50,11 @@ def __init__(
             or ``"final_assistant_turn"``.
         end_of_turn_template: String marking the end of a turn.
             Required for ``all_assistant_turns``.
+        tool_response_template: String or token IDs opening a tool result that the
+            chat template renders inside the assistant turn (e.g. gemma-4's
+            ``<|tool_response>``). Auto-detected by ``resolve_tool_response_template``.
```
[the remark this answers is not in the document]
shanghongsim: `<|tool_response` is correct

src/oumi/builders/collators.py
shanghongsim: fixed in follow up PR

tests/unit/core/collators/test_text_completions_collator_with_padding.py
shanghongsim: fixed

src/oumi/builders/collators.py
shanghongsim: fixed

src/oumi/builders/collators.py
shanghongsim: fixed

PR #2592, opened 2026-08-10
fix(verl): make the Countdown GRPO example runnable and checkpoint-producing

aniruddh-alt:
## What

Makes the verl Countdown example run, and makes a finished run leave a usable model behind.

## Why

The example currently fails before it generates a single sample, and a run that does complete
discards every weight it produced. Both were found by running the recipe end to end on 8xH100.

## How

Four changes, three of them one-liners in the example config:

- **Model.** The pinned model ships no chat template, and verl >= 0.7 rebuilds the tokenizer
  inside its rollout worker, so the fallback template Oumi installs never reaches it and the
  worker raises on startup. Switched to `Qwen/Qwen3-4B-Base`, which ships one, rather than
  carrying a hand-written template in the config. Qwen3 is the newest family the pinned stack
  loads — Qwen3.5 and Gemma 4 need transformers >= 5, which conflicts with verl 0.7.1's
  `trl<0.20`. Note newer does not imply templated: `gemma-3-4b-pt` and `Llama-3.2-3B` both ship
  without one.
- **`save_steps: -1` to `150`.** `-1` disables checkpointing entirely, final checkpoint
  included, so the example trained and threw the result away.
- **`max_actor_ckpt_to_keep: 3`.** verl retains every checkpoint by default, and each is ~47 GB
  at this model size — enough to fill a 500 GB volume partway through one epoch.
- **HF export.** verl training returned before the save step, leaving `_export_hf_model()`
  unreachable, so a finished run had only verl's FSDP shards and no loadable model. The export
  now runs at the end of training, gated on `save_final_model`. This regressed in #1749: verl
  0.4.0 dropped `hf_model` from the default checkpoint contents and the vendored config was
  refreshed wholesale.

## Testing

Full epoch on 8xH100 with the config as it ships here — 937 steps, 16h25m, exit 0. Held-out
accuracy on the 6000-row test split moved **0.041 to 0.651** (peak 0.663):

```
step     0    150    300    450    600    750    900    937
acc  0.041  0.565  0.603  0.606  0.628  0.632  0.663  0.651
```

Training ended and the HF export fired on its own, merging 8 FSDP shards into a loadable 8.3 GB
model in `output_dir` — on `main` that directory holds only `verl_output/`.

The chat-template failure was isolated by a three-arm comparison before the model swap: the old
config and the same config using Oumi's `model.chat_template` both raise identically inside the
rollout worker, while a verl-side template trains a step cleanly. That is what established the
fix had to be the model rather than a config knob.

Unit tests cover both export paths — that it is called when `save_final_model` is set, and that
a run which wrote no checkpoints returns cleanly instead of raising. Both fail on `main`. The
second exists because the first version of this change crashed a completed run with
`FileNotFoundError`: verl never creates its output directory when nothing was saved. Standard
lints clean.

## Note for reviewers

Two properties of this run are worth knowing before using these numbers as a baseline.

Entropy collapses early — 1.51 at step 2 to 0.074 by the end — so most of the gain lands in the
first 150 steps and the remaining 800 add ~9 points.

And the trained model is weak on the hard tiers for a behavioral reason rather than a budget
one. Per-tier greedy accuracy runs 1.00 / 0.99 / 0.91 / 0.58 / 0.36 / 0.15 across 2 to 7
numbers, and 32.5% of generations never emit an answer at all. Doubling the token budget to
2048 recovers only 3 points (0.663 to 0.693) and barely moves the timeout rate (32.5% to
29.2%), so the model is looping rather than running out of room.

### DIFF
--- a/configs/examples/grpo_verl_countdown/train.yaml
+++ b/configs/examples/grpo_verl_countdown/train.yaml
@@ -13,7 +13,7 @@
 #   - Other training configs: configs/**/*train.yaml
 
 model:
-  model_name: "d1shs0ap/cognitive-behaviors-Llama-3.2-3B"
+  model_name: "Qwen/Qwen3-4B"
 
 data:
   train:
@@ -28,7 +28,7 @@ data:
 training:
   trainer_type: "VERL_GRPO"
   num_train_epochs: 1
-  save_steps: -1
+  save_steps: 150
   eval_strategy: "steps"
   eval_steps: 50
 
@@ -71,6 +71,7 @@ training:
     trainer:
       critic_warmup: 0
       val_before_train: False
+      max_actor_ckpt_to_keep: 3   # verl's default is unbounded (~47GB per checkpoint)
       n_gpus_per_node: 2
       nnodes: 1
 
--- a/configs/examples/grpo_verl_countdown/train.yaml
+++ b/configs/examples/grpo_verl_countdown/train.yaml
@@ -13,7 +13,7 @@
 #   - Other training configs: configs/**/*train.yaml
 
 model:
-  model_name: "Qwen/Qwen3-4B"
+  model_name: "Qwen/Qwen3-4B-Instruct-2507"
 
 data:
   train:


wizeng23 (approved):

configs/examples/grpo_verl_countdown/train.yaml
```diff
@@ -13,7 +13,7 @@
 #   - Other training configs: configs/**/*train.yaml
 
 model:
-  model_name: "d1shs0ap/cognitive-behaviors-Llama-3.2-3B"
```
shanghongsim, line 16: I don't think we should suggest users to use base. Base models have not been instruction tuned and will unlikely produce good responses in conversational setting and thus get poor rewards

src/oumi/core/trainers/verl_grpo_trainer.py
```diff
@@ -589,8 +589,8 @@ def train(self) -> None:
         logger.info("Starting verl training...")
         self._verl_trainer.fit()
 
-    # TODO: OPE-1192 - Implement saving model/trainer state. verl training should
-    # already handle saving models, including the final checkpoint.
```
shanghongsim, line 593: split saving logic to different PR

src/oumi/train.py
```diff
@@ -268,13 +268,21 @@ def _verl_train(partial_trainer: Callable[[], BaseTrainer]):
     # decorator is only run if this function is run. This function should only be run
     # if ray is installed, preventing errors when it isn't.
     @ray.remote
-    def _run_verl_train(partial_trainer: Callable[[], BaseTrainer]):
```
shanghongsim, line 271: same thing here

tests/unit/test_train.py
```diff
@@ -208,3 +211,29 @@ def test_recompile_limit_env_override_wins(monkeypatch, restore_dynamo_limits):
         _config_with_collator_kwargs({"pad_to_multiple_of": 8192})
     )
     assert torch._dynamo.config.cache_size_limit == 128
+
+
+# Runs the remote function inline so no Ray cluster is needed. A MagicMock won't do:
+# it would swallow the call and the body would never execute.
+_INLINE_RAY = SimpleNamespace(
```
shanghongsim: Is this mocking ray?

configs/examples/grpo_verl_countdown/train.yaml
aniruddh-alt: agreed

tests/unit/test_train.py
aniruddh-alt: yes!

PR #2593, opened 2026-08-11
[Inference] Raise a clear error when a 2xx API response body parses to JSON null

rlehman221:
# Description

`RemoteInferenceEngine._convert_api_output_to_conversation` assumes the parsed response body is a JSON object. A provider can return an HTTP 2xx whose body parses to JSON `null` (or another non-object); `response.json()` succeeds, so the parse-failure retry branch is bypassed, and `if "error" in response:` then crashes with `TypeError: argument of type 'NoneType' is not iterable`. After retries this surfaces as the confusing `RuntimeError: Failed to process successful response: argument of type 'NoneType' is not iterable` (observed in production via Sentry, triggered by an OpenAI call).

This change guards the conversion with an `isinstance(response, dict)` check and raises a `RuntimeError` that states what actually happened (`Expected a JSON object in API response, got NoneType: None`). Retry behavior is unchanged: the existing handler in `_query_api` already catches conversion failures and retries up to `max_retries`, so transient null bodies are absorbed and persistent ones fail with an actionable message.

## Related issues

None — found while investigating a production Sentry alert.

## Before submitting

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/inference/remote_inference_engine.py
+++ b/src/oumi/inference/remote_inference_engine.py
@@ -573,6 +573,13 @@ class RemoteInferenceEngine(BaseInferenceEngine):
         Returns:
             Conversation: The conversation including the generated response.
         """
+        if not isinstance(response, dict):
+            # A 2xx response can still carry a body that parses to JSON `null`
+            # (or a non-object); surface it clearly instead of a TypeError below.
+            raise RuntimeError(
+                "Expected a JSON object in API response, got "
+                f"{type(response).__name__}: {str(response)[:200]}"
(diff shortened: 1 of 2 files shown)


wizeng23:
Do you have an example of a call where the body parses to null?

wizeng23 (approved):

rlehman221:
> Do you have an example of a call where the body parses to null?

I don't have an exact example since our logs didn't capture it but the errors I traced down was on this path and in theory this could happen again

PR #2595, opened 2026-08-11
Add muse_glimmer to the internal model type map

lefft:
MuseGlimmerForConditionalGeneration is registered only under `AutoModelForImageTextToText` in transformers — there is no `AutoModelForCausalLM` entry for `muse_glimmer` — so `build_model`'s default auto-class raises `ValueError: Unrecognized configuration class` for `meta-models/Muse-Glimmer-30B`, including for text-only SFT.

This adds the model type to `get_all_models_map` with the correct auto-class. `chat_template` is left empty so the tokenizer's built-in (ATEM) template is used rather than a named override.

Verified with a LoRA SFT run (TRL_SFT, completions-only collator, FSDP on `MuseGlimmerTextDecoderLayer`) that trains and saves cleanly.

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/configs/internal/supported_models.py
+++ b/src/oumi/core/configs/internal/supported_models.py
@@ -622,6 +622,15 @@ def get_all_models_map() -> Mapping[
             model_class=transformers.AutoModelForImageTextToText,
             config=_create_internvl_config(),
         ),
+        _ModelTypeInfo(
+            model_type="muse_glimmer",
+            model_class=transformers.AutoModelForImageTextToText,
+            # muse_glimmer has no AutoModelForCausalLM registration in
+            # transformers, so the class mapping is required even for
+            # text-only SFT. chat_template is left empty so the tokenizer's
+            # built-in template is used.
+            config=InternalModelConfig(chat_template=""),
+        ),
     ]
 
     # Make it immutable.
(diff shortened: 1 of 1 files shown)


src/oumi/core/configs/internal/supported_models.py
```diff
@@ -622,6 +622,11 @@ def get_all_models_map() -> Mapping[
             model_class=transformers.AutoModelForImageTextToText,
             config=_create_internvl_config(),
         ),
+        _ModelTypeInfo(
+            model_type="muse_glimmer",
+            model_class=transformers.AutoModelForImageTextToText,
+            config=InternalModelConfig(chat_template=""),
```
aniruddh-alt: nit: I saw that the file convention was to create a helper function rather than an inline InternalModelConfig(). Wanted to check if diverging from convention makes sense here.

aniruddh-alt (approved):
LGTM! This is the same patch I did for my onboarding run.

src/oumi/core/configs/internal/supported_models.py
lefft: Yeah valid point, the helper function is trivial in this case but will follow the convention and revise!

PR #2599, opened 2026-08-12
Fix tool-call arguments for HF chat templates

aniruddh-alt:
Tool-call `arguments` are stored in OpenAI wire format (a JSON string), but HF chat templates expect an object — and the Jinja sandbox has no `fromjson` filter, so a template receiving a string cannot convert it. This affects every tool-using model.

## Two-sided symptom

- **Ingest**: `FunctionCall(arguments={"city": "SF"})` raised `ValidationError`. Dict-shaped arguments are the HF chat-template convention and what tool datasets ship, so they could not be loaded at all.
- **Render**: string arguments render double-encoded — `"arguments": "{\"city\": \"SF\"}"`. Qwen accepts this silently and trains on the wrong target; templates that index into `arguments` raise outright.

## Fix

1. `field_validator(mode="before")` on `FunctionCall.arguments` encodes dict/list input to the stored string form.
2. New `Conversation.to_chat_template_dict()` emits `arguments` as an object, used at the four sites feeding `apply_chat_template` (`BaseSftDataset._conversation_to_hf_compatible_dict`, `BaseSftDataset._tokenize`, and the native + sglang `_apply_chat_template_impl`).
3. Those engines now also pass `tools=`. Without it, inference prompts the model with no tool schema and it invents function names — correct argument values, hallucinated names.

`to_dict()` is deliberately **unchanged**: `test_to_dict_byte_identical_to_dict_input_for_tool_round_trip` locks JSONL round-trip byte-identity, and the OpenAI wire path (`create_list_of_message_json_dicts`) is a separate serializer. All three pre-existing tool tests pass unmodified.

Arguments that don't parse as a JSON object are left as-is, preserving the documented tolerance for malformed provider output.

## Tests

10 new tests, including two that render through Jinja templates reproducing both real failure modes (`| tojson` and `.items()`), and one asserting `to_dict()` still emits the OpenAI string form.

## Related

Overlaps `shanghong/toolaware-collator-debug` (assistant-span detection for tool calls) — no file overlap, but same problem space; worth reconciling before merge.

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/analyze/base.py
+++ b/src/oumi/analyze/base.py
@@ -21,6 +21,7 @@ from typing import Any, ClassVar, Generic, TypeVar
 from pydantic import BaseModel
 from transformers import PreTrainedTokenizerBase
 
+from oumi.core.tokenizers import utils as tokenizer_utils
 from oumi.core.types.conversation import ContentItem, Conversation, Message
 
 TResult = TypeVar("TResult", bound=BaseModel)
@@ -236,8 +237,9 @@ class ConversationAnalyzer(BaseAnalyzer[TResult]):
                 "A chat template is required to format conversation text."
             )
 
-        result = tokenizer.apply_chat_template(
-            conversation=conversation,  # type: ignore
+        result = tokenizer_utils.apply_chat_template(
+            tokenizer,
+            conversation,
             tokenize=False,
             return_dict=False,
         )
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -22,6 +22,7 @@ from typing_extensions import override
 from oumi.core.datasets.base_map_dataset import BaseMapDataset
 from oumi.core.tokenizers import BaseTokenizer
 from oumi.core.tokenizers.utils import (
+    apply_chat_template,
     tokenize_for_completions_only_training_with_prefix,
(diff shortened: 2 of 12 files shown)


aniruddh-alt:
Closing in favour of @shanghong's series, which covers the same surface with a better approach: #2601 (argument conversion on `FunctionCall`) and #2605 (detecting the tool-call form a chat template accepts, rather than rendering-then-retrying on exception).

Two findings from reviewing this branch that may be worth carrying over:

1. The processor/VLM path is untouched by any of this. `BaseProcessor.apply_chat_template` takes no `tools` parameter, and `DefaultProcessor._convert_messages_to_dicts` emits `arguments` via `model_dump`, so it is always the string form. A VLM agent with an image + tool definitions never sees the tool schema and hits the same double-encoding. Call sites: `native_text_inference_engine.py:190`, `sglang_inference_engine.py:200`, `vision_language_conversation_feature_generator.py:258`.
2. `train_target: FINAL_ASSISTANT_TURN` silently under-supervises multi-tool-call data. Templates that render each tool call as its own assistant span (Muse Glimmer does) get only the last call trained — measured 1 of 3 calls, 117 supervised tokens vs 356 under `ALL_ASSISTANT_TURNS`. Loss and token-accuracy look healthy throughout; the tell is the model emitting exactly one call per row.

PR #2600, opened 2026-08-13
fix(collators): clamp auto-detected response template to the generati…

lefft:
# Description

Some chat templates render assistant headers that extend past the generation prompt — harmony-style templates (e.g. `meta-models/Muse-Glimmer-30B`, gpt-oss) render assistant turns as `<|start|>assistant to=user<|message|>…` while the generation prompt is the bare `<|start|>assistant`. When collator templates are auto-detected (`train_target` set, no explicit `response_template`), detection returns the full header, so the trailing ` to=user<|message|>` span is masked out of the loss even though the model must generate it at inference. Heavy fine-tuning corrupts that untrained span: in our Muse-Glimmer FFT runs, the model emitted `' to51'` instead of ` to=user<|message|>51`, leaving 405/500 eval rows unparseable.

**Fix:** after detection, clamp `response_template` to the generation-prompt suffix whenever that suffix is a strict prefix of the detected header — the header remainder then stays in the trained span. The invariant enforced: the masking boundary equals the point where generation begins at inference.

**Blast radius:** none for standard templates. The clamp only fires when the rendered assistant header is strictly longer than the generation prompt; for every template where the two coincide (the normal case) it is a no-op by construction. Verified empirically: all 11 real-tokenizer detection tests (Qwen 2.5/3/3.5/3.6, DeepSeek-R1 ×2, Olmo3, SmolLM 2/3, MiniMax) produce byte-identical results with and without the change. Explicitly-configured `response_template`s are untouched — only auto-detection output changes. New unit test covers the harmony-style case; `meta-models/Muse-Glimmer-30B` added to the integration detection matrix.

**Before/after** (`resolve_collator_templates` on real tokenizers; auto-detection path):

```python
# meta-models/Muse-Glimmer-30B — clamped
before: ('<|start|>assistant to=user<|message|>', '<|eot|>')
after:  ('<|start|>assistant', '<|eot|>')
# trained span before:  '54<|eot|>'                      (header masked — corrupts under FFT)
# trained span after:   ' to=user<|message|>54<|eot|>'   (header supervised)

# google/gemma-4-12B-it — unchanged
before == after: ('<|turn>model', '<turn|>\n')

# Qwen/Qwen3.5-0.8B — unchanged
before == after: ('<|im_start|>assistant', '<|im_end|>\n')
```

Validated end-to-end on the enterprise platform: Muse-Glimmer FFT trained with auto-detected (clamped) templates matches the explicit-template run on banking77 (0.862 vs 0.868 accuracy, 500/500 parseable outputs both).

## Related issues

Supports Muse-Glimmer-30B SFT onboarding (oumi-ai/api#6146).

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/builders/collators.py
+++ b/src/oumi/builders/collators.py
@@ -227,6 +227,90 @@ def _detect_response_template(
     return response_template
 
 
+def _clamp_to_generation_prompt(
+    tokenizer: "BaseTokenizer",
+    response_template: str,
+    msgs: list[dict[str, str]],
+) -> str:
+    r"""Clamp the response template to the generation-prompt boundary.
+
+    Some chat templates render assistant headers that extend past the
+    generation prompt: e.g. harmony-style templates render assistant turns
+    as ``<|start|>assistant to=user<|message|>…`` while the generation
+    prompt is the bare ``<|start|>assistant``. Masking through the full
+    header leaves the trailing span (`` to=user<|message|>``) untrained,
+    even though inference expects the model to generate it — heavy
+    fine-tuning then corrupts that span. When the generation-prompt suffix
+    is a strict prefix of the detected header, prefer it so the rest of the
+    header is trained.
+
+    Args:
+        tokenizer: The tokenizer whose chat template is being probed.
+        response_template: The auto-detected assistant header.
+        msgs: A sentinel conversation ending with a user turn.
+
+    Returns:
+        The clamped response template, or the original when the
+        generation prompt is unavailable or is not a strict prefix.
+
+    Examples:
+        Muse Glimmer renders assistant headers past the generation prompt, so
+        the suffix is a proper prefix of the header and the clamp fires — the
+        span between the two boundaries (`` to=user<|message|>``) moves into
+        the trained region::
+
+            detected    '<|start|>assistant to=user<|message|>'
+            gen_suffix  '<|start|>assistant'
+            -> '<|start|>assistant'
+
+        Qwen2.5's generation prompt IS its assistant header (modulo the
+        trailing newline both sides strip), so the suffix is not strictly
+        shorter and nothing changes — the no-op case covering standard
+        templates::
+
+            detected    '<|im_start|>assistant'
+            gen_suffix  '<|im_start|>assistant\n'  ->  '<|im_start|>assistant'
+            -> unchanged
+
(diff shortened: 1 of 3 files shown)


src/oumi/builders/collators.py
```diff
@@ -302,6 +358,9 @@ def resolve_collator_templates(
         header_text=rendered[second_user_end:second_asst],
         eot_ids=eot_ids,
     )
+    response_template = _clamp_to_generation_prompt(
+        tokenizer, response_template, msgs[:-1]
```
lefft: Here we are referencing `msgs` from the terminated loop above. Works, but somewhat confusing and could be problematic if the loop above is edited in the future and the author doesn't catch this reference. 

Would be cleaner if in the loop above we add e.g. `chosen_msgs = msgs` like:

```py
    for msgs in (msgs_with_sys, msgs_no_sys):
        try:
            rendered = tokenizer.apply_chat_template(
                msgs, tokenize=False, add_generation_prompt=False
            )
            chosen_msgs = msgs
            break
```

and then refer to `chosen_msgs` here.

lefft: addressed

src/oumi/builders/collators.py
```diff
@@ -302,6 +360,9 @@ def resolve_collator_templates(
         header_text=rendered[second_user_end:second_asst],
         eot_ids=eot_ids,
     )
+    response_template = _clamp_to_generation_prompt(
```
shanghongsim, line 392: The clamp approach is clever! I think we could use this as the general approach to detect response prefix `_detect_response_template`.

shanghongsim: It is also unclear when `_clamp_to_generation_prompt` will override `_detect_response_template` with a different result in the current code structure

lefft: Ya fair, to make it clearer, we can explicitly check whether the gen prompt suffix is a proper prefix of `response_template` (as given by `_detect_response_template()`), and then apply the clamp depending on the result of that check. 

Will also add some examples to the docstring similar to what we have for `_detect_eot_template()`.

shanghongsim: sounds like a reasonable approach for now. will approve to unblock.

shanghongsim (approved):

PR #2601, opened 2026-08-13
Add methods to convert tool arguments between json string and dictionary

shanghongsim:
# Description

`FunctionCall.arguments` is typed `str` to match the OpenAI wire format, but some datasets (such as those from real enterprise use cases) carry the arguments as a JSON object. Loading such a file raised a validation error, so every caller had to pre-serialize to dict to json string.

Decisions:
1. Keep string as the single storage contract rather than dict because other workflows like synthesis or inference may return malformed json, which will raise if we store as dict. We want to be able to store the malformed json for error logging purposes and re-feed into the system for the LLM to correct in next turn.
2. Add method coerce dict/list to a JSON string in a `mode="before"` validator to enable us to take in datasets carry the arguments as a JSON object.
3. Add `parsed_arguments()` so consumers that need a mapping decode it in one place with one error message instead of reimplementing `json.loads` and its failure handling.

The Anthropic engine, which needs a parsed `input` dict, switches to `parsed_arguments()`.

Other misc edits: narrow two `resp.text` assertions in the verl tool tests to satisfy pyright

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/core/types/tool_call.py
+++ b/src/oumi/core/types/tool_call.py
@@ -160,7 +160,7 @@ class FunctionCall(BaseModel):
 
     @pydantic.field_validator("arguments", mode="before")
     @classmethod
-    def _coerce_arguments(cls, raw: Any) -> str:
+    def _coerce_arguments_to_str(cls, raw: Any) -> str:
         """Normalize dict/list arguments to a JSON string.
 
         Runs in ``mode="before"`` so we can intercept non-string inputs
@@ -175,7 +175,7 @@ class FunctionCall(BaseModel):
             f"arguments must be a str, dict, or list; got {type(raw).__name__}."
         )
 
-    def parsed_arguments(self) -> dict[str, Any]:
+    def get_arguments_dict(self) -> dict[str, Any]:
         """Returns ``arguments`` decoded from its JSON string form.
 
         The string is the storage contract; every consumer that needs a dict
--- a/src/oumi/core/types/tool_call.py
+++ b/src/oumi/core/types/tool_call.py
@@ -25,9 +25,11 @@ keys at validation time would silently lose information that
 downstream code relies on.
 """
 
+import json
 from enum import Enum
 from typing import Any, Literal
 
+import pydantic
 from pydantic import BaseModel, ConfigDict, JsonValue
 
 # The seven primitive types defined by JSON Schema.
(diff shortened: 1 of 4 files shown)


jgreer013 (approved):

src/oumi/core/types/tool_call.py
```diff
@@ -156,6 +158,45 @@ class FunctionCall(BaseModel):
...
+        """
+        if isinstance(raw, str):
+            return raw
+        if isinstance(raw, (dict, list)):
+            return json.dumps(raw)
+        raise ValueError(
+            f"arguments must be a str, dict, or list; got {type(raw).__name__}."
+        )
+
+    def parsed_arguments(self) -> dict[str, Any]:
```
wizeng23: nit: get_arguments_dict as function name? IMO more self-documenting

wizeng23 (approved):

wizeng23:
Taking a step back, I'm wondering if we do want to keep string as the storage contract. IMO we want to flag malformed JSON as early as possible, so ideal behavior would be flagging them as failed examples at the end of a synthesis/eval job, instead of the job succeeding and silently having malformed output.

Do you know if passing a str or dict is the norm? ex. for other remote inference providers like Openrouter

shanghongsim:
Seems like both Openrouter and fireworks store their function call arguments as string

Openrouter (https://openrouter.ai/docs/guides/features/tool-calling)

```
{
  "model": "google/gemini-3-flash-preview",
  "messages": [
    {
      "role": "user",
      "content": "What are the titles of some James Joyce books?"
    },
    {
      "role": "assistant",
      "content": null,
      "tool_calls": [
        {
          "id": "call_abc123",
          "type": "function",
          "function": {
            "name": "search_gutenberg_books",
            "arguments": "{\"search_terms\": [\"James\", \"Joyce\"]}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "tool_call_id": "call_abc123",
      "content": "[{\"id\": 4300, \"title\": \"Ulysses\", \"authors\": [{\"name\": \"Joyce, James\"}]}]"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "search_gutenberg_books",
        "description": "Search for books in the Project Gutenberg library",
        "parameters": {
          "type": "object",
          "properties": {
            "search_terms": {
              "type": "array",
              "items": {"type": "string"},
              "description": "List of search terms to find books"
            }
          },
          "required": ["search_terms"]
        }
      }
    }
  ]
}
```

Fireworks (https://docs.fireworks.ai/guides/function-calling)

```
tools = [{
    "type": "function",
    "function": {
        "name": "get_weather",
        "description": "Get the current weather for a location",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {"type": "string", "description": "City name"}
            },
            "required": ["location"]
        }
    }
}]

response = client.chat.completions.create(
    model="accounts/fireworks/models/kimi-k2-instruct-0905",
    messages=[{"role": "user", "content": "What's the weather in San Francisco?"}],
    tools=tools,
    temperature=0.1
)

print(response.choices[0].message.tool_calls)
# Output: [ChatCompletionMessageToolCall(id='call_abc123', function=Function(arguments='{"location":"San Francisco"}', name='get_weather'), type='function')]
```

I guess in theory, there are a few options for whose responsibility it is to validate the json:
1. The storage layer
2. The tool executor or consumer eg training

I am more for option 2 because validation is decoupled from storing. 

> IMO we want to flag malformed JSON as early as possible, so ideal behavior would be flagging them as failed examples at the end of a synthesis/eval job, instead of the job succeeding and silently having malformed output.

I think we can still store as string and run the validation step right after the json is produced to catch malformed output. Then we can re-feed the malformed output and the error message to the model to do the next inference step. to my understanding, storing as dict means enforced a almost structured decoding like behavior because it errors on malformed json. We could possible still pass the error back to the model via `{ role : "tool", content: "[malformed_json_here_in_str] [error_msg]"} ` but handling the malformed json in the previous assistant turn that produced the tool call will be tricky. I think string is the cleaner solution.

src/oumi/core/types/tool_call.py
shanghongsim: changed!

PR #2602, opened 2026-08-13
Add the canonical tool-call conversation

shanghongsim:
# Description

Creating a centralized canonical conversation to be used as a probe in source code and also as a test fixture

A lot of code surfaces rely on the idea of a canonical tool based conversation: 
- (upcoming) collator response and eot prefix detection (current version does not but an upgraded version that respect the encode and do not decode to reencode rule will)
- (upcoming) Tool prefix detection 
- various test files (in current and upcoming PRs)

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low


jgreer013 (approved):

wizeng23:
Could we put the file somewhere else? The `types/` dir doesn't seem like the right place. `src/oumi/utils/` is an option, though it's a bit crowded. If you have other suggestions based on where the future callers will be, feel free to suggest!

shanghongsim:
> Could we put the file somewhere else? The `types/` dir doesn't seem like the right place. `src/oumi/utils/` is an option, though it's a bit crowded. If you have other suggestions based on where the future callers will be, feel free to suggest!

This makes sense, I will shift to `src/oumi/utils/`!

PR #2603, opened 2026-08-13
Ensure tool call args are in correct type for chat template

shanghongsim:
# Description

HuggingFace chat templates do not agree on the shape of a tool call. Some iterate `function.arguments` as a mapping and raise on a string (eg Gemma 4), others concatenate it as a string and raise on a mapping (eg Deepseek). 

`content` is the same story: some stringify `None` into the text they train on, others drop the whole tool call when it is `""`.

`Conversation.to_dict()` only emit tool calls as string, so a caller that needed tool calls in a different format need to adapt it themselves. `create_chat_template_inputs(conversation, *, tool_arguments_format, content_format)` takes an Oumi Conversation and returns (messages, tools). Specifically, it also converts the tool call arguments to the format that the caller expects (string or dict types). Malformed JSON arguments raise with the offending message index.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: low

### DIFF
--- a/src/oumi/utils/conversation_utils.py
+++ b/src/oumi/utils/conversation_utils.py
@@ -13,12 +13,13 @@
 # limitations under the License.
 
 import base64
-from typing import Any
+from typing import Any, Literal
 
 import PIL.Image
 
 from oumi.core.tokenizers.base_tokenizer import BaseTokenizer
 from oumi.core.types.conversation import ContentItem, Conversation, Message, Type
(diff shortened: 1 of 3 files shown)


jgreer013 (approved):

wizeng23:
Could you add a description in the PR for how this function is expected to be used?

wizeng23 (approved):

PR #2605, opened 2026-08-13
Detect the tool-call form a chat template accepts

shanghongsim:
# Description

Templates disagree irreconcilably on how a tool call must be handed to `apply_chat_template`, and they disagree silently. 
- Qwen3.5 and gemma-4 raise on a JSON-string `arguments` while DeepSeek raises on a mapping. 
- Qwen2.5 accepts a string and pipes it through `tojson` with no type check, so the model trains on `"{\"city\": \"...\"}"` where an object belongs and does not raise. 
- GLM-4.5 renders `content: None` as the literal word "None" before every tool call while DeepSeek drops the call entirely when `content` is `""`.

The difference are all encoded in the model's chat template defined in jinja. The most general way to detect the differences is via a probe based approach. For each model chat template:
1. Render the canonical tool conversation across the four (arguments x content) combinations 
2. Take the first option where the argument sentinel survives verbatim and unescaped. 

Notes:
- A template that raises rejects that form outright; one that drops the call, or double-encodes it so the sentinel arrives escaped, is rejected too. 
- Order  prefers `("mapping", "empty")`, which is what keeps GLM's stray "None" out of the data. Both content forms render there, so the probe cannot separate them and the preference has to.

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

> [!NOTE]
> **Liberate**
> Risk: medium

### DIFF
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -249,16 +249,11 @@ class BaseSftDataset(BaseMapDataset, ABC):
                 instruction_token_ids=self.instruction_token_ids,
             )
 
-    def _tokenize(
-        self, sample: dict | pd.Series | Conversation, tokenize: bool = True
-    ) -> dict:
+    def _tokenize(self, sample: Conversation, tokenize: bool = True) -> dict:
         if self._tokenizer is None:
             raise ValueError("Tokenizer is required for tokenization.")
 
-        chat_input: dict | pd.Series | Conversation | list[dict] = sample
-        tools = None
-        if isinstance(sample, Conversation):
-            chat_input, tools = apply_chat_template_inputs(self._tokenizer, sample)
+        chat_input, tools = apply_chat_template_inputs(self._tokenizer, sample)
 
         results = self._tokenizer.apply_chat_template(
             chat_input,  # type: ignore
--- a/src/oumi/core/datasets/base_sft_dataset.py
+++ b/src/oumi/core/datasets/base_sft_dataset.py
@@ -22,6 +22,7 @@ from typing_extensions import override
 from oumi.core.datasets.base_map_dataset import BaseMapDataset
 from oumi.core.tokenizers import BaseTokenizer
 from oumi.core.tokenizers.utils import (
+    apply_chat_template_inputs,
     tokenize_for_completions_only_training_with_prefix,
     tokenize_for_completions_only_training_with_template,
 )
@@ -254,14 +255,20 @@ class BaseSftDataset(BaseMapDataset, ABC):
         if self._tokenizer is None:
             raise ValueError("Tokenizer is required for tokenization.")
 
+        chat_input: dict | pd.Series | Conversation | list[dict] = sample
+        tools = None
+        if isinstance(sample, Conversation):
+            chat_input, tools = apply_chat_template_inputs(self._tokenizer, sample)
+
         results = self._tokenizer.apply_chat_template(
-            sample,  # type: ignore
+            chat_input,  # type: ignore
             tokenize=tokenize,
             return_dict=tokenize,
             return_tensors=self._return_tensors,
             max_length=self._tokenizer.model_max_length,
             truncation=True,
             add_generation_prompt=(self.task == "generation"),
+            tools=tools,  # type: ignore[arg-type]
         )
 
         if tokenize:
--- a/src/oumi/core/tokenizers/utils.py
+++ b/src/oumi/core/tokenizers/utils.py
@@ -21,7 +21,7 @@ import transformers
 
 from oumi.core.constants import LABEL_IGNORE_INDEX
 from oumi.core.tokenizers.base_tokenizer import BaseTokenizer
-from oumi.core.types import Conversation
+from oumi.core.types import Conversation, Role
 from oumi.utils.canonical_tool_conversations import (
     ARGUMENT_SENTINEL,
     canonical_tool_conversation,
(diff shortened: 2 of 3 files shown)


jgreer013 (approved):

src/oumi/core/tokenizers/utils.py
```diff
@@ -13,27 +13,180 @@
...
+    """
+    for candidate in _CANDIDATE_FORMATS:
+        if _probe_renders_sentinel(
+            tokenizer,
+            arguments_form=candidate.arguments,
+            content_form=candidate.content,
+        ):
+            return candidate
+
+    logger.warning(
```
lefft, line 161: IIUC, hitting this point would effectively mean that the resulting model would not produce tool calls as they are represented in the training data (or would not produce them at all?) 

Probably a non-issue if the use case doesn't involve tool calls, but def problematic if it does... 

Wonder if we should do one of the following:

1. Raise this warning every time a tool-use record is actually encountered dataprep/training -- i.e. make sure this is a loud warning; and/or 
2. dump more specific info in the warning, like the model name, an example record, and/or the chat template; or just  
3. fail the training job if the data includes tool use? (maybe too extreme) 

Other than that LGTM!

lefft (approved):

src/oumi/core/tokenizers/utils.py
```diff
@@ -13,27 +13,180 @@
...
+)
+from oumi.utils.conversation_utils import create_chat_template_inputs
 from oumi.utils.logging import logger
 
 
+class ChatTemplateToolFormat(NamedTuple):
+    """The form a chat template expects tool-call data in."""
+
+    arguments: Literal["mapping", "string"]
+    content: Literal["null", "empty"]
```
wizeng23, line 44: Please add docstrings describing these fields and their values

src/oumi/core/tokenizers/utils.py
```diff
@@ -13,27 +13,180 @@
...
+    # An escaped quote beside the value means the template JSON-encoded
+    # arguments that were already a JSON string, so the model would train on
+    # `"{\"city\": \"<sentinel>\"}"` where an object belongs.
+    if f'\\"{ARGUMENT_SENTINEL}' in rendered or f'{ARGUMENT_SENTINEL}\\"' in rendered:
+        return False
+
+    return True
+
+
+@functools.cache
```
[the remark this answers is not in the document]
wizeng23: +1. Is there a way to run the probe once, ex. at dataset init time, save the ChatTemplateToolFormat to the dataset class, then pass that in at tokenization time?

wizeng23:
### Code review

Found 1 issue:

1. `apply_chat_template_inputs` gates format detection on `conversation.tools`, but tool calls live on messages independently. A conversation with assistant `Message.tool_calls` but no top-level `conversation.tools` short-circuits at `if not conversation.tools: return conversation, None`, so it never reaches `create_chat_template_inputs`. That transformation is driven per-message off `msg_dict.get("tool_calls")` (decoding `function.arguments` to mapping form, normalizing `None` content) with no dependence on `conversation.tools` — so skipping it leaves arguments in JSON-string form and content as `None`, reproducing the exact double-encoding / literal-`None` corruption this PR is meant to prevent. Consider gating on whether any message carries `tool_calls` too, not just `conversation.tools`.

https://github.com/oumi-ai/oumi/blob/18c4d44599fa1c0f3a743f7edfe8fd177a620241/src/oumi/core/tokenizers/utils.py#L162-L172

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/oumi/core/tokenizers/utils.py
```diff
@@ -13,27 +13,180 @@
...
+
+def apply_chat_template_inputs(
+    tokenizer: BaseTokenizer, conversation: Conversation
+) -> tuple[Conversation | list[dict], list[dict] | None]:
+    """Returns the ``(conversation, tools)`` pair to hand ``apply_chat_template``.
+
+    A conversation without tools is returned untouched, so that path is
+    unchanged and never pays for a probe.
+    """
+    if not conversation.tools:
```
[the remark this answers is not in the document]
shanghongsim: Fixed

src/oumi/core/tokenizers/utils.py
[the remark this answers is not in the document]
shanghongsim: Tokenizer does not change during process lifetime and chat template cannot be mutated during training time

src/oumi/core/datasets/base_sft_dataset.py
```diff
@@ -254,14 +255,20 @@ def _tokenize(
         if self._tokenizer is None:
             raise ValueError("Tokenizer is required for tokenization.")
 
+        chat_input: dict | pd.Series | Conversation | list[dict] = sample
+        tools = None
+        if isinstance(sample, Conversation):
```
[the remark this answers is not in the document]
shanghongsim: Fixed redundant check!

src/oumi/core/datasets/base_sft_dataset.py
```diff
@@ -254,14 +255,20 @@ def _tokenize(
...
         results = self._tokenizer.apply_chat_template(
-            sample,  # type: ignore
+            chat_input,  # type: ignore
             tokenize=tokenize,
             return_dict=tokenize,
             return_tensors=self._return_tensors,
             max_length=self._tokenizer.model_max_length,
             truncation=True,
             add_generation_prompt=(self.task == "generation"),
+            tools=tools,  # type: ignore[arg-type]
```
wizeng23, line 266: Did we not pass in available tools here before?

src/oumi/core/tokenizers/utils.py
shanghongsim: Incorporated suggestions 1 and 2!

Re 1 -> if tool definitions or tool calls are encountered and tool call format cannot be resolved, the warning will flag
Re 2-> I added `_describe_fallback_rendering` to give more detailed debug logs

src/oumi/core/tokenizers/utils.py
shanghongsim: Fixed!

src/oumi/core/datasets/base_sft_dataset.py
shanghongsim: Not previously

PR #2617, opened 2026-08-18
Fix parsed inference content without tool calls

lefft:
## Summary

- Use a vLLM tool parser’s parsed content even when no tool call was detected.
- Preserve the existing tool-call payload, finish-reason, and parser-error fallback behavior.
- Convert parser-returned `None` content to an empty assistant message when there is no tool call.

This lets parsers that own model-specific response framing, such as `muse_glimmer`, remove protocol markers from ordinary assistant responses before Oumi serializes them. Previously Oumi only used parser output when `tools_called=True` and otherwise kept the raw model text. This also matches vLLM’s OpenAI serving path, which uses parser-returned content independently of whether a tool call was found.

## Reasoning behavior

Muse Glimmer natively emits private `to=self` reasoning before its user-visible `to=user` answer. The `muse_glimmer` tool parser returns only the latter as parsed content, so with this change the model still generates its reasoning but Oumi serializes only the clean final answer; the reasoning and its ATEM protocol markers are discarded.

Preserving that reasoning separately would require local vLLM reasoning-parser integration and consistent end-to-end handling of `Message.reasoning_content` by downstream consumers. That should be addressed as a broader reasoning-output feature rather than placing private reasoning or model-specific protocol tokens back into ordinary assistant `content`.

## Compatibility check

Reviewed representative non-tool response paths in the current vLLM parsers:

- **Qwen3.5 (`hermes`)**: returns the original model output as `content` when no `<tool_call>` tag is present, so ordinary answers are unchanged.
- **Llama 3.1–3.3 (`llama3_json`)**: returns the original model output as `content` when no tool-call payload is found, so ordinary answers are unchanged.
- **Muse Glimmer (`muse_glimmer`)**: returns only the visible user-channel content, or `None` when the response has no user-visible content; this is the behavior the fix is intended to preserve.
- Models that do not configure `tool_call_parser` are unaffected.

For both Qwen and Llama parsers, valid tool calls continue through the existing `tools_called=True` path; parsed content, tool-call payloads, and finish reasons are unchanged.

The remaining compatibility risk is limited to a custom parser that incorrectly returns `content=None` for a response that actually has visible content.

## Testing

- `ruff format` and `ruff check` on changed files
- focused parser-content unit tests (`2 passed`)
- targeted `pyright` on changed files (`0 errors`)

### DIFF
--- a/src/oumi/inference/vllm_inference_engine.py
+++ b/src/oumi/inference/vllm_inference_engine.py
@@ -668,8 +668,7 @@ class VLLMInferenceEngine(BaseInferenceEngine):
     ) -> tuple[list[Message], FinishReason | None]:
         """Builds assistant messages from a vLLM chat response.
 
-        When ``self._tool_parser`` is set, runs it over each completion's
-        text and populates ``Message.tool_calls`` with the extracted payload.
+        When ``self._tool_parser`` is set, uses its parsed content and tool calls.
         On parser failure, falls back to raw text.
         """
         new_messages: list[Message] = []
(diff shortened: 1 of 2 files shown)


wizeng23 (approved):

shanghongsim:
LGTM! Do you have some failure examples to illustrate the issue with muse glimmer? So I can visualize better 😅

lefft:
> LGTM! Do you have some failure examples to illustrate the issue with muse glimmer? So I can visualize better 😅

@shanghongsim sure -- here's an example from the banking77 task, which instructs the model to produce just an int label. 

Before this PR, an eval job saves something like this as the inference response: 
```
to=self<|message|>
... long messy thinking trace blah blah ... Okay I conclude the answer is 28.
<|eom|><|start|>assistant to=user<|message|>28
```

After this PR, the saved response is just:
```
28
```

The Muse parser was already extracting 28, but Oumi ignored parsed content unless a tool call was detected.

The model still emits reasoning tokens and ATEM markers (e.g. `to=user<|message|>`) but they're no longer retained in assistant content or inference output files. 

Like we discussed yesterday, longer-term we should have a general mechanism for parsing and retaining traces into a separate `reasoning_content` field -- this will require dedicated parsers/special handling for each base model family/chat template style, as well as consistent handling of this field downstream (e.g. in enterprise).

PR #2618, opened 2026-08-19
Fix FSDP merge handling of dim 0 tensors

shanghongsim:
# Description

GRPO training run launched in enterprise with Gemma 4 e2b trained successfully but failed at saving stage with the following error:

```
_run_verl_train
    _save_final_model(trainer, config)
  File "/usr/local/lib/python3.11/site-packages/oumi/train.py", line 254, in 
_save_final_model
    trainer.save_model(config=config)
  File 
"/usr/local/lib/python3.11/site-packages/oumi/core/trainers/verl_grpo_trainer.py
", line 621, in save_model
    self._export_hf_model()
  File 
"/usr/local/lib/python3.11/site-packages/oumi/core/trainers/verl_grpo_trainer.py
", line 676, in _export_hf_model
    merger.merge_and_save()
  File 
"/usr/local/lib/python3.11/site-packages/oumi/utils/verl_model_merger.py", line 
369, in merge_and_save
    merged_state_dict = self._load_and_merge_state_dicts(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File 
"/usr/local/lib/python3.11/site-packages/oumi/utils/verl_model_merger.py", line 
349, in _load_and_merge_state_dicts
    state_dict = torch.cat(state_dict, dim=0)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
```

The merging mechanism assumes every plain (non-DTensor) tensor in the rank checkpoints is dim-0-sharded and thus concatenates the pieces rankwise. Gemma-4 has parameters that are not stored as dim-0-sharded tensors and are too small to be sharded. FSDP replicates them across ranks. This causes two problems: 

- 0-dim scalars crash `torch.cat` -> results in the error
- `[1]`-shaped ones concatenate "successfully" into garbage (eg 35 `layer_scalar` params) -> silent error

This PR aims to handle the edge case of non-dim-0-sharded tensors via the following approach:
1. Build the target model architecture on the meta device and read the expected shape of every tensor in its state dict.
2. For each plain (non-DTensor) checkpoint entry, compare against the expected shape:
     - rank-0 shard shape == expected -> the copies are replicas, keep rank 0's (warn if copies differ across ranks)
     - dim-0-concatenated shape == expected -> the copies are shards, concatenate
     - neither -> raise ValueError naming the tensor.
3. If no expected shape is available (architecture couldn't be instantiated, or the key isn't part of it, e.g. a custom head), fall back with a warning to a value-based heuristic: keep rank 0's copy if the tensor is 0-dim or all rank copies are identical, otherwise concatenate along dim 0 (the previous behavior).

Gemma 4 has 963 of such plain tensors (non D tensor)
 - 928 zero-dim buffers — the input_min / input_max / output_min / output_max activation-range trackers on gemma-4's "clipped linear" layers in the vision and audio towers.
  - 35 shape-[1] parameters — the per-layer layer_scalars.

Most models (like qwen/llama) don't have these so they don't encounter this saving issue

| model | state-dict keys | 0-dim | shape-[1] | persistent buffers | min ndim |
  |---|---|---|---|---|---|
  | Llama-3.2-3B *(this yaml's original model)* | 255 | 0 | 0 | 0 | 1 |
  | Qwen2.5-3B-Instruct | 435 | 0 | 0 | 0 | 1 |
  | gemma-4-E2B-it | 1,952 | **928** | **35** | **963** | 0 |

## Tests
1. Added unit test to showcase and validate `_merge_unsharded_shards` behavior
3. Ran `/oumi/configs/examples/grpo_verl_countdown/train.yaml` for 3 steps. Training and saving finished successfully 

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/utils/verl_model_merger.py
+++ b/src/oumi/utils/verl_model_merger.py
@@ -346,9 +346,103 @@ class FSDPModelMerger(BaseModelMerger):
                     # 2-D list, FSDP + TP
                     raise NotImplementedError("FSDP + TP is not supported yet")
             else:
-                state_dict[key] = torch.cat(state_dict[key], dim=0)
+                # Expected shapes are only needed for plain (non-DTensor) entries,
+                # which carry no placement metadata.
+                if any(key not in param_placements for key in state_dict):
+                    expected_shapes = self._get_expected_shapes()
+
+                state_dict[key] = self._merge_unsharded_shards(
+                    key, state_dict[key], expected_shapes
+                )
 
         return state_dict
+
+    def _get_expected_shapes(self) -> dict[str, torch.Size] | None:
+        """Returns the target architecture's tensor shapes, keyed by name.
+
+        The model is instantiated on the meta device (shape/dtype metadata
+        only, no weight storage), so this is cheap even for large models.
+        Returns None if the architecture cannot be instantiated.
+        """
+        try:
+            auto_model_class = self.get_transformers_auto_model_class()
+            with init_empty_weights():
+                model = auto_model_class.from_config(self.model_config)
+            return {key: value.shape for key, value in model.state_dict().items()}
+        except Exception as e:
+            logger.warning(
+                "Could not build the target model on the meta device to look up "
+                f"expected tensor shapes ({e}); falling back to heuristics for "
+                "non-DTensor checkpoint entries."
+            )
+            return None
+
+    def _merge_unsharded_shards(
+        self,
+        key: str,
+        shards: list[torch.Tensor],
+        expected_shapes: dict[str, torch.Size] | None,
+    ) -> torch.Tensor:
+        """Merges per-rank copies of a tensor saved without placement metadata.
+
+        FSDP saves the tensors it does not shard (e.g. 0-dim learned scalars
+        and buffers) as plain tensors, so every rank's checkpoint holds a full
+        replicated copy. Genuinely dim-0-sharded plain tensors also exist in
+        older checkpoint formats. The target model's expected shape decides
+        between the two; when it is unavailable, identical rank copies are
+        treated as replicated.
+        """
+        # Ground truth is unavailable in two distinct cases; both fall back to
+        # the value-based heuristic.
+        if expected_shapes is None:
+            # Case 1: the target model could not be instantiated at all.
+            # (_get_expected_shapes already warned once for the whole merge.)
+            return self._merge_shards_heuristically(shards)
+        if key not in expected_shapes:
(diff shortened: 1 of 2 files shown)


jgreer013 (approved):

src/oumi/utils/verl_model_merger.py
```diff
@@ -346,10 +346,124 @@ def process_one_shard(rank: int, model_state_dict_lst: list):
                     # 2-D list, FSDP + TP
                     raise NotImplementedError("FSDP + TP is not supported yet")
             else:
-                state_dict[key] = torch.cat(state_dict[key], dim=0)
+                # Expected shapes are only needed for plain (non-DTensor) entries,
+                # which carry no placement metadata.
+                if any(key not in param_placements for key in state_dict):
+                    expected_shapes = self._get_expected_shapes()
```
lefft: Are we potentially rebuilding the model for every value of  `sorted(state_dict)` here? Would it be possible to pre-build it once before the loop and then re-use the shapes within the loop?

lefft (approved):
LG, just worth checking the comment I left to see if this can be done much more efficiently(?)

src/oumi/utils/verl_model_merger.py
shanghongsim: Fixed!

kaisopos (approved):

src/oumi/utils/verl_model_merger.py
```diff
@@ -346,10 +352,132 @@ def process_one_shard(rank: int, model_state_dict_lst: list):
...
+        Raises:
+            ValueError: Neither one per-rank copy nor the dim-0 concatenation
+                matches the tensor's expected shape.
+        """
+        if self._is_replicated(key, shards, expected_shapes, checkpoint_has_dtensors):
+            if any(
+                s.shape != shards[0].shape or not torch.equal(s, shards[0])
+                for s in shards[1:]
+            ):
+                logger.warning(
```
wizeng23, line 429: Is warning correct, or should it be info?

wizeng23 (approved):
We should remember to carefully test each model architecture we add for RL support to ensure they work

src/oumi/utils/verl_model_merger.py
shanghongsim: Tensors differing across ranks when replication was detected is unexpected and I wanted to alert the user to that, so I was thinking warning is appropriate!

PR #2620, opened 2026-08-21
fix(inference): fold Anthropic cache tokens into prompt_tokens

wizeng23:
# Description

Anthropic excludes cache read/write from input_tokens, while every other inference engine counts cached tokens inside prompt_tokens. Fold cache read and creation tokens into the normalized prompt_tokens so it means total prompt tokens uniformly across engines, keeping cached_tokens and cache_creation_tokens as a separate breakdown.

## Related issues

Fixes [LOU-3658](https://linear.app/oumi/issue/LOU-3658)

## Before submitting

### DIFF
--- a/src/oumi/inference/anthropic_inference_engine.py
+++ b/src/oumi/inference/anthropic_inference_engine.py
@@ -457,9 +457,7 @@ class AnthropicInferenceEngine(RemoteInferenceEngine):
             return None
         cached_tokens = usage.get("cache_read_input_tokens", 0)
         cache_creation_tokens = usage.get("cache_creation_input_tokens", 0)
-        # Anthropic excludes cache read/write from input_tokens; every other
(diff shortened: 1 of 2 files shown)


src/oumi/inference/anthropic_inference_engine.py
```diff
@@ -455,18 +455,22 @@ def _extract_usage_from_response(
         usage = response.get("usage")
         if not usage:
             return None
-        prompt_tokens = usage.get("input_tokens", 0)
+        cached_tokens = usage.get("cache_read_input_tokens", 0)
+        cache_creation_tokens = usage.get("cache_creation_input_tokens", 0)
+        # Anthropic excludes cache read/write from input_tokens; every other
```
lefft: If this is useful to remember in the future, perhaps something more concise like this? 

```
Anthropic excludes cache read/write from input, unlike other providers
```

lefft (approved):

PR #2625, opened 2026-08-26
feat(environments): add EndpointEnvironment for HTTP tool calls

min-oumi:
## Summary

Add `EndpointEnvironment`, registered as `endpoint`: a tool environment that answers each call by sending it to a remote endpoint. The endpoint owns the tool's behaviour; the environment owns the contract around it.

## How one call flows

```
call(tool_id, arguments, call_id=…, session_id=…)
  │
  ├─ tool known to this environment? ──── no ──►  ToolLookupError
  ├─ arguments vs the tool's schema ───── no ──►  ToolArgumentError
  │
  ▼
protocol.call(RemoteToolCall)          owns the wire format
  │
  ▼
client.post_json(payload)              owns the URL, the credential, egress
  │        payload = {name, arguments, call_id, session_id}
  │
  ├─ 4xx ─────────────────────────────────────►  ToolError: the tool rejected it
  ├─ any other non-2xx, or unreachable ───────►  EndpointCallError
  ▼
response body  ─ is the tool's output
  │
  ├─ vs the tool's output_schema ──────── no ──►  EndpointCallError
  ▼
ToolResult(output=…)
```

## Changes

- `EndpointEnvironment` and `EndpointEnvironmentKwargs` (`endpoint_url`, `timeout_seconds` default 30, `max_retries` default 3).
- `EndpointProtocol`, a single-method `Protocol` injected at construction, owning the wire format. `JsonHttpProtocol` is the one implementation: one JSON POST per call, the response body is the output. A second protocol (MCP, OpenAPI) is a new class and leaves `EndpointEnvironment` untouched.
- `JsonHttpClient`, the `Protocol` `JsonHttpProtocol` sends over, so the URL, credential and egress policy are never the protocol's concern. `RequestsJsonClient` is the default; an embedder supplying pinned DNS or a proxy substitutes its own. A protocol needing more than a JSON POST declares its own client rather than widening this one.
- Two failure lanes: `ToolError` is the tool's own answer and reaches the model verbatim; `EndpointCallError` is the endpoint failing to serve the call. `EndpointCallError` deliberately does **not** subclass `ToolError`, so a caller can tell them apart.
- `EndpointStatusError` carries the status a non-2xx answer came back with, which is what lets the protocol — not the client — decide what counts as a refusal.
- Retries on the default client for connection failures and 429/5xx. `urllib3` excludes POST by default; it is opted in here because every attempt resends the identical body.
- Exported from `oumi.environments`. 32 unit tests, including a flaky endpoint proving a retry resends the identical body, and protocols carrying the same call as JSON-RPC and as operation-routed requests.

## Two details worth review

**Identifiers.** Every request carries a `session_id` for the conversation and a `call_id` for the call within it. A retry resends the identical request, so an endpoint that deduplicates on `{session_id}:{call_id}` performs a side effect once however often it is re-sent. Both are required on `call()`: this environment is shared across samples, so it cannot know which conversation a call belongs to. `step()` treats its batch as one conversation and generates both.

**Base class.** It subclasses `BaseEnvironment` rather than `ExecutableEnvironment`, making `LookupEnvironment` its closest sibling. `ExecutableEnvironment` exists to dispatch to per-tool, user-named executors, which would mean stamping a placeholder executor onto every tool for no gain here. The cost of that choice is `call()` sitting beside `step()`: `step()`'s signature has nowhere to carry per-call identifiers.

## Before submitting


src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+            raise ValueError("endpoint_url is required.")
+        if self.timeout_seconds <= 0:
+            raise ValueError("timeout_seconds must be positive.")
+
+
+@dataclass(frozen=True)
+class RemoteToolCall:
+    """One tool call to send, identified within its conversation."""
+
+    name: str
```
wizeng23, line 99: nit: Can you add comments per field?

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+    tool_params_cls = ToolParams
+
+    def __init__(self, params: EnvironmentParams, protocol: EndpointProtocol) -> None:
+        """Bind the env to the protocol that reaches its endpoint."""
+        self._params = params
+        self._protocol = protocol
+        self._tools_by_id: dict[str, ToolParams] = {
+            tool.id: tool for tool in params.tools
+        }
+        self._session_id = uuid.uuid4().hex
```
wizeng23: The environment class may be shared by multiple rollouts, ex. we spin up one endpoint environment, and use it for several rows in a dataset. Thus, it should be passed in by the caller as opposed to managed by the environment. I'm not certain if the framework is set up in OSS to properly track session_id per rollout and pass it into envs. If we have blockers to implement it here, we can not supply it for now IMO and fix it after

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+        tool_id: str,
+        arguments: dict[str, Any],
+        *,
+        call_id: str,
+        session_id: str | None = None,
+    ) -> ToolResult:
+        """Execute one tool call against the endpoint.
+
+        Raises:
+            ToolLookupError: If the environment does not serve ``tool_id``.
```
wizeng23: nit: in this PR, do you mind switching over LookupEnvironment to use this as well at lookup_environment.py:224?

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def __init__(self, url: str, timeout_seconds: float) -> None:
+        """Send every call to ``url``, waiting at most ``timeout_seconds``."""
+        self._url = url
+        self._timeout_seconds = timeout_seconds
+
+    def post_json(self, payload: JsonValue) -> JsonValue:
+        """POST ``payload`` and return the decoded response body."""
+        response = requests.post(self._url, json=payload, timeout=self._timeout_seconds)
+        response.raise_for_status()
```
wizeng23: I think it makes sense for 4xx errors to be ToolError instead of EndpointCallError

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+        """Build the env from its configured kwargs, over a plain JSON POST."""
+        kwargs = parse_env_kwargs(
+            EndpointEnvironmentKwargs, params, env_label="EndpointEnvironment"
+        )
+        client = RequestsJsonClient(kwargs.endpoint_url, kwargs.timeout_seconds)
+        return cls(params, JsonHttpProtocol(client))
+
+    def close(self) -> None:
+        """Release the protocol's resources, if it holds any."""
+        self._protocol.close()
```
wizeng23: `requires_isolation()` is false, which means close() actually never gets called on this env. That's correct, because this env is stateless and can be shared by multiple rollouts. Compare that to the database env, which has per-rollout state. Thus, you don't need to override this function, and whoever creates this env should be responsible for cleanup.

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+
+class RequestsJsonClient:
+    """Default client: a plain JSON POST with no egress policy of its own."""
+
+    def __init__(self, url: str, timeout_seconds: float) -> None:
+        """Send every call to ``url``, waiting at most ``timeout_seconds``."""
+        self._url = url
+        self._timeout_seconds = timeout_seconds
+
+    def post_json(self, payload: JsonValue) -> JsonValue:
```
wizeng23, line 189: nit: could we add 3 retries here? I think that's a reasonable default, and what actually exercises call_id. We can expose it as a param next to timeout_seconds. Not sure if there's some specific setting we'd need to enable retries on a POST though

src/oumi/environments/endpoint_environment.py
```diff
@@ -0,0 +1,275 @@
...
+from oumi.core.configs.params.tool_params import ToolError, ToolLookupError, ToolParams
+from oumi.core.registry import register_environment
+from oumi.core.types.tool_call import ToolResult
+from oumi.environments.base_environment import BaseEnvironment
+from oumi.environments.utils import parse_env_kwargs
+
+_DEFAULT_TIMEOUT_SECONDS = 30.0
+
+
+class EndpointCallError(ToolError):
```
wizeng23: We should take a look at what each error type means and how it's communicated to the model. Subclassing ToolError means our tool router/harness won't be able to distinguish the error types when returning to the model (not being able to reach the endpoint vs. the tool call itself failed). See LOU-3722 for tracking task

src/oumi/environments/endpoint_environment.py
min-oumi: I actually had close here for mcp as it requires a cleanup, but seems like this could be removed and let the caller close the client , removed!

src/oumi/environments/endpoint_environment.py
min-oumi: Makes sense to me, just updated to Exception plus docstring!

src/oumi/environments/endpoint_environment.py
min-oumi: oops I missed this part, yes it makes sense to pass in session id from the caller so retries are controlled by caller. fixing

src/oumi/environments/endpoint_environment.py
min-oumi: added retries to just connect errors

src/oumi/environments/endpoint_environment.py
min-oumi: Just noting, the previous close from protocol is removed, and this close in line 200 is the client.close which will be used for mcps

wizeng23: As long as whatever creates the environment is responsible for calling RequestsJsonClient.close(), that sounds good to me. We just don't want it to be closed in-between rollouts if this is a stateless env

wizeng23 (approved):

PR #2626, opened 2026-08-27
Adding `RubricJudge` to support multiple evaluation criteria 

kaisopos:
# Description

Adds a judge that scores several criteria of the same input in one inference call, returning one judgment per criterion plus a weighted aggregate score, where SimpleJudge returns a single judgment.

Criteria are declared in a new `rubric_judge_params` config block; the judge builds the rubric prompt, the response schema, and the output fields from them. With JSON responses, guided decoding constrains the model to the rubric's schema.

Also extracts the prompt/schema helpers shared with SimpleJudge into `judge_utils`, and moves the inference-engine factory onto BaseJudge. Existing judges produce byte-identical prompts and schemas.

**Note**: Though this PR seems large, the only meaningful changes are: 
- src/oumi/core/configs/judge_config.py
- src/oumi/core/configs/params/rubric_judge_params.py
- src/oumi/judges/rubric_judge.py
- src/oumi/cli/judge.py

...everything else is testing or plumbing or moving things around (no need to thoroughly review)!

## Related issues

Towards LOU-3713

## Before submitting

## Reviewers


src/oumi/core/configs/params/rubric_judge_params.py
```diff
@@ -0,0 +1,231 @@
...
+class JudgeAggregation(str, Enum):
+    """How to combine per-criterion scores into a single overall score."""
+
+    NONE = "none"
+    """Do not aggregate; the overall score is left undefined."""
+
+    WEIGHTED_MEAN = "weighted_mean"
+    """Weighted average of the per-criterion scores, using each criterion's weight."""
+
+    MIN = "min"
```
wizeng23, line 41: nit: do we want MAX for completeness?

src/oumi/core/configs/params/rubric_judge_params.py
```diff
@@ -0,0 +1,231 @@
...
+    """The type of output that this criterion's judgment should be provided with."""
+
+    judgment_scores: dict[str, float | None] | None = field(default=None)
+    """For ENUM judgment_type, the mapping from category names to numeric scores.
+
+    Example:
+        {"excellent": 1.0, "good": 0.7, "poor": 0.3, "not_applicable": None}
+    """
+
+    include_explanation: bool = field(default=False)
```
wizeng23, line 81: nit: wdyt about making the default True? I feel like that could be the more reasonable default behavior, as the explanation makes the judge score more trustable. Users can turn it off to save tokens, but IMO that's something they should opt into.

Tying to the comment later, if we're using guided decoding to force the judge to only output the JSON, it doesn't have token space to reason before its answer, hurting its quality. To the best of my understanding, guided decoding oftentimes can lead to performance degradation due to preventing the model from thinking freely. Explanations could mitigate this; since the judge generates the explanation before the score, it could use the explanation text as a way to reason.

src/oumi/core/configs/params/rubric_judge_params.py
```diff
@@ -0,0 +1,231 @@
...
+        seen_ids: set[str] = set()
+        for criterion in self.criteria:
+            if criterion.id in seen_ids:
+                raise OumiConfigError(
+                    f"Duplicate criterion id: '{criterion.id}'. Criterion ids must "
+                    "be unique."
+                )
+            seen_ids.add(criterion.id)
+
+        if self.aggregation == JudgeAggregation.WEIGHTED_MEAN and not any(
```
wizeng23, line 225: nit: I saw you warn when weight is set when aggregation method isn't weighted mean, but it's done in rubric_judge.py. That in theory could live here instead right?

src/oumi/core/configs/params/rubric_judge_params.py
```diff
@@ -0,0 +1,231 @@
...
+    Example:
+        >>> rubric_params = RubricJudgeParams( # doctest: +SKIP
+        ...     criteria=[
+        ...         JudgeCriterion(
+        ...             id="groundedness",
+        ...             description="Every claim is supported by the context.",
+        ...             include_explanation=True,
+        ...         ),
+        ...         JudgeCriterion(
+        ...             id="tone",
```
wizeng23, line 185: nit: what about "formality" instead? tone is a neutral word, which doesn't indicate that a formal tone scores better

src/oumi/core/configs/params/rubric_judge_params.py
```diff
@@ -0,0 +1,231 @@
...
+        ...             description="The register of the response.",
```
wizeng23, line 186: nit: description is pretty vague

src/oumi/judges/rubric_judge.py
```diff
@@ -0,0 +1,390 @@
...
+
+        if self._inference_config is None:
+            raise ValueError(
+                "inference_config must be provided in JudgeConfig for RubricJudge. "
+                "Please ensure your JudgeConfig includes a valid inference_config."
+            )
+        json_format = self._judge_params.response_format == JudgeResponseFormat.JSON
+        inference_engine = self._create_inference_engine(
+            inference_config=self._inference_config,
+            response_schema=self._build_response_schema() if json_format else None,
```
wizeng23, line 112: Is it documented anywhere that we enable guided decoding for JSON format? IMO it may be worth making this configurable. My personal experience is guided decoding can hurt quality, since it limits the LLM's ability to reason. An alternative is to not constrain the response and have a tolerant JSON parser (i.e. strip text, try to parse with json.loads(), and if it fails fall back to manual parsing).

wizeng23 (approved):

src/oumi/core/configs/params/rubric_judge_params.py
min-oumi: Min says hi to min equals min

kaisopos: I dont think MAX would be particularly useful. MIN practically means zero tolerance (no offense @min-oumi): all criteria have to be met. MAX would mean that if any criteria is met then we score 1.0 ( I think nobody will use this)

src/oumi/core/configs/params/rubric_judge_params.py
kaisopos: Agree and fixed in #2628

src/oumi/core/configs/params/rubric_judge_params.py
kaisopos: Great suggestion - thanks!
Fixed fixed in #2628

src/oumi/core/configs/params/rubric_judge_params.py
kaisopos: fixed in #2628

src/oumi/core/configs/params/rubric_judge_params.py
kaisopos: fixed in #2628

src/oumi/judges/rubric_judge.py
kaisopos: Made it configurable in #2628
I have not had a similar experience, but will do a proper quality review once this is complete E2E

PR #2628, opened 2026-08-28
RubricJudge - part 2

kaisopos:
# Description

Follow up on Rubric Judge
1. Addressing Will's feedback on #2626 
2. Making guided decoding configurable

Specifically
- `include_explanation` now defaults to True. The explanation is emitted before the judgment, so the judge reasons before committing to an answer, which matters most under guided decoding where a schema-constrained response otherwise leaves no room to think.
- Guided decoding is now configurable via `judge_params.use_guided_decoding` (default True, JSON only), for callers who would rather let the model generate freely.
- The judge's JSON parser no longer raises on valid JSON that is not an object, and now tolerates prose around the object. Both cases become reachable once guided decoding can be turned off.
- Aggregation warnings moved from RubricJudge into RubricJudgeParams, next to the existing aggregation/weight validation. They now fire once at config load rather than once per judge.
- Docstring example: renamed `tone` to `formality`, since its scores reward formality, and made its description concrete.

Follow up:  Documentation for RubricJudge will follows in a separate PR (by EOD)

## Related issues

Towards LOU-3713

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/configs/params/judge_params.py
+++ b/src/oumi/core/configs/params/judge_params.py
@@ -120,6 +120,9 @@ class JudgeParams(BaseParams):
     include_explanation: bool = field(default=False)
     """Whether the judge should provide an explanation before the judgment."""
 
+    use_guided_decoding: bool = field(default=True)
+    """Whether to constrain the judge's response (JSON only) to the expected schema."""
+
     examples: list[dict[str, str]] = field(default_factory=list)
     """Few-shot examples for the judge as a list of field value dictionaries.
 
--- a/src/oumi/judges/base_judge.py
+++ b/src/oumi/judges/base_judge.py
@@ -215,6 +215,14 @@ class JudgeOutput(pydantic.BaseModel):
         Returns:
             Dictionary of field names to values, empty dict if parsing fails
         """
+
+        def _loads_json_object(text: str) -> dict | None:
+            try:
+                parsed = json.loads(text)
+            except json.JSONDecodeError:
+                return None
+            return parsed if isinstance(parsed, dict) else None
+
         if not json_output:
             return {}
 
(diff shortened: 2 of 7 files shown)


src/oumi/core/configs/params/judge_params.py
```diff
@@ -120,6 +120,9 @@ class JudgeParams(BaseParams):
     include_explanation: bool = field(default=False)
     """Whether the judge should provide an explanation before the judgment."""
 
+    use_guided_decoding: bool = field(default=True)
```
wizeng23, line 123: What if the tool output schema isn't provided? This does nothing then right? Can we document that

src/oumi/judges/base_judge.py
```diff
@@ -224,12 +232,19 @@ def _parse_json_output(cls, json_output: str | None) -> dict[str, str]:
...
-        except json.JSONDecodeError:
-            return {}
+        parsed = _loads_json_object(json_output)
+        if parsed is None:
+            # Models that are not schema-constrained often wrap the object in
+            # prose; retry on the outermost {...} span before giving up.
+            start, end = json_output.find("{"), json_output.rfind("}")
+            if start == -1 or end <= start:
+                return {}
+            parsed = _loads_json_object(json_output[start : end + 1])
```
wizeng23, line 242: As a future extension, we can also attempt manual parsing here as a fallback, ex. if the LLM makes a small mistake that messes up json.loads, we can still try to recover the result

wizeng23 (approved):

src/oumi/core/configs/params/judge_params.py
kaisopos: The schema is not provided by the user; we build from the judge's output fields, so it's never absent.

src/oumi/judges/base_judge.py
kaisopos: Agreed, worth doing at some point. Only caveat is scoping it to deterministic fixes like trailing commas or single quotes. Recovering truncated output risks silently producing a wrong judgment, which is worse than the current `None`.

PR #2632, opened 2026-08-28
fix(inference): keep the scratch-file hash per infer() call

rlehman221:
## Description

`BaseInferenceEngine` keeps the current run's dataset hash on the engine instance (`self._dataset_hash`, set in `_prepare_inference_run`) and derives the scratch-file path from it. Two `infer()` calls in flight on the same engine — e.g. one `SimpleJudge` scored from a thread pool — overwrite each other's hash, resolve to the same scratch file, and race in `_cleanup_scratch_file` (`if exists: unlink`), which then fails with `FileNotFoundError` on `~/.cache/oumi/tmp/temp_inference_output_<hash>.jsonl`.

### Changes

- `base_inference_engine.py`: the dataset hash lives in a `contextvars.ContextVar` instead of on the instance, exposed through the same `_dataset_hash` property/setter so existing callers and tests are unchanged. Concurrent calls (threads, or tasks on one loop) each see their own hash and scratch path.
- `async_utils.safe_asyncio_run`: copies the caller's context into the helper thread (`contextvars.copy_context().run(...)`) so the hash survives the thread hop into `_infer`.
- `_cleanup_scratch_file`: `unlink(missing_ok=True)` — two runs with an identical dataset legitimately share a path, and either may clean up first.

### Tests

- `test_concurrent_infer_calls_keep_separate_scratch_files`: two threads share one engine with an overlapping save; both complete and resolve distinct scratch paths. Fails on `main` (paths collapse to one).
- `test_cleanup_scratch_file_tolerates_missing_file`.
- `test_safe_asyncio_run_propagates_context_variables`.

## Related issues

Towards https://linear.app/oumi/issue/LOU-3735

## Before submitting

## Reviewers

### DIFF
--- a/src/oumi/core/inference/base_inference_engine.py
+++ b/src/oumi/core/inference/base_inference_engine.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextvars
 import copy
 import dataclasses
 import hashlib
@@ -36,6 +37,11 @@ from oumi.core.inference.progress_reporter import ProgressFileReporter
 from oumi.core.types.conversation import Conversation
 from oumi.utils.logging import logger
 from oumi.utils.math_utils import is_power_of_two
+
+# Per-call, not per-engine: concurrent infer() calls must not share a scratch file.
+_dataset_hash_var: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "oumi_inference_dataset_hash", default=None
+)
 
 
 @dataclass
@@ -172,7 +178,15 @@ class BaseInferenceEngine(ABC):
 
         self._latency_histogram_online = HdrHistogram(1, 60 * 1000, 1)
         self._latency_histogram_from_file = HdrHistogram(20, 180 * 1000, 1)
-        self._dataset_hash = None
+
+    @property
+    def _dataset_hash(self) -> str | None:
+        """Dataset hash of the ``infer()`` call running in this execution context."""
+        return _dataset_hash_var.get()
+
+    @_dataset_hash.setter
+    def _dataset_hash(self, value: str | None) -> None:
+        _dataset_hash_var.set(value)
 
     def _prepare_inference_run(
         self,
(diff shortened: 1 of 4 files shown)


tests/unit/inference/test_base_inference_engine.py
```diff
@@ -734,3 +735,49 @@ def test_existing_infer_tests_unaffected_by_partial_run(mock_engine):
...
+
+    def run(idx: int) -> None:
+        try:
+            results[idx] = mock_engine.infer(
+                input=[create_test_conversation(idx)],
+                inference_config=InferenceConfig(
+                    generation=GenerationParams(max_new_tokens=10)
+                ),
+            )
+        except BaseException as e:  # noqa: BLE001 - surface any thread failure
```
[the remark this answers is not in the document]
rlehman221: Added!

src/oumi/core/inference/base_inference_engine.py
```diff
@@ -680,9 +694,9 @@ def _cleanup_scratch_file(self, output_filepath: str | None) -> None:
             output_filepath: The path to the output file. This is used to determine the
                 location of the scratch file.
         """
-        scratch_filepath = self._get_scratch_filepath(output_filepath)
-        if Path(scratch_filepath).exists():
-            Path(scratch_filepath).unlink()
+        # missing_ok: a concurrent run with an identical dataset shares the path
```
shanghongsim, line 697: Does this mean that two runs with an identical dataset can override each other? Perhaps not in the scope of this PR but is that error mode caught somewhere? Did a bit of claude investigation and i see that infer calls append results. So two calls with same dataset means output size will be 2x with duplicates for every sample?

shanghongsim (approved):

src/oumi/core/inference/base_inference_engine.py
rlehman221: So output can't double since the final output_path write is write mode over a conversation_id keyed dict, so it's always exactly len(input).

But identical dataset runs do share a scratch path which was the functionality we've always had from before

PR #2646, opened 2026-09-04
Wrap verl's compute metrics function for richer metric logging

shanghongsim:
# Description

GRPO learns by comparing rewards among the rollouts generated for the same prompt. If every rollout in a prompt group receives the same or very similar rewards, the group produces zero or very little relative policy-gradient signal.

The relevant metrics currently emitted by VERL are:

- `critic/score/{mean,max,min}`: raw sequence-level score across all rollouts in the batch.
- `critic/rewards/{mean,max,min}`: effective reward after any reward-side KL penalty.
- `critic/advantages/{mean,max,min}`: token-weighted advantages across the entire batch.
- `critic/returns/{mean,max,min}`: token-weighted returns across the entire batch.
- `actor/pg_loss` and `actor/grad_norm`: downstream indicators of whether an update occurred.

However, VERL does not retain or log the group standard deviations after computing the advantages.

This PR wraps VERL's driver-side `compute_data_metrics` function, where the complete rollout batch is available, and adds the following metrics:

  - `grpo/frac_reward_zero_std`: fraction of prompt groups with effectively zero reward standard deviation.
  - `grpo/frac_reward_low_std`: fraction of groups whose reward standard deviation is below a configurable threshold.
  - `grpo/reward_group_std/mean`: mean reward standard deviation across prompt groups.
  - `grpo/reward_group_std/p50`: median prompt-group reward standard deviation.
  - `grpo/reward_group_std/p90`: 90th percentile prompt-group reward standard deviation.

I considered implementing this both via 
1. patching VERL’s installed source with a .patch file in the launch script and wrapping its metric collector at runtime
2. runtime wrapper

I opted for 2 in the end because the metric calculation is called only by the central trainer and does not span the separate rollout processes. Having it as runtime wrapper also makes the code testable and not all users of OSS will launch experiments through shell script. 

Validated by running a training job for 5 steps. The added metrics log successfully to wandb.
image

## Related issues

Fixes # (issue)

## Before submitting

## Reviewers


lefft (approved):
This will be nice to have!

PR #2658, opened 2026-09-11
Tool-use training support: tool-call template access, trainable adapters, text DPO JSONL, collator masking

jgreer013:
## Summary

Four independent changes supporting supervised/preference training on tool-use conversations. Each is small and self-contained with unit tests; happy to split into separate PRs.

- **tool-call types: dict-style field access for chat templates.** Some chat templates read their inputs as mappings (`tool_call.get('function')`, `function['arguments']`); a Pydantic `ToolCall`/`FunctionCall` reaches them with no `.get` or `__getitem__`, and rendering fails. Adds a `_TemplateFieldAccess` mixin exposing declared fields dict-style. `FunctionCall` additionally decodes `arguments` to its mapping on subscript, while `.arguments` keeps the OpenAI wire string as the storage contract. `Message` already carries the same accommodation.
- **`ModelParams.adapter_trainable`.** PEFT loads adapters frozen by default (`is_trainable=False`), which blocks continuing to train an existing adapter (e.g. DPO on top of an SFT LoRA). Threads the flag (default `False`) through the adapter load in `build_huggingface_model`.
- **`TextDpoJsonlinesDataset` (`text_dpo_jsonl`).** A text counterpart to the existing `VisionDpoJsonlinesDataset`: DPO over JSONL rows of pre-rendered `prompt`/`chosen`/`rejected` strings, tokenized directly without re-applying the chat template.
- **Collator: keep the model's tool-response hand-off token in the loss.** In completion-only masking the opening tool-response marker is masked with the environment output. Where a template has the model emit that marker as a stop token (ending its turn to hand control to the environment), masking it removes the only signal for stopping after a tool call. Kept in the loss where it follows the model's own calls; masked only where it follows another result (the template framing a parallel block the model cannot predict).

## Before submitting

- [x] Unit tests added for each change
- [ ] Ran `make test`


taenin (approved):

shanghongsim:
@jgreer013 Could you take a look at the potential GPU-test regression introduced by this PR? I traced it to the exact commit boundary:
  
- The last run before #2658, at d66a2a92, was green: https://github.com/oumi-ai/oumi/actions/runs/34641423362
- The next commit was solely the squash merge of #2658, 5aaa5e5e.
- Its merge-queue run immediately failed on gemma-4-nested and glm-4.5-nested: https://github.com/oumi-ai/oumi/actions/runs/34652803810
- Scheduled main runs have continued failing identically.

The failure also follows directly from the masking change: the implementation now leaves the first tool-response opener unmasked, while `_masked_bracket_regions_are_complete` still requires every token from the opener through the closer to be masked. That makes the assertion deterministically fail for these nested templates.
 
Could you please reconcile the intended behavior with the integration test so main goes green again?

PR #2661, opened 2026-09-14
fix(configs): resolve dual-mode/VLM detection under transformers 5

wizeng23:
# Description

_model_mapping is keyed by model_type strings in transformers 5, so indexing it by config class always returned None — every model was misclassified as non-dual-mode and non-VLM, breaking text_only detection for Qwen3.5. Resolve via mapping[cfg_cls] (which does the lazy class load) instead, and add a real-config regression test the mocked tests missed.

Towards OPE-2158

## Before submitting


src/oumi/core/configs/internal/supported_models.py
```diff
@@ -767,6 +767,20 @@ def find_internal_model_config(
     )
 
 
+def _resolve_auto_model_class(mapping, cfg_cls: type) -> type | None:
```
lefft: @wizeng23 sorry if I'm missing something, but why a new helper with multi-line comment instead of just saying `MODEL_FOR_CAUSAL_LM_MAPPING.get(config_cls, None)` inline at the call sites?

wizeng23: Nope you're right, was written inefficiently

min-oumi (approved):

PR #2668, opened 2026-09-16
Fix Gemma 4 KV-shared inference on vLLM 0.19.1

shanghongsim:
# Description

Gemma 4 checkpoints saved from transformers 5.5.2 no longer store K/V projection and norm weights for KV-shared attention layers. vLLM 0.19.1 still builds those layers with a packed QKV projection, so loading one of these checkpoints fails on the missing weights.

```
ValueError: Following weights were not initialized from checkpoint:
language_model.model.layers.15.self_attn.k_norm.weight
...
language_model.model.layers.34.self_attn.k_norm.weight
```

vLLM fixed this upstream in [#54917](https://github.com/vllm-project/vllm/pull/54917). This PR carries the same behavior for Oumi users who are pinned to vLLM 0.19.1:

- KV-shared Gemma 4 layers use a Q-only projection and reuse the producer layer's cached K/V tensors.
- The weight loader accepts both the newer checkpoint layout and older Google checkpoints that still contain the redundant shared-layer tensors.
- Non-shared Gemma 4 layers are unchanged.

The compatibility code is registered as a `vllm.general_plugins` entry point. That matters because vLLM starts the engine and workers in spawned processes; patching only the Oumi driver does not affect the processes that construct the model. The plugin remains inactive unless Oumi is creating a vLLM inference engine, and it refuses to install for any vLLM version other than 0.19.1.

This is temporary compatibility code. It can be removed once Oumi moves to a verified vLLM release containing commit `f2d45f26bd6a2c841ffbd0030ebaefe87c274e58`.

Testing:

- Unit tests cover activation/version guards, idempotency, layer topology, weight filtering, forward behavior, plugin registration, and engine setup.
- Tested in BF16 on an H100 with `VLLM_WORKER_MULTIPROC_METHOD=spawn` against both the original Google checkpoint layout and a Transformers 5.5.2 checkpoint with the shared K/V weights omitted.
- For eight fixed prompts, vLLM matched Transformers greedy token IDs. Two fresh vLLM runs also produced identical output. Selected-token log-probability differences stayed within the test thresholds.
- A 1,000-example platform evaluation completed successfully using the patched Oumi commit and vLLM 0.19.1.

## Related issues

- Related to [vLLM issue #44788](https://github.com/vllm-project/vllm/issues/44788)
- Backports [vLLM PR #54917](https://github.com/vllm-project/vllm/pull/54917)

## Before submitting

## Reviewers


shanghongsim:
GPU failures are from a pre existing bug introduced by another PR

lefft:
@shanghongsim One question re:

> Gemma 4 checkpoints saved from transformers 5.5.2 no longer store K/V projection and norm weights for KV-shared attention layers.

In api, [we are pinning `transformers==5.5.0` for E4B](https://github.com/oumi-ai/api/blob/main/shared/src/shared/model_setup_overrides.py#L157-L170) for both training and inference. 

Under what conditions will gemma4_compat.py be applied? 

If this fixes transformers==5.5.2 + vllm==0.19.1, then we should validate that we no longer need to pin to 5.5.0 in api.

lefft:
@shanghongsim Do you know what an alternative that's scoped to GRPO with verl + tf 5.5.2 in api would look like? 

AFAIK this patch would apply to any scenario where a gemma4 model is being served with vllm 0.19.1. But IIUC, the problem that it solves arises only when it is a customized gemma4 that has been saved with tf 5.5.2. 

If this is temporary and only resolves a problem that arises for customized gemma4 models saved with tf 5.5.2, wondering how we can scope it to only apply in that specific scenario.

lefft (commented):
Before finalizing, wondering if we can compare this fix to an alternative that is scoped only to the scenario that is causing the problem?

lefft (changes requested):
@shanghongsim  I think we can pick this one up if we can confirm that the SFT setups in api work smoothly while consuming the fix here!

lefft (approved):
Was able to verify that this unblocks 5.5.2 for SFT as well. 

So after this merges, we can advance oumi submodule in https://github.com/oumi-ai/api/pull/7051 to safely bump to 5.5.2 for E2B/E4B SFT. 

Thanks!

PR #2673, opened 2026-09-22
Fix DPO reference model loading with FSDP

lefft:
## Description

TRL automatically loads a frozen reference model for full-fine-tuning DPO. Under FSDP, its default `device_map="auto"` can place that model on a different local device before FSDP wraps it, causing a device mismatch. Its default FP32 loading also uses unnecessary memory when the policy model uses a lower-precision dtype.

For DPO with FSDP, the Oumi wrapper now:

- disables automatic device placement for the reference model;
- defaults the reference model to the policy model's dtype; and
- preserves other caller-provided model initialization arguments.

Other distributed strategies are unchanged.

## Testing

- `PYTHONPATH=src python -m pytest -p no:rerunfailures tests/unit/core/trainers/test_trl_dpo_trainer.py -q` — 11 passed
- Ruff check and format verification passed for both changed files.


lefft:
Closing out, see https://github.com/oumi-ai/oumi/pull/2675

